PR #40, opened 2019-03-29
fix(install): Try to fix installation of tools on windows by appending .exe

jaysonsantos:
Hey there, here I tried to fix installation on windows but when it
downloaded cargo generate from
https://github.com/ashleygwilliams/cargo-generate/releases/download/v0.2.2/cargo-generate-v0.2.2-x86_64-pc-windows-msvc.tar.gz
The final tar ball didn't have the exe file there, just licenses and
readme.

### DIFF
--- a/src/install/mod.rs
+++ b/src/install/mod.rs
@@ -1,6 +1,7 @@
 mod krate;
 mod target;
 
+use failure::ResultExt;
 use krate::Krate;
 
 use binary_install::{Cache, Download};
@@ -14,12 +15,7 @@ pub fn install(tool_name: &str, owner: &str, cache: &Cache) -> Result<Download, 
 
     let latest_version = get_latest_version(tool_name)?;
     let download = download_prebuilt(cache, tool_name, owner, &latest_version);
-    match download {
-        Ok(download) => Ok(download),
-        Err(_) => {
-            failure::bail!("could not download pre-built `{}`.", tool_name);
-        }
-    }
+    Ok(download.context(format!("could not download pre-built `{}`.", tool_name))?)
 }
 
 fn download_prebuilt(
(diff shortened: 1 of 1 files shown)


src/install/mod.rs
```diff
@@ -37,7 +33,13 @@ fn download_prebuilt(
         )),
     };
 
-    let binaries = &[tool_name];
+    let binary_name = if cfg!(target_os = "windows") {
+        format!("{}.exe", tool_name)
+    } else {
+        tool_name.to_owned()
```
jaysonsantos: Here I had to make a copy of it because of the format above, any way to make it better?

ashleygwilliams:
oh! i *think* this is likely a bug in cargo-generate and how it makes windows binaries... the repo is https://github.com/ashleygwilliams/cargo-generate- i'd take a look at the .appveyor.yml (i recall we had this issue for wasmbindgen and we fixed it so i can also look that up in a sec)

ashleygwilliams:
if you can file an issue on cargo-generate and take a look i would welcome a PR there, this is definitely something we should fix (did wasm-pack have the same issue?)

jaysonsantos:
I will just create the issue because I checked appveyor file and it seems fine and also looks like it didn't run the build for a while, maybe it is that?

jaysonsantos:
Actually, I found a build that has the problem https://ci.appveyor.com/project/ashleygwilliams/cargo-generate/builds/20437209

ashleygwilliams:
i'm closing this in favor of the PR on cargo generate! thank you!

jaysonsantos:
Hey there, I am not sure if only that would fix the problem because the library binary-install will check for an exact match:

```rust
       let mut binaries: HashSet<_> = binaries.into_iter().map(ffi::OsStr::new).collect();
...
        for entry in archive.entries()? {
            let dest = match entry.path()?.file_stem() {
                Some(f) if binaries.contains(f) => {
                    binaries.remove(f);
                    dst.join(entry.path()?.file_name().unwrap())
                }
                _ => continue,
            };
        }
```
As soon as you release cargo-generate 0.3.0 I will test again with this branch and see what happens.

PR #41, opened 2019-03-29
Update README.md

### LINKED ISSUE #35: document openssl fix
I attempted the tutorial [here](https://developers.cloudflare.com/workers/webassembly/tutorial/) but ran into a problem with openssl:
I installed cargo 
`$ curl https://sh.rustup.rs -sSf \| sh`
and configured my shell with:
`$ source $HOME/.cargo/env`
I then succesfully installed wrangler with
`$ cargo install wrangler`
and tried to generate a "Hello, World" project with:
`$ wrangler generate demo`
However, that throws an error:

dyld: Library not loaded: /usr/local/opt/openssl@1.1/lib/libssl.1.1.dylib
  Referenced from: /Users/userspace/Library/Caches/.wrangler/cargo-generate-2665b6ba17ab8008/cargo-generate
  Reason: image not found
Error: ErrorMessage { msg: "failed to execute `/Users/userspace/Library/Caches/.wrangler/cargo-generate-2665b6ba17ab8008/cargo-generate generate --git https://github.com/cloudflare/rustwasm-worker-template --name demo`: exited with signal: 6" }

So I instead tried
`$ cargo install cargo-generate`
which led to a similar error:

It looks like you're compiling on macOS, where the system contains a version of
OpenSSL 0.9.8. This crate no longer supports OpenSSL 0.9.8.

As a consumer of this crate, you can fix this error by using Homebrew to
install the `openssl` package, or as a maintainer you can use the openssl-sys
0.7 crate for support with OpenSSL 0.9.8.

Unfortunately though the compile cannot continue, so aborting.

', /Users/username/.cargo/registry/src/github.com-1ecc6299db9ec823/openssl-sys-0.9.43/build/find_normal.rs:150:5
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

warning: build failed, waiting for other jobs to finish...
error: failed to compile `cargo-generate v0.2.2`, intermediate artifacts can be found at `/var/folders/j7/_vypfcb179g8wbw84pkrrkx00000gn/T/cargo-installdX02Cv`

AustinCorridor:
fixes #35 

Documents a solution to Issue #35 where a specific version of OpenSSL is needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 Cloudflare Workers. This tool gives you the follow commands:
 
   - 👯 `generate`: scaffold  a `hello-wasm-worker` project, including boilerplate for a Rust library and a
-     Cloudflare worker 
+     Cloudflare worker
   - 🦀⚙️ `build`: build your project using `wasm-pack`
   - 🔬 `preview`: preview your project using the cloudflareworkers.com API
   - ☁️ 🆙 `publish`: publish your worker and WebAssembly to Cloudflare
@@ -23,6 +23,13 @@ To set up `wrangler` to work with your Cloudflare account, use the following com
 
     ⚠️ NEVER PUBLISH CREDENTIALS TO VERSION CONTROL! ⚠️
 
+##
+
+Before you begin, you'll need to install OpenSSL 1.1 on your machine:
+
+```
+brew install openssl@1.1
+```
 
 ## ⚡ Quick Start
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,6 +23,13 @@ To set up `wrangler` to work with your Cloudflare account, use the following com
 
     ⚠️ NEVER PUBLISH CREDENTIALS TO VERSION CONTROL! ⚠️
 
+##
```
ashleygwilliams: so- i think we want to be careful here because not all users will have this issue and not all users are on a Mac, with Homebrew.

i'd consider taking step 1 of the quickstart out and making it its own "Installation" section. then i would add this as a note saying if you get an SSL error and are on a mac with homebrew (link to how to install homebrew) and then give the commend :)

AustinCorridor: Good point; fixed!

PR #43, opened 2019-03-30
Adds install instructions and troubleshoot notes for OpenSSL errors

AustinCorridor:
* Adds detail for install rustup
* Adds detail for installing Cargo
* Adds note about OpenSSL version error

I deleted an old branch after wrangling with a couple silly issues on my side


AustinCorridor:
Tabs vs spaces in markdown, and differences in the Atom markdowns rendering package from the real thing, this should be ready to go.

PR #68, opened 2019-04-10
feat(publish): add force flag

### LINKED ISSUE #67: only overwrite on force... for now

ashleygwilliams:
🤕 do no merge 🤕 

fixes #67


ashleygwilliams:
after conversations in air and chat, we've decided that overwrite should be the default and not an option, closing this.

Issue #71, opened 2019-04-23
Generate command always prints rustwasm copy


signalnerve:
https://github.com/cloudflare/wrangler/blob/9d4f33137a787ac5eb8887d291d50cdc9705df92/src/commands/generate.rs#L23

The generate command will always print `Generating a new rustwasm worker project` even if the passed template is unrelated to rustwasm.

Maybe just `Generating a new Wrangler project` would be more appropriate here?

ashleygwilliams:
this is fixed by #72 - you can test it by installing from master

PR #72, opened 2019-04-24
feat(generate): default to js template

### LINKED ISSUE #3: put wasm behind a flag, default to js only worker
not actually sure we want this for 0.1.0

ashleygwilliams:
fixes #3 #71

### DIFF
--- a/src/commands/generate.rs
+++ b/src/commands/generate.rs
@@ -1,4 +1,4 @@
-use crate::user::settings::ProjectSettings;
+use crate::user::settings::{ProjectSettings, ProjectType};
 use crate::{commands, install};
 use binary_install::Cache;
 use std::path::PathBuf;
@@ -10,18 +10,19 @@ pub fn generate(name: &str, template: &str, cache: &Cache) -> Result<(), failure
 
     let args = ["generate", "--git", template, "--name", name];
 
-    let command = command(name, binary_path, &args);
+    let project_type = project_type(template);
+    let command = command(name, binary_path, &args, &project_type);
     let command_name = format!("{:?}", command);
 
     commands::run(command, &command_name)?;
-    ProjectSettings::generate(name.to_string())?;
+    ProjectSettings::generate(name.to_string(), project_type)?;
     Ok(())
 }
 
-fn command(name: &str, binary_path: PathBuf, args: &[&str]) -> Command {
+fn command(name: &str, binary_path: PathBuf, args: &[&str], project_type: &ProjectType) -> Command {
     println!(
-        "🐑 Generating a new rustwasm worker project with name '{}'...",
-        name
+        "🐑 Generating a new {} worker project with name '{}'...",
+        project_type, name
     );
 
     let mut c = if cfg!(target_os = "windows") {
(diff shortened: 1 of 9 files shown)


src/commands/generate.rs
```diff
@@ -36,3 +37,10 @@ fn command(name: &str, binary_path: PathBuf, args: &[&str]) -> Command {
     c.args(args);
     c
 }
+
+fn project_type(template: &str) -> ProjectType {
+    if template.contains("rust") {
```
steveklabnik: This should probably be:

```suggestion
    if template == "rust" {
```

src/commands/generate.rs
```diff
@@ -36,3 +37,10 @@ fn command(name: &str, binary_path: PathBuf, args: &[&str]) -> Command {
...
+        return ProjectType::Rust;
+    }
+    ProjectType::JavaScript
```
steveklabnik: Also consider:

```rust
if template == "rust" {
    ProjectType::Rust
} else {
    ProjectType::JavaScript
}
```

Though this is of course subjective

src/commands/publish/mod.rs
```diff
@@ -65,16 +39,48 @@ fn multi_script(user: &User, name: &str) -> Result<(), failure::Error> {
...
+    }
 
     Ok(())
 }
 
+fn build_js() -> Result<String, failure::Error> {
+    let mut f = fs::File::open("worker.js")?;
+    let mut buffer = String::new();
+    f.read_to_string(&mut buffer)?;
+    Ok(buffer)
```
steveklabnik: This function body can be replaced entirely with

```rust
std::fs::read_to_string("worker.js")
```

though, that returns an `io::Error`, so you *may* need

```rust
Ok(std::fs::read_to_string("worker.js")?)
```

src/main.rs
```diff
@@ -130,11 +109,25 @@ fn main() -> Result<(), failure::Error> {
...
+            commands::build(&cache, &user.settings.project.project_type)?;
+        }
+
+        if let Some(matches) = matches.subcommand_matches("preview") {
+            let method = HTTPMethod::from_str(matches.value_of("method").unwrap_or("get"));
+
+            let body = match matches.value_of("body") {
+                Some(s) => Some(s.to_string()),
+                None => None,
+            };
```
steveklabnik: This is

```rust
let body = matches.value_of("body").map(|s| s.to_string());
```

src/commands/generate.rs
ashleygwilliams: no- this is matching the url so contains is correct here

src/commands/generate.rs
ashleygwilliams: yeah i prefer it the way it currently is :)

src/commands/publish/mod.rs
ashleygwilliams: yesss, as said in air i don't know why but the docs i end up getting on google search never show this very good optimized way

src/main.rs
ashleygwilliams: yeah i think i actually like this but friggin clippy thinks ur suggestion is a redundant closure and has no problem with what's there atm

Issue #73, opened 2019-04-24
'loudflare' is not recognized as an internal or external command


steveklabnik:
Copy/paste is being a bit wonky right now, but the TL;DR is that the call to `cargo-generate` when you run `wrangler generate` is not properly quoted. This means if there's a space in the path, things break, and it won't properly generate.

I'll try to get a more proper report later.

steveklabnik:
This also affects `wrangler preview`.

PR #74, opened 2019-04-24
properly quote commands

steveklabnik:
This is a start of a fix to #73; I'd like feedback on this code. If you like it, I can fix the other commands as well.

The TL;DR is that you should pass each arg separately, so they get quoted properly, rather than use string formatting to create the command. I tried to modify the surrounding code as little as possible.

### DIFF
--- a/src/commands/publish/route.rs
+++ b/src/commands/publish/route.rs
@@ -28,7 +28,9 @@ impl Route {
 }
 
 fn multi_script(user: &User, script: String) -> Result<Route, failure::Error> {
-    let pattern = &user.settings.clone().project.route.expect("⚠️ Your project config has an error, check your `wrangler.toml`: `route` must be provided.");
+    let pattern = &user.settings.clone().project.route.expect(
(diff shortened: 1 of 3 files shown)


xtuc:
Yes I agree and your fix looks good to me.

src/commands/publish/route.rs
```diff
@@ -28,7 +28,9 @@ impl Route {
 }
 
 fn multi_script(user: &User, script: String) -> Result<Route, failure::Error> {
-    let pattern = &user.settings.clone().project.route.expect("⚠️ Your project config has an error, check your `wrangler.toml`: `route` must be provided.");
+    let pattern = &user.settings.clone().project.route.expect(
+        "⚠️ Your project config has an error, check your `wrangler.toml`: `route` must be provided.",
+    );
```
steveklabnik: I'm not sure how this change got in here, other than a change to `cargo fmt`?

ashleygwilliams: could have been! i'll keep an eye out for that

Issue #69, opened 2019-04-16
RFC: leverage Webpack as backend


xtuc:
## Problem

Currently `wrangler` is coupled to wasm-pack, which makes it unpracticable for non-rust projects (especially JavaScript projects).

## Proposal

Use Webpack as a backend to build any projects that use a Webpack configuration.

For prototyping, I made `wranger-js` and came up with a few demos https://github.com/xtuc/wrangler-js-demo to give you a better idea.

Note that we don't want to have multiple tools but rather add that support in `wrangler` to have a single tool.

### Context about Webpack

With the integration of WebAssembly in Webpack, any loader can emit a wasm binary in Webpack that will be bundled along the JavaScript project.

That integration already implements the [ECMAScript module integration](https://github.com/WebAssembly/esm-integration), which make working with Wasm module easier in JavaScript.

### Allows npm dependecies

Webpack is a bundler, npm dependecies are supported.

### Optimize JavaScript

Webpack runs, in production mode, a JavaScript minifier. Also we could add Wasm optimization.

### Auto generate the Worker's metadata.json

Webpack constructs the module graph of the project, we can look at it to find any wasm modules and generate the corresponding metadata.json.

ashleygwilliams:
1. what do we expect in the user's env (so we can install it and run it)
2. what are the steps to run
3. what are the templates we run the steps on

xtuc:
I made `wrangler-js` public: https://github.com/xtuc/wrangler-js/blob/master/index.js.

Here is what we use it for:
- combine JavaScript packages: `webpack`
- run Webpack
    - introspect Webpack's module graph for wasm binaries
    - concatenate JavaScript files
    - generate the bindings file
- inject a prologue, that polyfills `fetch` + `window`

Here are the changes for moving the logic in `wrangler`:

I don't think it would be a good idea to send back to `wrangler` the entire Webpack graph because it can get pretty big; we should aggregate the information in wrangler-js and concatenate files I think.
The files could be written from `wrangler`.

The bindings file can be generated in `wranger`.

The prologue is just a string, it's easy to do in `wrangler`.

What do you think @ashleygwilliams ?

PR #75, opened 2019-04-29
Use underlying wrangler-js

xtuc:
refs #69 

I have a private demo: https://github.com/xtuc/wrangler-tmp, please let me know if you want access.

I would appreciate some early feedback.

### DIFF
--- a/src/commands/build.rs
+++ b/src/commands/build.rs
@@ -1,4 +1,5 @@
 use crate::user::settings::ProjectType;
+use crate::wranglerjs;
 use crate::{commands, install};
 use binary_install::Cache;
 use std::path::PathBuf;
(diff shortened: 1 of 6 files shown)


steveklabnik:
It's not entirely clear to me what this PR actually *does*.

That being said, it looks like this adds a dependency on a node environment. One of the advantages of Rust for CLI stuff is that you don't need to have a bunch of stuff installed to use it; regardless of other details, this feels like a regression to me.

xtuc:
@steveklabnik you can find more information at https://github.com/cloudflare/wrangler/wiki/Draft-Webpack-design.

I agree that having a single binary would be simpler, but given that we will depend JavaScript code, I'm not sure what's the best solution.

One thing I have in mind is that I could be opt-in, for Rust based project invoke `wasm-pack` and if you install `xxx` run Webpack based project. What I don't link about that is that the `wasm-pack` workflow could be run using Webpack too, which adds some features and is actually an implementation details from the user's percpective.

xtuc:
I think it would be easier to have two branches, there are too many moving parts to be in master.

PR #77, opened 2019-04-29
Revert "properly quote commands"

ashleygwilliams:


ashleygwilliams:
this PR broke this but it will be easier to patch than revert. closing.

PR #78, opened 2019-04-30
fix(command): unbreak nix

ashleygwilliams:
unbreaking breaking change in #74


steveklabnik:
I can confirm that this PR works on Windows.

Issue #82, opened 2019-05-01
Webpack: introducing JS deps


xtuc:
I realized that the PR might not the best way to track this discussion.

@steveklabnik you commented in https://github.com/cloudflare/wrangler/pull/75#issuecomment-487603482, could you please elaborate? Are you aware of any precedence?

I agree that our JavaScript dependecy should be transparent for our users, apart from requiring nodejs.

It would be simpler for the user to not handle a `node_modules`/`package.json` setup in their project, especially if they are not used to JS.

xtuc:
While we can't get rid of the JavaScript deps, It make sense to use wrangler to control it; ensuring the version and the install.

steveklabnik:
So, there's sorta two main use-cases here. Simple use-cases where you don't use any JavaScript dependencies or bundlers. And more complex cases where you do. With the latter, you can't get away with not including a JS environment, of course. But for cases where you're not using any JavaScript, requiring an install of a completely separate environment is an extra step that's not going to be used.

ashleygwilliams:
we've decided that we are doing this. there are still branches that allow folks to use wrangler without the node env.

xtuc:
If we want to make it fully independent of the user's env, we can still bundle node/npm in our releases. Which isn't that hard.

PR #94, opened 2019-05-21
Use Webpack as backend

### LINKED ISSUE #69: RFC: leverage Webpack as backend

xtuc:
Closes #69


ashleygwilliams:
- have this run npm install
- capture  and stream standard out on wrangler publish
- print a message on wrangler publish to say it will take a while

src/commands/build.rs
```diff
@@ -19,7 +20,27 @@ pub fn build(cache: &Cache, project_type: &ProjectType) -> Result<(), failure::E
 
             commands::run(command, &command_name)?;
         }
+        ProjectType::Webpack => {
+            let wasm_pack_path =
+                install::install("wasm-pack", "rustwasm", cache)?.binary("wasm-pack")?;
+
+            wranglerjs::run_npm_install().expect("could not run `npm install`");
```
nickbalestra: I think  that this assume that the user has `npm` installed and available globally. Should we mention this on the readme or what will be the strategy for updating the docs once this PR gets merged?

src/commands/publish/mod.rs
```diff
@@ -61,6 +61,15 @@ fn publish_script(user: &User, name: &str) -> Result<(), failure::Error> {
                 .body(build_js_script()?)
                 .send()?
         }
+        ProjectType::Webpack => {
```
nickbalestra: Do you see `ProjectType::Webpack` replacing `ProjectType:: JavaScript `? I guess we're keeping both until we reach feature parity (ie for the preview). Was just wondering how you see this moving forward and evolving in the longer term

wrangler-js/index.js
```diff
@@ -0,0 +1,70 @@
...
+  config = require(join(process.cwd(), "./webpack.config.js"));
+}
+
+if (config.target !== "webworker") {
+  if (config.target !== undefined) {
+    console.warn("Unsupported Webpack target; webworker has been set.");
+  }
+  config.target = "webworker";
+}
+const compiler = webpack(config);
```
nickbalestra: Will relying on an in-memory [filesystem for for webpack ](https://webpack.js.org/api/node/#custom-file-systems) allow us not having a `dist_to_clean`?

src/commands/build.rs
xtuc: In order to use it the user must have `node` and `npm` available, see #82

src/commands/publish/mod.rs
xtuc: In discussion #102

wrangler-js/index.js
xtuc: I made #116, that's a good question.

src/wranglerjs/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+use std::fs::File;
+use std::io::prelude::*;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+// This structure represents the communication between {wrangler-js} and
+// {wrangler}. It is send back after {wrangler-js} completion.
+// FIXME(sven): make this private
+#[derive(Deserialize, Debug)]
+pub struct WrangerjsOutput {
```
signalnerve: `s/Wranger/Wrangler`

src/commands/publish/mod.rs
ashleygwilliams: there should always be an option to not use webpack.

src/commands/publish/preview/mod.rs
```diff
@@ -34,6 +34,7 @@ pub fn preview(
             .post(create_address)
             .body(publish::build_js_script()?)
             .send(),
+        ProjectType::Webpack => panic!("unimplemented"),
```
ashleygwilliams: this should be implemented ;)

src/user/settings.rs
```diff
@@ -14,10 +14,14 @@ pub struct GlobalUserSettings {
     pub api_key: String,
 }
 
+fn default_project_type() -> ProjectType {
```
ashleygwilliams: this function should really be an implementation of the Default trait: https://doc.rust-lang.org/std/default/trait.Default.html

wrangler-js/index.js
```diff
@@ -0,0 +1,88 @@
...
+  }, "");
+
+  if (hasWasmModule === true) {
+    bundle.wasm = Buffer.from(assets[wasmModuleAsset].source()).toString();
+    bundle.wasm_name = wasmModuleAsset;
+  }
+
+  writeFileSync(args["output-file"], JSON.stringify(bundle));
+
+  // FIXME(sven): stats could be printed in {wrangler}, avoiding any confusion.
```
ashleygwilliams: can we get this stats printing in wrangler before merge?

wrangler-js/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "wrangler-js",
+  "version": "0.1.9",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
ashleygwilliams: since there's no scripts we can just remove this from the package json ;)

src/commands/build.rs
ashleygwilliams: we need this to check that node and npm are present. see how we check using the `which` crate here: https://github.com/cloudflare/wrangler/blob/85f09e0d7b12971a224ba683fcd230ef31509cf4/src/install/mod.rs#L10

instead of install node and npm we'll need to panic and tell the user they need to install node and npm OR to run with whatever we decide is the way to override to the ProjectType::JavaScript

ashleygwilliams: i've added this functionality.

src/user/settings.rs
xtuc: Serde seem to support better default now too https://serde.rs/attr-default.html, but I wans't able to make them work. I will look into the Default trait.

PR #99, opened 2019-05-21
Use main in pkg json for publishing and preview single file js projects

ashleygwilliams:


src/commands/publish/mod.rs
```diff
@@ -73,11 +76,12 @@ fn multi_script(user: &User, name: &str) -> Result<(), failure::Error> {
     Ok(())
 }
 
-fn build_js() -> Result<String, failure::Error> {
-    Ok(fs::read_to_string("worker.js")?)
+fn build_js_script() -> Result<String, failure::Error> {
+    let package= Package::new("./")?;
```
xtuc: I think you will need to run fmt here.

ashleygwilliams: good call. pushed up changes.

PR #105, opened 2019-05-21
chore: emoji fallbacks

### LINKED ISSUE #34: emoji fallback
not all terminal support emoji. there is a library to wrap this:
https://github.com/rustwasm/wasm-pack/blob/master/src/emoji.rs
create a file like this for wrangler and then replace instances of emoji with constants from that file

use wasm-pack as an example. the implementation should basically be Exactly the Same

ashleymichal:
fixes #34

### DIFF
--- a/src/commands/generate.rs
+++ b/src/commands/generate.rs
@@ -24,7 +24,7 @@ pub fn generate(name: &str, template: &str, cache: &Cache) -> Result<(), failure
 fn command(name: &str, binary_path: PathBuf, args: &[&str], project_type: &ProjectType) -> Command {
     println!(
         "{} Generating a new {} worker project with name '{}'...",
-        emoji::GOAT,
+        emoji::SHEEP,
         project_type,
         name
     );
--- a/src/commands/generate.rs
+++ b/src/commands/generate.rs
@@ -4,6 +4,8 @@ use binary_install::Cache;
 use std::path::PathBuf;
 use std::process::Command;
 
+use crate::emoji;
+
 pub fn generate(name: &str, template: &str, cache: &Cache) -> Result<(), failure::Error> {
     let tool_name = "cargo-generate";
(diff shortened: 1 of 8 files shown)


xtuc:
Thanks for your PR! Emoji in the source are hard to work with IMO.

src/commands/generate.rs
```diff
@@ -21,8 +23,10 @@ pub fn generate(name: &str, template: &str, cache: &Cache) -> Result<(), failure
 
 fn command(name: &str, binary_path: PathBuf, args: &[&str], project_type: &ProjectType) -> Command {
     println!(
-        "🐑 Generating a new {} worker project with name '{}'...",
-        project_type, name
+        "{} Generating a new {} worker project with name '{}'...",
+        emoji::GOAT,
```
ashleygwilliams: GOAT? lol

src/main.rs
```diff
@@ -17,12 +18,12 @@ fn main() -> Result<(), failure::Error> {
...
     let cache = get_wrangler_cache()?;
 
-    let matches = App::new("👷‍♀️🧡☁️ ✨ wrangler")
+    let matches = App::new(format!("{}{} wrangler", emoji::WORKER, emoji::SPARKLES))
         .version(env!("CARGO_PKG_VERSION"))
         .author("ashley g williams <ashley666ashley@gmail.com>")
         .subcommand(
             SubCommand::with_name("generate")
-                .about("👯 Generate a new wasm worker project")
+                .about(&*format!("{} Generate a new wasm worker project", emoji::DANCERS))
```
ashleygwilliams: is this `&*` necessary?

ashleygwilliams:
to install and run rustfmt:

1. ` rustup component add rustfmt`
2. `cargo fmt` 

you don't have clippy errors but you'll also probably want to have this set up
1. `rustup component add clippy`
2. `cargo clippy`

Issue #111, opened 2019-05-22
Webpack: send stats to wrangler and display more meaningful stats


xtuc:
See wrangler-js/index.js.

EverlastingBugstopper:
Looks like this has been fixed with #114

PR #114, opened 2019-05-22
Show bundle size

xtuc:
Closes https://github.com/cloudflare/wrangler/issues/111


xtuc:
This was a try, the goal was to send Webpack stats to `wrangler`, like sizes but also a idea of your module graph. But I doubt this is useful now and the current size logic could be entirely in `wrangler`.

PR #120, opened 2019-05-23
Port wasm-pack docs site

steveklabnik:
image

Some notes:

1. removed the book entirely since there's no wrangler book. not sure if you'd still want a link to some sort of docs, but I can add it
2. logo is 404 until one comes, @ashleygwilliams said this was okay for now
3. i am 99% sure i got everything, but still should be double checked.
4. I am not sure if you want changes to CI; if so i can add them in too by also looking at wasm-pack


docs/_installer/index.html
```diff
@@ -0,0 +1,123 @@
+<!DOCTYPE html>
+<head>
+  <meta charset="utf-8">
+  <title>wasm-pack</title>
+  <link href="https://cdnjs.cloudflare.com/ajax/libs/skeleton/2.0.4/skeleton.min.css" rel="stylesheet">
```
ashleygwilliams: need to remove this

docs/_installer/index.html
```diff
@@ -0,0 +1,123 @@
...
+  <link rel="stylesheet" href="../public/custom.css"/>
```
ashleygwilliams: need to remove this

steveklabnik:
github is being weird, https://github.com/cloudflare/wrangler/pull/120/commits/3637f471a762c7c8844f71a60da9d29b5db3c5c0 removed most of the head stuff, so i marked those as resolved

steveklabnik:
Updated!

steveklabnik:
I've confirmed that, at least, wrangler still works after these changes 😅 

I'm still not 100% sure how to best test the install capability itself. Hm.

steveklabnik:
So, when building the releases, you need to do this:

https://github.com/rustwasm/wasm-pack/blob/master/.appveyor.yml#L26

but if you do, the installer does work.

PR #131, opened 2019-05-24
Test this thing2

xtuc:


xtuc:
I incorrectly included commits from master, sorry for the diff.

@ashleygwilliams what do you think?

PR #143, opened 2019-05-25
Show command help when given no arguments

Electroid:
Running just `wrangler` with no arguments will show the command help
(same output as `wrangler help`).


ashleygwilliams:
i'm going to need to fix the tests on this ;) i'll merge this as is- but in case you wonder how i fix it i'll mention you in my follow up PR so you can see. the testing strategy on this project is less than great for sure- so apologies for the state of things!

PR #149, opened 2019-05-26
Steveklabnik gh140

ashleygwilliams:


src/commands/build/webpack/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+pub fn install() -> Result<(), failure::Error> {
+    for tool in &["node", "npm"] {
+        env_dep_installed(tool)?;
+    }
+
+    if which::which("wrangler-js").is_err() {
+        let mut command = build_npm_command();
+        command
+            .arg("install")
+            .arg("https://github.com/ashleygwilliams/wrangler-js")
```
xtuc: I'm not sure about that. The package is published under `wrangler-js` and it seems to requires root now. 

I was working on the global install, I can submit it next week.

ashleygwilliams:
closing as irrelevant

Issue #92, opened 2019-05-15
`wrangler deploy` un-does KV namespace bindings


steveklabnik:
In three screenshots:

![image](https://user-images.githubusercontent.com/27786/57802941-c6d2bc80-771c-11e9-9e5d-c42aa7e4a6c4.png)

![image](https://user-images.githubusercontent.com/27786/57802950-cfc38e00-771c-11e9-8ef7-95d0231bdee7.png)

![image](https://user-images.githubusercontent.com/27786/57802973-d94cf600-771c-11e9-8bf9-1dc6be6a18e3.png)

xtuc:
Ideally we would use JavaScript Module syntax to detect what KV things needs to be declared:
```js
import posts from "cloudflare:kv/posts".
```

Alternatively, we could fetch the current bindings, merge it with the local version and publish.

kreitlowbryan:
It looks like an update was merged but it doesn't seem to honor the addition of `kv_namespaces = ["NAMESPACE"]` in the wrangler.toml file. Am I missing something?

xtuc:
The change will create a kv-namespace when it doesnt already exists during the publishing. However, the bug here is that they are not defined in the Worker bindings.

kreitlowbryan:
There is currently no support to add Worker KV bindings to the wrangler.toml file to have those bindings generated when the metadata.json is built as part of the `wrangler publish` command? It is really painful to try and dev using worker KV with wranger when these bindings are removed upon every publish.

xtuc:
We are aware, see https://github.com/cloudflare/wrangler/issues/215

grempe:
For those who find this. The current process to workaround that seems to be working for me now is:

* Deploy script with `wrangler publish` (destroys any currently configured bindings)
* Load CF Worker Editor Web UI to edit your worker.
* Click the resources tab for the worker and RE-ADD the binding between variable and namespace.
* Click the code tab AGAIN and `DEPLOY` the script you previously published with wrangler AGAIN by clicking the button.

Next time you do `wrangler publish` or `wrangler publish --release` you'll need to do this again.

Hopefully this gets cleanly resolved soon. /smh

ashleymichal:
closed by #334 . Being released as part of 1.1.0.

PR #160, opened 2019-05-27
initial kv namespace support

steveklabnik:
* set your namespaces in wrangler.toml
* attempt to create them upon publish
* If they exist, don't worry about it

So, this doesn't address #92 yet, because you need to modify your worker's `worker/metadata_wasm.json`. However, since we're mostly doing this to get docs upload working, rather than KV as a general feature, it's good to go.

EDIT: ignore the rest of the discussion on this PR, as it's now invalid.

### DIFF
--- a/src/commands/publish/mod.rs
+++ b/src/commands/publish/mod.rs
@@ -46,7 +46,7 @@ pub fn create_kv_namespaces(user: &GlobalUser, project: &Project) -> Result<(), 
 
     if let Some(namespaces) = &project.kv_namespaces {
         for namespace in namespaces {
-            info!("Attempting to create namesapce '{}'", namespace);
+            info!("Attempting to create namespace '{}'", namespace);
 
             let mut map = HashMap::new();
             map.insert("title", namespace);
--- a/src/commands/publish/mod.rs
+++ b/src/commands/publish/mod.rs
@@ -9,6 +9,7 @@ use package::Package;
 use log::info;
 
 use reqwest::multipart::Form;
+use std::collections::HashMap;
 use std::fs;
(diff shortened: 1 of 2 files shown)


ashleygwilliams:
you can test this by publishing to a diff domain.... or give the script a new name and put it at a different route

steveklabnik:
that seems to make it not work

steveklabnik:
Okay, so an update: this only works if you modify your worker's `worker/metadata_wasm.json`. However, since we're mostly doing this to get docs upload working, rather than KV as a general feature, it's good to go.

src/commands/publish/mod.rs
```diff
@@ -34,6 +36,48 @@ pub fn publish(user: &GlobalUser, project: &Project, release: bool) -> Result<()
...
+    let kv_addr = format!(
+        "https://api.cloudflare.com/client/v4/accounts/{}/storage/kv/namespaces",
+        project.account_id,
+    );
+
+    let client = reqwest::Client::new();
+
+    if let Some(namespaces) = &project.kv_namespaces {
+        for namespace in namespaces {
+            info!("Attempting to create namesapce '{}'", namespace);
```
xtuc:
```suggestion
            info!("Attempting to create namespace '{}'", namespace);
```

steveklabnik: how embarrassing, thanks!

steveklabnik:
> With Webpack path there could be an automatic way for declaring KV-bindings (as described in my ESM Worker proposal).

Yep! That won't work for non-webpack stuff though, so we're still gonna need it in some capacity.

PR #171, opened 2019-05-28
Upload static content

steveklabnik:
This adds a new command to wrangler:

```command
> wrangler upload-static-files namespace directory
```

it will:

1. walk through `directory`, and get the list of files
2. upload each of those files to `namespace`, with the filename + path as the key, and the contents of the file as the value

This means that we can really easily use workers + kv as a static site host; run this command as part of the deploy process.

I've tested this with my personal website, but I bet some things are a bit sloppy; please review!


xtuc:
Interesting! @nickbalestra would probably be interested by that (he built something similar).

I think the bulk call will be required very soon, people with big websites will hit the rate limit I think.

One thing I don't like about static website in a worker is that other Cloudflare's transformations (minification, image optimizations, etc) won't apply (as far as I can tell).

src/commands/mod.rs
```diff
@@ -7,6 +7,7 @@ pub mod config;
 pub mod generate;
 pub mod publish;
 pub mod subdomain;
+pub mod upload_static_files;
```
ashleygwilliams: we should not have this as a separate mod- let's put this into publish

src/main.rs
```diff
@@ -122,6 +122,23 @@ fn main() -> Result<(), failure::Error> {
             "{} Retrieve your user info and test your auth config",
             emoji::SLEUTH
         )))
+        .subcommand(
```
ashleygwilliams: this should be part of publish, either by default and/or a --static-files flag

ashleygwilliams:
preferred interface here:

```
wrangler publish --release --static name_space assets_dir
```

open to change the flag name to `data`, or ... whatever, but i think a flag off of publish is good for now

PR #163, opened 2019-05-28
refactor wranglerjs installation

xtuc:
Downloads a wranglerjs release from GitHub and install it in {Wrangler}'s cache;
globally.

Closes https://github.com/cloudflare/wrangler/issues/133

This change also contains a few {wranglerjs} renames.

PR #173, opened 2019-05-29
Refactor wranglerjs installation (part 1)

xtuc:
Removed the build step from https://github.com/cloudflare/wrangler/pull/163.

Merging it as is will break because no wranglerjs releases exist now, but #174 fixes it.


Cargo.toml
```diff
@@ -10,7 +10,7 @@ categories = ["wasm"]
 
 [dependencies]
 atty = "0.2.11"
-binary-install = "0.0.2"
+binary-install = { git = "https://github.com/xtuc/binary-install.git", branch = "feat-download-artifact" }
```
xtuc: rustwasm/binary-install#1

PR #178, opened 2019-05-30
Add npm install

xtuc:
This adds an npm package that after install will download wrangler from
the latest GitHub release and provide the binary.


xtuc:
We can merge this, and we should check that we got all the right perms to publish it.

PR #191, opened 2019-06-01
fix: make sure to remove lock

xtuc:


src/wranglerjs/mod.rs
```diff
@@ -197,26 +197,25 @@ pub fn run_npm_install(dir: PathBuf) -> Result<(), failure::Error> {
     // avoid running multiple {npm install} at the same time (eg. in tests)
     flock.lock_exclusive()?;
 
-    if dir.join("node_modules").exists() {
-        info!("skipping npm install because node_modules exists");
-        return Ok(());
-    }
+    if !dir.join("node_modules").exists() {
```
ashleygwilliams: for posterity- we may need to change this logic at somepoint

PR #192, opened 2019-06-01
Check for keys

### LINKED ISSUE #170: Validate non empty zone ID when doing `wrangler publish --release`
If a user tries to run `wrangler publish --release` without a zone ID, they get a fairly opaque message back – we should validate that the zone ID is set before hitting the API at all

```
Error: ErrorMessage { msg: "⛔ There was an error featching your project\'s routes.\n Status Code: 400 Bad Request\n Msg: {"success\":false,\"errors\":[{\"code\":7003,\"message\":\"Could not route to \\/zones\\/workers\\/routes, perhaps your object identifier is invalid?\"},{\"code\":7000,\"message\":\"No route for that URI\"}],\"messages\":[],\"result\":null}" }
```

ashleygwilliams:
fixes #170

### DIFF
--- a/src/commands/publish/mod.rs
+++ b/src/commands/publish/mod.rs
@@ -83,6 +83,9 @@ fn publish_script(
     project: &Project,
     release: bool,
 ) -> Result<(), failure::Error> {
+    if project.account_id.is_empty() {
+        failure::bail!("You must provide an account_id in your wrangler.toml before you publish!")
+    }
     let worker_addr = format!(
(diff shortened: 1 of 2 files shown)


xtuc:
This includes unrelated commits, lgtm otherwise.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "wrangler"
-version = "0.1.1"
+version = "1.0.0-alpha"
```
signalnerve: 🙌

PR #203, opened 2019-06-03
Document `kv-namespaces` in README

signalnerve:
Documents the `kv-namespaces` key/value added in #160 in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,9 @@ There are two types of configuration that `wrangler` uses: global user and per p
         - `*example.com/*`
         - `http://example.com/hello`
         This key is optional if you are using a workers.dev subdomain and is only required for `publish --release`.
+    - `kv-namespaces`: A listing of [Workers KV](https://workers.cloudflare.com/docs/reference/storage/overview/) namespace bindings that you want configured for your project, in uppercase format. _Note: Workers KV is available only to paid Workers users._ Examples:
+        - `[CACHE]`
+        - `[TODOS, USERS]`
 
 ## ⚓ Installation
 


adaptive:
@signalnerve can add an example on the docs?

xtuc:
I agree that an example would be helpful. cc'ing @steveklabnik do we actually want to document it? Or keep it internal?

kreitlowbryan:
It looks like this is part of wrangler v1.0.0 but I am unable to get it to work

steveklabnik:
My understanding when talking with @ashleygwilliams was that she is not sure that she actually *likes* the interface that I added for KV. see some of the comments in https://github.com/cloudflare/wrangler/pull/171, for example

So, at first I'd like to defer to her; is this part something we are certainly keeping? If so, then yes, let's document it and I'll review. But it may change significantly before we actually consider it a public API.

ashleygwilliams:
i'm mostly ok with the wrangler-toml interface, it's mostly the CLI interface that i'd like to change. i need to snag #171 and get it merged and then i think we can merge this. at the moment it doesn't entirely work so it's not urgent. (don't want to doc something that doesn't entirely work ;) )

README.md
```diff
@@ -144,6 +144,9 @@ There are two types of configuration that `wrangler` uses: global user and per p
         - `*example.com/*`
         - `http://example.com/hello`
         This key is optional if you are using a workers.dev subdomain and is only required for `publish --release`.
+    - `kv-namespaces`: A listing of [Workers KV](https://workers.cloudflare.com/docs/reference/storage/overview/) namespace bindings that you want configured for your project, in uppercase format. _Note: Workers KV is available only to paid Workers users._ Examples:
+        - `[CACHE]`
+        - `[TODOS, USERS]`
```
kentonv: Don't you also need to specify the namespace ID?

xtuc: Yes correct, it's in #215. That said, it's currently in discussion I believe.

ashleygwilliams:
closing in favor of the work @ashleymichal is doing on KV work

PR #205, opened 2019-06-03
feat: show sizes

xtuc:
Once the build is finished print the size of the script, in gzip, and
Wasm binary size if available.


xtuc:
Sure thing

Issue #90, opened 2019-05-15
Linting phase


xtuc:
We could:
- warn on filesize
- check for forbidden functions (eval, Function)

For kv-namespaces we could:
- ensure the name is a valid JavaScript identifier (in strict mode).
- unsused kv_namespace bindings

ashleymichal:
we have implemented some of these;  others are already implemented in the api  and so are for sure nice to haves but in the interest of reducing WIP, i am closing  this.

PR #215, opened 2019-06-05
Add kv_namespaces in metadata

### LINKED ISSUE #92: `wrangler deploy` un-does KV namespace bindings

xtuc:
Fixes #92

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,9 @@ There are two types of configuration that `wrangler` uses: global user and per p
         This key is optional if you are using a workers.dev subdomain and is only required for `publish --release`.
     - `webpack_config`: This is the path to the webpack configuration file for your worker. This is optional and
         defaults to `webpack.config.js`
+    - `kv-namespaces`: Bind kv namespaces to your worker. Must be created in the dashboard before. An array of:
+        - `binding`: name that will be used to bind the kv-namespace to your script, in JavaScript.
+        - `id`: Identifer of the namespace (can be found on the Cloudflare dashboard).
 
 ## ⚓ Installation
 
--- a/src/commands/build/wranglerjs/bundle.rs
+++ b/src/commands/build/wranglerjs/bundle.rs
@@ -6,12 +6,13 @@ use std::fs::File;
 use std::io::prelude::*;
 use std::path::{Path, PathBuf};
 
+use crate::terminal::message;
 use log::info;
 
 use crate::commands::build::wranglerjs::output::WranglerjsOutput;
 use crate::settings::binding::Binding;
 use crate::settings::metadata;
-use crate::terminal::message;
+use crate::settings::project::KvNamespace;
 
 // Directory where we should write the {Bundle}. It represents the built
 // artifact.
@@ -34,7 +35,11 @@ impl Bundle {
         Bundle { out }
     }
 
-    pub fn write(&self, wranglerjs_output: &WranglerjsOutput) -> Result<(), failure::Error> {
+    pub fn write(
+        &self,
+        wranglerjs_output: &WranglerjsOutput,
+        kv_namespaces: Vec<KvNamespace>,
+    ) -> Result<(), failure::Error> {
         let bundle_path = Path::new(&self.out);
         if !bundle_path.exists() {
             fs::create_dir(bundle_path)?;
@@ -52,7 +57,7 @@ impl Bundle {
 
         script_file.write_all(script.as_bytes())?;
 
-        let metadata = create_metadata(self).expect("could not create metadata");
+        let metadata = create_metadata(self, &kv_namespaces).expect("could not create metadata");
         let mut metadata_file = File::create(self.metadata_path())?;
         metadata_file.write_all(metadata.as_bytes())?;
 
@@ -112,13 +117,23 @@ pub fn create_prologue() -> String {
 }
 
 // This metadata describe the bindings on the Worker.
-fn create_metadata(bundle: &Bundle) -> Result<String, serde_json::error::Error> {
+fn create_metadata(
+    bundle: &Bundle,
+    kv_namespaces: &Vec<KvNamespace>,
+) -> Result<String, serde_json::error::Error> {
     let mut bindings = vec![];
 
     if bundle.has_wasm() {
         bindings.push(Binding::new_wasm_module(
-            bundle.get_wasm_binding(),
-            bundle.get_wasm_binding(),
+            bundle.get_wasm_binding(), // name
+            bundle.get_wasm_binding(), // part
+        ));
+    }
+
+    for namespace in kv_namespaces {
+        bindings.push(Binding::new_kv_namespace(
+            namespace.binding.clone(), // local_binding
+            namespace.id.clone(),      // namespace_id
         ));
     }
 
@@ -154,7 +169,7 @@ mod tests {
         };
         let bundle = Bundle::new_at(out.clone());
 
-        bundle.write(&wranglerjs_output).unwrap();
+        bundle.write(&wranglerjs_output, vec![]).unwrap();
         assert!(Path::new(&bundle.metadata_path()).exists());
         let contents =
             fs::read_to_string(&bundle.metadata_path()).expect("could not read metadata");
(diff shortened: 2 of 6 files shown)


src/commands/build/mod.rs
```diff
@@ -45,8 +45,13 @@ pub fn build(cache: &Cache, project_type: &ProjectType) -> Result<(), failure::E
                 failure::bail!("Webpack returned an error");
             }
 
+            let mut kv_namespaces = vec![];
+            if let Some(value) = &project.kv_namespaces {
+                kv_namespaces = value.clone();
+            }
```
steveklabnik: This whole block is more idiomatically written as:

    let kv_namespaces = project.kv_namespaces.unwrap_or(Vec::new());

(You may need a `.clone()` before the `.unwrap_or`, I'm not actually 100% sure)

xtuc:
It kinda works, you need to specifiy the namespace_id in your wrangler.toml.

```toml
# ...

[[kv-namespaces]]
local_binding = "local_binding"
name = "name"
namespace_id = "abc"
```

signalnerve:
Re: the `namespace_id` comment - I agree that having to specify it inside of `wrangler.toml` isn't great, but if users are creating the bindings from inside of a Wrangler-enabled project, can we write it to `wrangler.toml` for them on initial creation? 

There may be situations where they'll need to go look up the namespace manually, but I imagine the average workflow will take place (hopefully!) in Wrangler, so we should know the `namespace_id` of a KV store since we're creating it for them during the publishing process.

steveklabnik:
What is currently holding this PR back from landing?

xtuc:
@steveklabnik it would require a couple of rebases but I'm happy to see this landing. If you need it I can continue to work on it.

steveklabnik:
It'd be nice to get it landed so that we can continue the work on KV support overall; if you've got the time that would be great.

ashleymichal:
I think what I'm missing to be able to judge this PR is the README section on how to configure your namespaces in the wrangler.toml; @xtuc touched on it in his comment but what would my wrangler.toml look like if I had two or more namespaces exposed to my worker?

steveklabnik:
@ashleymichal it currently looks like this:

```toml
kv-namespaces = ["STATIC_CONTENT"]
```

if you had two

```toml
kv-namespaces = ["STATIC_CONTENT", "FOOBAR"]

```

ashleymichal:
@steveklabnik based on the earlier comment from @xtuc [here](https://github.com/cloudflare/wrangler/pull/215#issuecomment-500426900), that does not appear to be the case in the context of this PR; the user is required to specify not only the name of the binding as it should appear in the Worker, but the "title" and id of the namespace, in order to both create the namespace (if it does not already exist), and specify the binding in the metadata.

The question still stands, as once we provide any interface to users, it becomes difficult to change. In order to land this, we need to care about a few things:

1. Providing enough information to allow Wrangler to provision a KV namespace on behalf of a user OR requiring the user to provision the namespace before integrating with their Worker project.
2. Providing enough information for Wrangler to generate a KV Binding entry for the metadata part of the upload form. This can either be provided by the user manually (asking them to enter their namespace id manually as shown in @xtuc 's comment), or it can be retrieved by Wrangler behind the scenes with a query to the KV API based on the namespace title.
3. Providing configuration of the relationship between the binding variable name and the namespace title. This can be done as it stands currently (by assuming they are one and the same and requiring they adhere to JavaScript variable name standards), or it can be done more manually as specified in @xtuc 's comment.

I think 1. is well handled here and lays a decent enough foundation for later improvements and features.

In the case of 2 and 3, I think it would be short-sighted to go with the seemingly "simpler" approach of requiring the user to only provide the namespace title. It is much easier in the long term to make those three connections explicitly and provide "sane defaults" later than to do the reverse. We should therefore take the time to consider what the entries should look like in the wrangler.toml now.

ashleygwilliams:
config wise- re multiple kv name spaces i'd probably do something like this... keying on name

```
[kv-namespaces]
"name1": {"local_binding": "thing", "id" : "aksdhjfkahg;a"}

[kv-namespaces.name1]
local_binding = "thing"
id = "aksdhjfkahg;a"
```

ashleygwilliams:
it should also be noted this is only for the webpack project type which is a deliberate decision we should make

steveklabnik:
Ah, see when I implemented it, it used the title of the namespace and made a binding of the same name; we decided not to support the case where the two differ.

xtuc:
Maybe we could iterate on this PR, to get something useful I would:
- remove the auto creation of namespaces (because we don't store the id). The user/we should do it from the UI.
- let the namespace_id, name and binding name be in the wrangler.toml

This will unlock our deployement and provide a base used for future iterations. We shouldn't document it for now tho. What do you think?

cwndrws:
I agree that the KV config should be in `wrangler.toml`. I like the idea of requiring all three values: (title, id, and binding) like @ashleymichal suggested. I think requiring title be a valid JS symbol would bite us eventually.

src/commands/build/wranglerjs/metadata.rs
```diff
@@ -0,0 +1,16 @@
+use serde::Serialize;
+
+#[derive(Serialize, Debug)]
+pub struct Bindings {
```
cwndrws: We should make this generic and move out of webpack specific project type. Each binding type has different configuration values. This is what I was referring to in #273. We don't need to add all possible binding types, but make it possible add them in the future and make it possible to use the same binding type for both KV namespaces and wasm modules.

ashleymichal: i need to brush up on my rust type system, but i think maybe binding is a trait... #hottake. gonna write this into the design doc i think.

ashleymichal: just to update, @cwndrws clarified that yes that's exactly what he was saying.

xtuc:
I made an update, mainly based on https://github.com/cloudflare/wrangler/pull/215#issuecomment-505381101.

@cwndrws comments:
> I agree that the KV config should be in wrangler.toml. I like the idea of requiring all three values: (title, id, and binding) like @ashleymichal suggested.

The kv-namespace configuration is only possible in `wrangler.toml`; I removed the auto-creation feature (for now).

The config looks like the following:
```toml
[[kv-namespaces]]
name = "KV_NAMESPACE_A"
id = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
``` 

I'm not sure we should ask the user for the namespace name because we won't use it, the API doesn't support it. It would be more confusing if these names get out-of-sync and it would still work, until we add a check which would be a breaking change.

> I think requiring title be a valid JS symbol would bite us eventually.

I like that idea actually, I updated https://github.com/cloudflare/wrangler/issues/90 and we can discuss there.

cwndrws:
@xtuc 

> I'm not sure we should ask the user for the namespace name because we won't use it, the API doesn't support it. It would be more confusing if these names get out-of-sync and it would still work, until we add a check which would be a breaking change.

I agree with not using the namespace name. 

Your example config:
```toml
[[kv-namespaces]]
name = "KV_NAMESPACE_A"
id = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
```
doesn't contain a binding symbol. Should it be:
```toml
[[kv-namespaces]]
name = "KV_NAMESPACE_A"
binding = "KV_NAMESPACE_A"
id = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
```
Or something similar?

xtuc:
@cwndrws the name is the binding, I'm happy to rename it if it's confusing to you.

xtuc:
Blocked by #285

ashleymichal:
> @cwndrws the name is the binding, I'm happy to rename it if it's confusing to you.

Please reference the design document:
https://docs.google.com/a/cloudflare.com/document/d/1Pe2xymJ5dvyNhFB5eJszqTGq70t3pemYm85YEjkEe44/edit?disco=AAAADNBj3nY

If there are disagreements on this syntax, please add them there.

ashleymichal:
upon further reflection, i actually do think that long term this is the right way to specify these namespaces. the wrangler.toml for a workers project should regard kv namespaces as an external dependency, and the configuration of that dependency is a separate issue.

xtuc:
I noticed that when the namespace id is wrong, the publishing API returns a unclear error (generic 400, workers.api.error.internal_server); it would be great to change it.

@ashleymichal I did the updated.

xtuc:
Merging into feat-better-kv

juancampuzano:
Is it possible to make kv-namespaces in a production project that uses wrangler or is it a feature in development?

Currently I do not see the documentation of this in the master.

PR #217, opened 2019-06-05
Use Metadata struct

xtuc:


xtuc:
other build paths doesn't generate the JSON binding?

ashleygwilliams:
@xtuc - the RustWasm one does- the javascript one does not!

xtuc:
Actually I don't think we generate it in the RustWasm part, see https://github.com/cloudflare/wrangler/blob/47dd18dd533817caae927c64dbf3360a2c2d5fb9/src/commands/publish/mod.rs#L198.

ashleymichal:
Ultimately, we want to provide the same KV integration experience in all project types. This means that we'll have to generate the bindings entries for those unless we ask people to configure in a metadata.json, which would be decidedly crappy experience (here's your config. and your other config).

I don't think the hard-coded, FS-accessed metadata_wasm.json is a good long-term solution. The name and "part" (whatever that means, I guess the path to the compiled .wasm file?) are things that should be either configured in your wrangler.toml, or following a convention. I'm inclined to put it in the toml, and have the template(s) updated accordingly, but i'm open to objections on this one.

To me, this suggests that this PR is incomplete: we should generate metadata for all multipart workers. Either this PR needs to be updated to address rustwasm projects, or it needs to be followed up by a second PR that does. Both of those should be in place before landing #215.

xtuc:
Closes in favor of https://github.com/cloudflare/wrangler/compare/refactor-use-metadata-struct2?expand=1

PR #221, opened 2019-06-06
Follow operating system config location standards

### LINKED ISSUE #180: Follow operating system config location standards
It would be great if Wrangler followed the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). Instead of placing the config directory `.wrangler` in the home directory it should be placed in the OS specific appropriate directory. This, among other things, allows a user to specify a directory in which they want all applications to look for their user configurable setting files and keep their home directory clean.

Since this project already uses [dirs](https://crates.io/crates/dirs) it would use `dirs::config_dir();` as a default and fall back to the home directory so this won't be a major version change.

Siilwyn:
Write configuration to a operating system config directory instead of
the home directory according to the XDG Base Directory Specification.

Closes #180

### DIFF
--- a/src/settings/global_user.rs
+++ b/src/settings/global_user.rs
@@ -1,5 +1,6 @@
 use config::{Config, Environment, File};
 use serde::{Deserialize, Serialize};
+use std::path::PathBuf;
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct GlobalUser {
@@ -11,18 +12,37 @@ impl GlobalUser {
     pub fn new() -> Result<Self, failure::Error> {
         get_global_config()
     }
+    pub fn config_directory() -> Option<PathBuf> {
+        get_global_config_directory()
+    }
+}
+
+fn get_global_config_directory() -> Option<PathBuf> {
+    let deprecated_directory = dirs::home_dir()
+        .expect("oops no home dir")
+        .join(".wrangler")
+        .join("config");
+
+    let directory;
+    if deprecated_directory.exists() {
+        directory = Some(deprecated_directory);
+    } else {
+        directory = dirs::config_dir().map(|p| p.join("wrangler"));
+    }
+
+    directory
 }
 
 fn get_global_config() -> Result<GlobalUser, failure::Error> {
     let mut s = Config::new();
 
-    let config_path = dirs::home_dir()
-        .expect("oops no home dir")
-        .join(".wrangler/config/default");
-    let config_str = config_path
+    let config_str = get_global_config_directory()
+        .expect("oops no config dir")
+        .join("default")
         .to_str()
-        .expect("global config path should be a string");
-    s.merge(File::with_name(config_str))?;
+        .expect("global config path should be a string")
+        .to_owned();
+    s.merge(File::with_name(&config_str))?;
 
     // Eg.. `CF_ACCOUNT_AUTH_KEY=farts` would set the `account_auth_key` key
     s.merge(Environment::with_prefix("CF"))?;
--- a/src/settings/global_user.rs
+++ b/src/settings/global_user.rs
@@ -20,7 +20,7 @@ impl GlobalUser {
 
 fn get_global_config_directory() -> Option<PathBuf> {
     let legacy_directory = dirs::home_dir()
-        .expect("oops no home dir")
+        .unwrap_or(PathBuf::from("/"))
         .join(".wrangler")
         .join("config");
 
--- a/src/settings/global_user.rs
+++ b/src/settings/global_user.rs
@@ -1,6 +1,7 @@
 use config::{Config, Environment, File};
 use serde::{Deserialize, Serialize};
 use std::path::PathBuf;
+use log::info;
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct GlobalUser {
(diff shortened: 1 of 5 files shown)


src/settings/global_user.rs
```diff
@@ -11,18 +12,37 @@ impl GlobalUser {
     pub fn new() -> Result<Self, failure::Error> {
         get_global_config()
     }
+    pub fn config_directory() -> Option<PathBuf> {
+        get_global_config_directory()
+    }
+}
+
+fn get_global_config_directory() -> Option<PathBuf> {
```
Siilwyn: I created this function because:
a) the logic for getting the config dir was duplicated in `commands/config.rs`, this way I can re-use it from one place.
b) it has some logic of it's own, doing the whole `exists` dance though in the future with a major version bump it will be a lot simpler I think returning an `Option` is nice so that the config command can print a pretty error.

src/settings/global_user.rs
```diff
@@ -11,18 +12,37 @@ impl GlobalUser {
...
+    let deprecated_directory = dirs::home_dir()
```
xtuc: Instead of deprecated could we rename it to legacy?

src/settings/global_user.rs
```diff
@@ -11,18 +12,37 @@ impl GlobalUser {
...
+        .expect("oops no home dir")
+        .join(".wrangler")
+        .join("config");
+
+    let directory;
+    if deprecated_directory.exists() {
+        directory = Some(deprecated_directory);
+    } else {
+        directory = dirs::config_dir().map(|p| p.join("wrangler"));
```
xtuc: For debbuing it would be useful to add `info!` in these branches. Something like:
```rust
        info!("using legacy directory {:?}", directory);
```
and
```rust
        info!("using xdg directory {:?}", directory);
```

src/settings/global_user.rs
```diff
@@ -11,18 +12,37 @@ impl GlobalUser {
...
+    pub fn config_directory() -> Option<PathBuf> {
+        get_global_config_directory()
+    }
+}
+
+fn get_global_config_directory() -> Option<PathBuf> {
+    let deprecated_directory = dirs::home_dir()
+        .expect("oops no home dir")
+        .join(".wrangler")
+        .join("config");
```
xtuc: There's a slight logic issue here; user with no HOME set but XDG_* prefs will throw here.

Siilwyn: Yes I realized that, since it's such a niche case I was thinking it won't be worth the complexity. What do you think about returning root?

Siilwyn:
Sorry about that, forgot to run `cargo fmt`, will rebase.

Siilwyn:
In the CI the cargo generate command fails, do I need to change something?

src/settings/global_user.rs
xtuc: > What do you think about returning root?
Can you please clarify?

xtuc:
I don't understand the failure in the CI, maybe some package changed version.

Siilwyn:
I've triggered the CI again but now it fails at the next command, I can't reproduce it locally `cargo test --locked` passes for me. :thinking:

xtuc:
I will take a look at it, It's likely not releated to your change.

src/settings/global_user.rs
Siilwyn: I mean returning '/', if there is no home directory.

xtuc: I doubt someone will configure Wrangler in `/`. XDG fallback to the home dir, I think that we should use the same logic.

Siilwyn: But the legacy directory will never be present in the XDG fallback right? Could you clarify what you expect?

xtuc: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html:

> If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used.

My understanding is that:
1. legacy config exists at `$HOME/.wrangler/config`; use it
1. XDG_* is specified; use it
1. XDG_* is empty, use fallback $HOME/.config/wrangler

Does it make sense to you?

xtuc:
I'm going to commit a fix on master, CI is green is https://github.com/cloudflare/wrangler/pull/226.

src/settings/global_user.rs
Siilwyn: To clarify with point 2 expected would be $XDG_CONFIG_HOME/wrangler.
Yes that makes sense, since `dirs::config_dir` already follows the logic of 2 and 3. And first we check the legacy config the current code matches your description right?

Siilwyn: The unwrapping of the legacy location is a bit odd but since in the old code there was no fallback but the code would just exit. But since we still want to continue when the location is not found even when no home dir exists I am not sure how to handle this.

xtuc: Excellent (should have read the doc) and your point is right about the current resolution being broken.

xtuc:
As you pointed out, I don't think maintaining the current (named legacy here) configuration is possible. If would suggest introducing a breaking change (the user would have to `wrangler config` again), such that the new resolution uses XDG_*, fallback to $HOME and throw if impossible to configure.

Siilwyn:
:heavy_check_mark:  Rebased!
The current code is working without a breaking change, unless changing from exiting when no home directory is found to using the new config location is considered a major breaking change.

Siilwyn:
For the npm code I would recommend using https://github.com/sindresorhus/env-paths.

Siilwyn:
@xtuc I have changed the npm code too, though I don't really understand how this is related to the Rust side. It seems to download a wrangler release but that doesn't belong in the config directory. Could you explain?

Siilwyn:
Rebased the branch, should be up-to-date with master. Would like a review @xtuc =)

Siilwyn:
What's up with this PR? @ashleygwilliams

swisster4:
Wheres the money you stole from. The government

Issue #220, opened 2019-06-06
RFC: introducing environments


xtuc:
## Problem

As described https://github.com/cloudflare/wrangler/issues/199, we should have a way to configure Wrangler based on an environment name.

### New command line argument

I'm suggesting to add an `--env NAME` argument for `wrangler publish` and `wrangler build`. Additionally we could use an environment variable.

The build commande is included because a custom webpack configuration could take advantage of it to provide a specific configuration. However, to be able to access it we need to inject an environment variable called `WRANGLER_ENV`.

### Segmented configuration

I think that the entire configuration needs to be segmented by env, like the following example:

```toml
[env.prod]
name = "wasm-worker-prod"
zone_id = "zone-prod"
account_id = "account-prod"
route = "https://prod.site"
kv-namespaces = ...

[env.test]
name = "wasm-worker-test"
zone_id = "zone-test"
account_id = "account-test"
route = "https://test.site"
kv-namespaces = ...
```

Some people expressed concern about `kv-namespaces` being common to all environment, using that kind of configuration it can be different in each environment while maintaining the same JavaScript interface.

## Why arbitrary env?

I doubt we know everyone's workflow, we should let the user make Wrangler fit to their needs.

While we can't indentify which env is for prod (optimized) and dev (debug) I don't think it's an issue; Cloudflare's Worker is just a compilation target I think we should build for prod all the time. For debugging we could take advantage of source maps (or Wasm equivalent, if any).

gabbifish:
Instead of `kv-namespaces`, could we even more broadly allow people to specify production vs staging/test bindings? This could be useful for testing other binding changes (e.g. WASM) in staging before moving to production.

victoriabernard92:
This is a must! What about incorporating this in `wrangler config` instead of in a very long toml?
Workflow to set up would be 
```
wrangler config global <api-key> <email>
wrangler config zone -name v.workers.dev -zone_tag <tag> -account_tag <tag> #zone_tag optional
wrangler config env -name dev -zone v.workers.dev -type webpack -route "ex.com/*" - #ideally this would be a prompt that generates these lines
```
workflow to use the config
```
wrangler init -env dev
#or
wrangler publish -env dev
```
I know this is more work, but. This would prevent users from having to copy/paste their wrangler.toml everywhere and then getting your account and zone tags becomes part of the once per user task versus a once per project

ashleymichal:
[...]

Issue #213, opened 2019-06-04
wrangler.toml routes example


adaptive:
I am having trouble setting routes to
`pub routes: Option<HashMap<String, String>>`

can you show/add to docs example of routes for wrangler.toml? 👏

AaronO:
@adaptive The current `master` does not support publishing multiple routes, I've fixed it here: https://github.com/cloudflare/wrangler/pull/224

Otherwise here's how you define them in your `wranger.toml` (tested locally with my PR's changes):

```toml
[routes]
"sub1.example.com/*" = "worker-script-name" # Should match script variable most of the times
"sub1.example.com/except/*" = "" # Don't route this to any worker
"sub2.example.com/*" = "worker-script-name" # Another subdomain
```

xtuc:
Is there a reason why we need to specifiy the name of the script each time? If we implement #220 that's going to be really error-prone.

If we see the routes as a firewall, I would like to express something like that:
```
[routes]
accept = "sub1.example.com/*"
deny = ["sub1.example.com/wp-admin/*", "sub1.example.com/favicon.ico"]
```

We could even add extra pre-processing on the routes to make it easier:
```
[routes]
accept = "sub1.example.com/{a,b,c}/*"
```

AaronO:
@xtuc I thought about a whitelist/blacklist approach too, I think your suggestion makes sense 👍 

I think in `toml` you could write it as:

```toml
[routes.accept]
"sub1.example.com/*"

[routes.deny]
"sub1.example.com/wp-admin/*"
"sub1.example.com/favicon.ico"
```

The whitelist/blacklist config is not only be less error-prone, but it also prevents leaking API implementation details to the user.

It also solves another issue that bothered me: it felt awkward defining `routes` that are zone-level resources in a `wrangler.toml` file that is meant (for now)to be scoped to a single `script`.

There's a broader question of wrangler's scope (that you breach in #220), i.e: is a single `wrangler` project meant to manage:
a. A single worker
b. Multiple workers in the same zone
c. All the workers in a zone
d. Multiple workers across multiple zones

I think it's important to clarify the scope, sooner rather than later, because it impacts a lot of other design decisions, including #220 (e.g: if we decide that a `wrangler` project is scoped to a single worker, it would make more sense to allow users to specify config files via a `-c, --config` flag instead of bundling multiple workers into the config file and requiring the user to pass a `--env` arg)

EverlastingBugstopper:
This has been released, closing the issue 😄

PR #224, opened 2019-06-06
Support publishing multiple routes

AaronO:
This PR implements 3 major improvements:
1. It enables publishing multiple routes
2. It changes how routes are defined in the config (`wrangler.toml`), to be less error-prone
3. It improves route syncing via diffs (before `wrangler` could only add routes and could not modify or delete routes).

`1.` is pretty straightforward, I'll explain `2.` and `3.` below:

# 2. Routes Config

Currently routes are a `Map<Pattern, ScriptName>`, this has the following shortcomings:
1. It's redundant, for every route we must explicitly specify the target script (or empty string)
2. It's [error-prone](https://github.com/cloudflare/wrangler/issues/213#issuecomment-499808123) (typos may happen when repeating the name)
3. It's got a leaky scope, since a `wrangler.toml` for a script can define routes for other scripts

It was originally built that way because, that's how the API stores them and routes are technically a zone-level resource.

However I think it's cleaner for the end-user if each `wrangler.toml` is scoped to a single script and all resources (such as routes) are viewed as script-resources (even if they're truly zone-level behind the scenes).

Thus, semantically what we want to capture is:
> What routes is this script enabled/disabled on ?

## Before

```toml
[routes]
"sub1.example.com/*" = "worker-script-name"
"sub2.example.com/*" = "worker-script-name"
"sub1.example.com/except/*" = ""
```

## After

```toml
on = [
    "sub1.example.com/*",
    "sub2.example.com/*"
]
off = ["sub1.example.com/except/*"]
```

I considered other alternatives to `on/off` (e.g: `yes/no`, `accept/deny`, `enabled/disabled`), but finally opted for `on/off` because (I think) it strikes the right balance of being:
1. **Familiar** (even for non-native speakers)
2. **Unambiguous** (in this context, on/off doesn't really have another meaningful interpretation)
3. **Short / Simple** (I don't think we need a `routes_` or `routes.` prefix to disambiguate)
4. **"Flat"** (which makes it easier to have multiple script config's in one file, if we do https://github.com/cloudflare/wrangler/issues/220)

# 3. Routes Sync

Prior to this PR, `wrangler` could only create new (single) routes, which had the following shortcomings:
1. You could only have one route (multiple routes are often needed in real-world use-cases)
2. If you specified a route that already existed (pointing to another script), `wrangler` would crash
3. If you changed your route, old routes pointing to this script would not be deleted, so you would have "dangling" routes that could build up over time

## Syncing logic

When syncing routes (for a given script), I think users expect `wrangler` to:
1. **Create** new routes (as defined in `on = [...]`) pointing to this script
2. **Delete** old routes that used to point to this script (but are no longer in the local config)
3. **Ensure disabled routes** (defined in `off = [...]`) point to another script (possibly `""`)
    - **Note:** it was not possible to express this with the previous config approach
4. **Does not overwrite routes** that point to other scripts by default (except `""`), unless `--force` flag is provided

(This is what I'm implementing in this PR)

### Remaining pitfalls

This new sync logic is a significant improvement over what's currently implemented, but it's not perfect. The following pitfalls still exist:
1. Dangling routes to `""` ("null script") are still a problem, since when syncing a given script we can't delete those dangling `""` routes because we don't know (and can't know) if another script requires those routes to be "off"
    * **Note:** If the set of scripts being deployed in this "bundle" is equal to the set of scripts who have routes in production, we could consider our deploy "exhaustive" and thus trim all non-referenced routes (including the aforementioned "dangling null scripts")
2. Some sync operations (e.g: taking a route previously used by another script) still require `--force`
    *  **Note:** If we add "multi-script" support and allow deploying multiple scripts together as a "bundle" (all defined in the same `wrangler.toml`), we could resolve route conflicts (between the scripts in the bundle) and avoid unnecessary `--force` operations

I'm leaving these notes as improvements to be made in the future.

# Tasks

- [x] Support multiple routes, extracted from either `project.route` or `project.routes`
- [x] No longer crash when updating routes (for now show warning that it's not yet implemented)
- [x] Implement diffs
- [x] Change how routes are defined in config (use `on = ['/app/*']` & `off = [ ... ]`)
- [ ] Update & delete routes that can be (that don't belong to another script)
- [ ] Support taking a route used by another script by passing a `--force` flag

# Finishing notes

- [ ] Maybe we could add a `diff` command that shows which `routes` would be changed if we run `publish` ?

### DIFF
--- a/src/commands/publish/mod.rs
+++ b/src/commands/publish/mod.rs
@@ -32,7 +32,7 @@ pub fn publish(user: &GlobalUser, project: &Project, release: bool) -> Result<()
         info!("release mode detected, publishing routes...");
         let routes = ProjectRoutes::new(&project)?;
         routes.publish(&user, &project)?;
-        println!(
+        let msg = format!(
             "✨ Success! Your worker was successfully published. You can view it at:\n - {}\n✨",
             routes.patterns().join("\n - ")
         );
--- a/src/commands/publish/mod.rs
+++ b/src/commands/publish/mod.rs
@@ -1,7 +1,7 @@
 mod krate;
 pub mod preview;
 mod route;
-use route::Route;
+use route::ProjectRoutes;
 
 pub mod package;
 use package::Package;
@@ -29,12 +29,12 @@ pub fn publish(user: &GlobalUser, project: &Project, release: bool) -> Result<()
     create_kv_namespaces(user, &project)?;
     publish_script(&user, &project, release)?;
     if release {
-        info!("release mode detected, making a route...");
-        let route = Route::new(&project)?;
-        Route::publish(&user, &project, &route)?;
-        let msg = format!(
-            "Success! Your worker was successfully published. You can view it at {}.",
-            &route.pattern
+        info!("release mode detected, publishing routes...");
+        let routes = ProjectRoutes::new(&project)?;
+        routes.publish(&user, &project)?;
+        println!(
+            "✨ Success! Your worker was successfully published. You can view it at:\n - {}\n✨",
+            routes.patterns().join("\n - ")
         );
         message::success(&msg);
     } else {
--- a/src/commands/publish/route.rs
+++ b/src/commands/publish/route.rs
@@ -6,7 +6,7 @@ use serde::{Deserialize, Serialize};
 
 use log::info;
 
-#[derive(Deserialize, Serialize)]
+#[derive(Deserialize, Serialize, Debug, Clone)]
 pub struct Route {
     script: Option<String>,
     pub pattern: String,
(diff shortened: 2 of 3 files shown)


stof:
what is the status on this ? Is it expected to be accepted ? I'm also interested in defining several routes for my script.

AaronO:
@stof @ashleygwilliams This is close to being mergeable, I'll provide extra info for the reviewers and wrap up finishing touches in a few hours.

ashleymichal:
This feature is definitely one we want to implement, however there's some design work to be done, and we cannot prioritize it this quarter.

oneeyedziggy:
Looking to use this for an enterprise use-case, and the statement:

> However I think it's cleaner for the end-user if each wrangler.toml is scoped to a single script

is a step in the right direction, but wrangler, right at the top of the project, is billed as a tool to "wrangle your cloudflare workers", plural, not just one worker. So while, as is called out in section 3: "multiple routes are often needed in real-world use-cases", so are multiple scripts. So while I'm just one user, I'd love to see each wrangler.toml scoped to a site rather than a script, but I understand... one step at a time.

In the mean time, is it the current assumption that more complex sites have the choice between writing their own controller to do several wrangler runs, or deploying one mega-worker that contains everything and handles its own routing internally ( or any hybrid that chunks routes into logical groupings and handles some of the routing internally and some through cloudflare's config)?

Philbob:
@ashleymichal - Hi, i'm just wondering is there any update with regards to if this is being worked on  as a feature or not?

ashleymichal:
@Philbob yes! this is on my roadmap for the quarter.

EverlastingBugstopper:
closing in favor of #916 which should be released as part of a beta shortly

AaronO:
@EverlastingBugstopper Do you have an ETA of when the next (NPM) release / version-bump ?

I don't really have a lot of bandwidth to work on this, but I think `wrangler` should have smart sync, similar to what I had outlined here, to avoid route management being a mental "burden" to the end-user, users shouldn't have to manually delete old-routes or do "housekeeping".

PR #225, opened 2019-06-06
Settings: make global config file optional

AaronO:
Right now we crash in `get_global_config()` if the global config file does not exist.

Removing this hard requirement, positions envs and the global config file as 2 alternative credential sources as one would expect.

Relying on a global config file in the user's $HOME makes it awkward for CI, environment variables are usually preferred especially for passing credentials.


AaronO:
> However, if the config doesn't exists on the fs we should make sure it is provided in env, instead of ignoring this misconfiguration.
> 
> I believe that with your change we won't ask the user to run `wrangler config` because it's now optional, which is wrong.

If neither the config file or the envs provide the required values the `.try_into()` part will fail (because the `api_key` and `email` attributes of `GlobalUser` are not optional `Option<...>`): https://github.com/cloudflare/wrangler/blob/3d4fa27329247427f5e2a16addd495bb2d65ae3b/src/settings/global_user.rs#L36-L47

so with this change incomplete settings will show the correct error, hinting to run `wrangler config`.

We could probably improve the error message suggesting to run either `wrangler config` or set `$CF_EMAIL` and `$CF_API_KEY`

AaronO:
@xtuc To sum it up the code already ensures that the config is not incomplete, by `match`ing against the `Result<>` of `.try_into()`.

This change basically ensures that we throw an error only on incomplete settings, whereas prior to this change you would get an error if the global config file doesn't exist, despite providing valid config via ENVs.

xtuc:
Thanks for the clarification. I'm not a fan of the way we merge the config. I think that we have two cases:
- CI, where you can provide config in env
- Desktop, where you can provide the config in a file

However, I will see what the other reviews think about that.

AaronO:
@xtuc Misconfiguration throws with a relatively clear error message (see https://github.com/cloudflare/wrangler/blob/3d4fa27329247427f5e2a16addd495bb2d65ae3b/src/settings/global_user.rs#L41)

What behavior do you expect ?

It would be good to get this merged, because it prevents users from using it in a CI setup (where you don't have an interactive terminal and it would be ugly to artificially generate a config file in CI)

signalnerve:
Thanks for the contribution, @AaronO!

PR #228, opened 2019-06-07
Pin webpack version

xtuc:
Adds a better control over webpack's version, avoiding possible upstream
issues.

Also we depend on one internal.


xtuc:
Maybe that should even be a peerDependecy?

xtuc:
For now I would prefer the security, just pin it.

PR #231, opened 2019-06-07
feature: compare bundle size to limit

### LINKED ISSUE #223: compare size to size limit
since https://github.com/cloudflare/wrangler/pull/205 we now print the size of the Webpack build path worker bundles. we should use this number to compare to size limits and send a warning to let the user know their script size is too big and likely to fail. i don't think we should gate folks that have obviously too big of a worker from publishing.. e.g. potentially wrangler publish --force ...

granjef3:
Closes #223.

### DIFF
--- a/src/commands/build/wranglerjs/output.rs
+++ b/src/commands/build/wranglerjs/output.rs
@@ -1,9 +1,9 @@
+use crate::terminal::emoji;
 use flate2::write::ZlibEncoder;
 use flate2::Compression;
 use number_prefix::{NumberPrefix, Prefixed, Standalone};
 use serde::Deserialize;
 use std::io::prelude::*;
-use crate::terminal::emoji;
 
 // This structure represents the communication between {wranglerjs} and
 // {wrangler}. It is send back after {wranglerjs} completion.
@@ -50,7 +50,7 @@ impl WranglerjsOutput {
             Standalone(bytes) => format!("{} bytes", bytes),
             Prefixed(prefix, n) => format!("{:.0} {}B", n, prefix),
         };
-        
+
         let human_leftover = if let Some(bytes_left) = bytes_left {
             let msg = match NumberPrefix::binary(bytes_left as f64) {
                 Standalone(bytes) => format!("{} bytes", bytes),
--- a/src/commands/build/wranglerjs/output.rs
+++ b/src/commands/build/wranglerjs/output.rs
@@ -3,6 +3,7 @@ use flate2::Compression;
 use number_prefix::{NumberPrefix, Prefixed, Standalone};
 use serde::Deserialize;
 use std::io::prelude::*;
+use crate::terminal::emoji;
 
 // This structure represents the communication between {wranglerjs} and
 // {wrangler}. It is send back after {wranglerjs} completion.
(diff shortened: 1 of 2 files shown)


src/commands/build/wranglerjs/output.rs
```diff
@@ -28,23 +28,45 @@ impl WranglerjsOutput {
...
+        if let Some(wasm) = &self.wasm {
+            e.write_all(wasm.to_owned().as_bytes())
+                .expect("could not write wasm buffer");
+        }
+
         let compressed_bytes = e.finish();
+        let compressed_size = compressed_bytes.unwrap().len();
+
+        //hardcoded 1MB limit
+        let mut bytes_left = (1_000_000) - (compressed_size as isize);
```
xtuc: It would make sense to me to move this value into `settings/mod.rs`

src/commands/build/wranglerjs/output.rs
```diff
@@ -28,23 +29,44 @@ impl WranglerjsOutput {
...
         let compressed_bytes = e.finish();
+        let compressed_size = compressed_bytes.unwrap().len();
+
+        let mut bytes_left = (MAX_BUNDLE_SIZE) - (compressed_size as isize);
+        let mut oversized = false;
 
-        match NumberPrefix::decimal(compressed_bytes.unwrap().len() as f64) {
+        if bytes_left < 0 {
+            oversized = true;
+            bytes_left = -bytes_left;
```
xtuc: Is there a builtin Math.abs in Rust?

src/commands/build/wranglerjs/output.rs
```diff
@@ -28,23 +29,44 @@ impl WranglerjsOutput {
...
-        match NumberPrefix::decimal(size as f64) {
+        let leftover = match NumberPrefix::decimal(bytes_left as f64) {
             Standalone(bytes) => format!("{} bytes", bytes),
             Prefixed(prefix, n) => format!("{:.0} {}B", n, prefix),
+        };
+
+        if oversized {
+            return format!("{}. ⚠️  Your bundle is too large by {}!", size, leftover);
+        } else {
+            return format!("{}. Your bundle has {} left to grow!", size, leftover);
```
xtuc: It's not a race; I think we should remove this branch

xtuc: Also you don't need the return here if you remove the ; at the end

src/commands/build/wranglerjs/output.rs
ashleygwilliams: https://doc.rust-lang.org/std/primitive.i32.html#method.abs

ashleygwilliams:
@rita3ko please chat with Matt about how we want to talk about size limits

src/commands/build/wranglerjs/output.rs
```diff
@@ -24,50 +25,87 @@ impl WranglerjsOutput {
         self.errors.join("\n")
     }
 
-    pub fn script_size(&self) -> String {
+    fn project_size_bytes(&self) -> u64 {
```
granjef3: When https://github.com/cloudflare/wrangler/pull/313 lands I think this logic can/should be moved into there so _all_ project types get a bundle size info message, not just wranglerjs/webpack projects.

src/commands/build/wranglerjs/output.rs
```diff
@@ -24,50 +25,87 @@ impl WranglerjsOutput {
...
+                Standalone(bytes) => format!("{} bytes", bytes),
+                Prefixed(prefix, n) => format!("{:.0} {}B", n, prefix),
+            };
+            Some(msg)
+        } else {
+            None
+        };
+
+        match compressed_size {
+            WARN_THRESHOLD...MAX_PROJECT_SIZE => format!("{}. {2} Your built project is {} away from reaching the 1MiB size limit. {2}", human_size, human_leftover.expect("failed to get leftover bytes"), emoji::WARN),
```
granjef3: I have no opinion on "bundle" vs "built project" here.

src/commands/build/wranglerjs/output.rs
```diff
@@ -24,50 +25,87 @@ impl WranglerjsOutput {
...
+            0...WARN_THRESHOLD => format!("{}.", human_size),
+            _ => format!("{}. {1} Your built project has grown past the 1MiB size limit and may fail to deploy. {1}", human_size, emoji::WARN)
         }
     }
+
+    pub fn project_size(&self) -> String {
+        Self::project_size_message(self.project_size_bytes())
```
granjef3: Did this to make testing easier

src/commands/build/wranglerjs/output.rs
```diff
@@ -24,50 +25,87 @@ impl WranglerjsOutput {
...
+
+        e.finish().expect("failed to compress project").len() as u64
     }
 
-    pub fn wasm_size(&self) -> String {
-        let size = self.wasm.to_owned().unwrap().len();
-        match NumberPrefix::decimal(size as f64) {
+    fn project_size_message(compressed_size: u64) -> String {
+        const MAX_PROJECT_SIZE: u64 = 1 << 20; // 1 MiB
+        const WARN_THRESHOLD: u64 = MAX_PROJECT_SIZE - 81_920; //Warn when less than 80 KiB left to grow, ~92% usage
```
EverlastingBugstopper: didn’t know you could put underscores in numeric literals, that’s pretty nifty

PR #233, opened 2019-06-09
Display commands in their defined order

Electroid:
Clap was re-arranging commands alphabetically, this change lets us define our own order. This moves help to the bottom so it's less awkward (since it has no emoji!) Also fixed some emoji spacing issues.

### **OLD:**
Screen Shot 2019-06-09 at 11 48 04 AM

### **NEW:**
Screen Shot 2019-06-09 at 11 47 36 AM

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -41,7 +41,7 @@ fn main() -> Result<(), failure::Error> {
         .subcommand(
             SubCommand::with_name("generate")
                 .about(&*format!(
-                    "{} Generate a new workers project",
+                    "{} Generate a new worker project",
                     emoji::DANCERS
                 ))
                 .arg(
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -37,10 +37,11 @@ fn main() -> Result<(), failure::Error> {
         .version(env!("CARGO_PKG_VERSION"))
         .author("ashley g williams <ashley666ashley@gmail.com>")
         .setting(AppSettings::ArgRequiredElseHelp)
+        .setting(AppSettings::DeriveDisplayOrder)
         .subcommand(
             SubCommand::with_name("generate")
                 .about(&*format!(
-                    "{} Generates a new worker project",
+                    "{} Generate a new workers project",
```
xtuc:
```suggestion
                    "{} Generate a new worker project",
```

PR #239, opened 2019-06-10
Use stdin instead of arguments for wrangler config

xtuc:
This changes the way `wrangler config` works, previously both the email
and the api_key were passed as arguments which would be captured by the
terminal history.

When launching the command it will prompt for an email, once stdin
receive a line it will prompt for the api_key. Both lines are values and
will be stored in the config.

Introduce a `WRANGLER_HOME` environment variable that will configure
where Wrangler defines its home directory. It will contain a `config`
directory.


ashleygwilliams:
i think this is headed in the right direction- how can i help it get across the finish line?

xtuc:
I was stuck on the tests, If that looks good to you I can continue.

ashleygwilliams:
@xtuc sorry for the delayed response- let's definitely finish this up

xtuc:
Currently the tests are missing, they should open a `wrangler config` command, enter the data in stdin and test the output (or the config file).

If someone from @cloudflare/workers-devexp wants to take it over I would be happy to help.

ashleygwilliams:
windows tests are panicking: 

```
thread 'it_generates_the_config' panicked at 'could not read config at "D:\\a\\1\\s\\.it_generates_the_config\\.wrangler\\config\\default.toml": Os { code: 3, kind: NotFound, message: "The system cannot find the path specified." }', src\libcore\result.rs:997:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

xtuc:
My hack about providing a test $HOME doesn't work on windows because of its resolution:  https://docs.rs/dirs/2.0.1/dirs/fn.home_dir.html.

ashleygwilliams:
so, let's only run this test for linux and macos and let's write a separate test for windows that shows it works :) - let me know if you need help with that

src/main.rs
```diff
@@ -1,5 +1,8 @@
 #![allow(clippy::redundant_closure)]
 
+#[macro_use]
+extern crate text_io;
```
granjef3: Nitpick: you can get rid of this extern crate + macro use and just do a ‘use text_io::read’

tests/config.rs
```diff
@@ -0,0 +1,56 @@
+use assert_cmd::prelude::*;
+use std::env;
+use std::fs;
+use std::io::prelude::*;
+use std::process::{Child, Command, Stdio};
+
+// Does not work on Windows because of the home_dir resolution
```
granjef3: could you check $HOME yourself in the config command (instead of using dirs::home_dir) when `cfg!(test)` is true on windows?, that way you can still configure the home dir on windows to run this test?

xtuc: Maybe CONFIG_LOCATION could be an env variable. It will complicate the logic then in https://github.com/cloudflare/wrangler/pull/221 (for XDG support).

granjef3: Thinking about this, I think it makes sense to copy [cargo's behavior](https://doc.rust-lang.org/cargo/reference/environment-variables.html) and have a $WRANGLER_HOME that defines the location of `.wrangler`... then people can put their stuff wherever they want, and the tests can configure it as needed

src/main.rs
xtuc: `text_io` only exposes macros and if I use `text_io::read!` it `cannot find macro `try_read!` in this scope`.

tests/config.rs
xtuc: Good idea

src/settings/global_user.rs
```diff
@@ -42,3 +45,15 @@ fn get_global_config() -> Result<GlobalUser, failure::Error> {
         }
     }
 }
+
+pub fn get_global_config_dir() -> Result<PathBuf, failure::Error> {
+    let home_dir = if let Ok(value) = env::var("WRANGLER_HOME") {
```
ashleymichal: do we (want to) indicate use of this env var in any of our documentation?

xtuc: Currently it's undocumented, once this lands we can create an issue. Using XDG standarts (#226) might changes how it works.

PR #241, opened 2019-06-11
Make it more clear that you can install Wrangler though npm

zackbloom:
This is particularly important for people being linked directly to the installation section. (see: https://twitter.com/alexellisuk/status/1138460356576256000)


alexellis:
@zackbloom thanks for the PR 👍 I left a comment above btw.

PR #245, opened 2019-06-13
Enforce one webpack config

xtuc:
We don't support multiple webpack configurations, for better UX we
should make sure to enforce it.

The motiviation is that we can't infer which bundle (or configuration)
is meant to be deployed as a worker.


tests/webpack_multiple_config/package.json
```diff
@@ -0,0 +1 @@
+{}
```
xtuc: Would be good to auto generate the package.json too, but not very important at the moment

wranglerjs/index.js
```diff
@@ -3,10 +3,16 @@ const { join } = require("path");
 const { writeFileSync } = require("fs");
 const WasmMainTemplatePlugin = require("webpack/lib/wasm/WasmMainTemplatePlugin");
 
+function error(msg) {
+  console.error("Error: " + msg);
```
xtuc: I made this to avoid showing a stack strace

PR #252, opened 2019-06-14
feat: Implement livereloading for wrangler preview

### LINKED ISSUE #235: Add watch option to wrangler build
Can you guys add a `--watch` argument to `wrangler build` so wrangler watches the source files and rebuilds whenever changes are made? Would be super useful.

xortive:
this PR adds the `--watch` flag to `build` and `preview`, which watches your Wrangler project for changes and builds it when it's ready. It works for all project types

- `javascript` -> watches the entry point in `package.json`
- `rust` -> watches the `/src` directory, and the entry point in `package.json`
- `webpack` -> uses webpack's watching, which _just works_

When run in `--watch` mode, `preview` hides the fiddle editor, since you would be using a local text editor instead.

to test:
`cargo install --force --debug --git https://github.com/xortive/wrangler --branch malonso/hotreload`

Closes #235


src/commands/build/wranglerjs/mod.rs
```diff
@@ -164,3 +231,11 @@ pub fn install(cache: &Cache) -> Result<PathBuf, failure::Error> {
     run_npm_install(wranglerjs_path.clone()).expect("could not install wranglerjs dependencies");
     Ok(wranglerjs_path)
 }
+
+fn random_chars(n: usize) -> String {
```
xtuc: We could split this one out of Wranglerjs build command

wranglerjs/index.js
```diff
@@ -93,4 +93,10 @@ compiler.run((err, stats) => {
   }
 
   writeFileSync(args["output-file"], JSON.stringify(bundle));
-});
+};
+
+if (args["watch"] === "1") {
+  compiler.watch(fullConfig.watchOptions, compilerCallback);
```
xtuc: have you tried with the wasm-pack-plugin workflow? Which has it's own watcher.

src/commands/build/wranglerjs/mod.rs
```diff
@@ -67,22 +139,17 @@ pub fn run_build(
...
-    let status = command.status()?;
-    let output = fs::read_to_string(temp_file.clone()).expect("could not retrieve ouput");
-    fs::remove_file(temp_file)?;
-
-    if status.success() {
-        Ok(serde_json::from_str(&output).expect("could not parse wranglerjs output"))
-    } else {
-        failure::bail!("failed to execute `{:?}`: exited with {}", command, status)
+    if watch {
+        command.arg("--watch=1");
```
xtuc: we could pass `command.arg("--watch={:?}", watch)`

wranglerjs/index.js
granjef3: I have not. I'm not sure how plugins that implement their own watcher interact with webpack's built in one. This may need to be revisited

src/commands/build/wranglerjs/mod.rs
granjef3: done

src/commands/build/wranglerjs/mod.rs
granjef3: right now it's only used to generate the wranglerjs temp file. If anything, I would just put this next to where the temp file is created in wranglerjs

wranglerjs/index.js
xtuc: Actually I think this should be transparent, but we should try it. The code for wasm-pack-plugin is here https://github.com/wasm-tool/wasm-pack-plugin/blob/master/plugin.js.

src/commands/build/wranglerjs/mod.rs
xtuc: Could we inline it in the temp_file generation?

granjef3:
This is feature complete!

Tagging @ejcx for security review on the wrangler side of this, specifically directed towards the websocket portion of this PR.

src/commands/build/watch/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+use crate::terminal::message;
+use crate::{commands, install};
+
+use notify::{watcher, RecursiveMode, Watcher};
+use std::env;
+use std::sync::mpsc::{channel, Sender};
+use std::thread;
+use std::time::Duration;
+
+pub const COOLDOWN_PERIOD: Duration = Duration::from_millis(2000);
```
granjef3: Considering making this a config option, it refers to the amount of time wrangler will ignore new fs events when it detects the first one. Any new events during the cooldown period restart the timer.

granjef3:
the ctrl-c handler doesn't work after the notifier detects it's first change. Have no idea where to even begin debugging something like this

src/commands/publish/preview/mod.rs
```diff
@@ -3,89 +3,93 @@ use std::process::Command;
...
 
     let preview_host = "example.com";
-    let https = 1;
-    let script_id = &p.id;
+    let https = true;
+    let script_id = &upload_and_get_id()?;
 
     let preview_address = "https://00000000000000000000000000000000.cloudflareworkers.com";
     let cookie = format!(
         "__ew_fiddle_preview={}{}{}{}",
```
ejcx: It's clearly out of scope for review but I went and read the implementation of this cookie serverside and, yeesh.

granjef3: @ejcx Thanks for the review, wanted to know if there's an internal ticket you made for the serverside cookie implementation, and if it's blocking this feature.

wranglerjs/index.js
ashleygwilliams: is this convo resolved? if so please mark it as such! or update the issue to changes requested status :)

granjef3:
I'm mildly concerned about the stability of this feature, and I think maybe we should include it in a RC release first or gate it under an experimental flag. The worst thing that can happen however is a poor user experience -- the watcher not seeing an update or wrangler `panic`ing since it expects a file to be present that is not present.

Especially with the `webpack` project type that effectively has two file watchers connected together (webpack watches project, we watch webpack output) there may be a race condition or two lingering that could cause a `panic`. Better error handling or tests for livereloading would mitigate this concern.

wranglerjs/index.js
granjef3: This was pre-webpack watcher so I believe it is resolved. Any plugin that works with webpack in watch mode will work with wrangler

signalnerve:
Hey @granjef3 - planning ahead to a future release where this is merged: can you update the PR description with a few sentences about what this PR does, and how this feature works?

ashleygwilliams:
need to coordinate with UI release

xortive:
to test:
`cargo install --force --debug --git https://github.com/xortive/wrangler --branch malonso/hotreload`

EverlastingBugstopper:
@xortive when you get the chance could you edit your top-level comment to include an overview of how to QA and what exactly this PR does?

src/commands/publish/preview/mod.rs
```diff
@@ -49,15 +49,25 @@ pub fn preview(
...
+        let mut ws_port: u16 = 8025;
+
+        let server = loop {
+            let result = WebSocket::new(|out| FiddleMessageServer { out })?
+                .bind(format!("127.0.0.1:{}", ws_port)); //explicitly use 127.0.0.1, since localhost can resolve to 2 addresses
+
+            match result {
+                Ok(server) => break server,
+                //if 65535-8025 ports are filled, this will cause a panic due to the overflow
+                //if you are using that many ports, i salute you
```
EverlastingBugstopper: lol

gabbifish: Interesting! I've never used websockets before, but I always thought a lot of these libraries would bake in their own random selection of an ephemeral port to use for egress traffic... I'm going to take a look :)

gabbifish: Huh, looks like you do need to iterate until you find an available port! It is what it is :)

gabbifish: Actually, could we add nice error handling that doesn't panic and throws a proper error if ws_port >= 65535 (max port)? We can even be fancy and ensure that this counter doesn't go over 49152 (IANA recommendation for bottom number of ephemeral port: https://en.wikipedia.org/wiki/Ephemeral_port)

EverlastingBugstopper:
As per discussion in UI channel - let's have regular ol' `wrangler preview` not open the sidebar either. I think this is going to do wonders for the wrangler experience! 🔥

EverlastingBugstopper:
Needs `cargo fmt` as well 😄

EverlastingBugstopper:
Also before we merge this it's probably best (but won't block on it) to squash it down to a couple of commits - if at all possible.

wranglerjs/index.js
```diff
@@ -65,31 +65,41 @@ fetchCompileWasmTemplatePlugin.fn = function(compilation) {
...
-  const hasWasmModule = wasmModuleAsset !== undefined;
-
-  bundle.script = jsAssets.reduce((acc, k) => {
-    const asset = assets[k];
-    return acc + asset.source();
-  }, "");
-
-  if (hasWasmModule === true) {
-    bundle.wasm = Buffer.from(assets[wasmModuleAsset].source()).toString("base64");
+  if (stats.hash != lastHash) {
```
xtuc:
```suggestion
  if (stats.hash !== lastHash) {
```

If both are strings.

src/commands/build/wranglerjs/mod.rs
```diff
@@ -39,26 +47,82 @@ pub fn run_build(project: &Project) -> Result<(), failure::Error> {
...
-        Ok(())
-    } else {
-        failure::bail!("failed to execute `{:?}`: exited with {}", command, status)
+    thread::spawn(move || {
+        let _command_guard = util::GuardedCommand::spawn(command);
+
+        let (watcher_tx, watcher_rx) = channel();
+        let mut watcher = watcher(watcher_tx, Duration::from_secs(1)).unwrap();
+
+        watcher.watch(&temp_file, RecursiveMode::Recursive).unwrap();
```
xtuc: I guess you don't need Recursive mode here since it's just a file

signalnerve:
this 502s for me locally :( unclear if it's related to #429, what would be useful for debugging here?

```
☁  worker  ~/.cargo/bin/wrangler preview --watch
⬇️ Installing wasm-pack...
✨   Built successfully, built project size is 2 KiB.
Error: https://cloudflareworkers.com/script: Server Error: 502 Bad Gateway
```

signalnerve:
this worked really well for me with a new application, nice work @xortive! 

one piece of feedback: as i watched the live reloading happen in the terminal, i found that it was unclear _when_ the browser should be up-to-date. here's the terminal output i saw a few times as i made changes:

```
🌀  Detected changes...
🌀  Detected change during cooldown...
🌀  Cooldown over, propogating changes...
✨   Built successfully, built project size is 570 bytes.
👷  Updating preview with changes
```

it seems like the "Updating preview with changes" step is always the last thing that happens before the preview actually, well, updates. would it make sense to change that text output to "**Updated** preview with changes", _or_ have another line of output that indicates when the preview window is done updating? not a blocker, just stood out to me as i was trying this out

src/commands/publish/preview/mod.rs
```diff
@@ -90,23 +56,37 @@ pub fn preview(
         };
 
         open_browser(&format!(
-            "https://cloudflareworkers.com/ui/staging/index.html?wrangler_session_id={}\\&wrangler_ws_port={}\\&hide_editor#{}:{}{}",
+            "https://cloudflareworkers.com/?wrangler_session_id={}\\&wrangler_ws_port={}\\&hide_editor#{}:{}{}",
             &session.to_string(), ws_port, script_id, https_str, preview_host,
         ))?;
 
+        //don't do initial GET + POST with livereload as the expected behavior is unclear.
```
EverlastingBugstopper: what do you mean here?

signalnerve:
my previous issue w/ `wrangler preview --watch` on a project with kv has been fixed - yay! i'll work on some readme documentation, will track it in a separate PR

src/commands/build/wranglerjs/mod.rs
xortive: it just gets ignored, but I will change it since it makes the intent unclear! thanks for the catch

src/commands/build/watch/mod.rs
```diff
@@ -0,0 +1,86 @@
+mod watcher;
+pub use watcher::wait_for_changes;
+
+use crate::commands::build::{command, wranglerjs};
+use crate::commands::publish::Package;
+use crate::settings::project::{Project, ProjectType};
+use crate::terminal::message;
+use crate::{commands, install};
+
+use notify::{watcher, RecursiveMode, Watcher};
```
ashleygwilliams: i would just `use notify;` so it's clear what these are imported from

src/commands/build/watch/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+use std::env;
+use std::sync::mpsc::{channel, Sender};
```
ashleygwilliams: same here: i would `use std::sync` so it's clear that channel and Sender are from mpsc

src/commands/build/watch/watcher.rs
```diff
@@ -0,0 +1,50 @@
...
+    rx: &Receiver<DebouncedEvent>,
+    cooldown: Duration,
+) -> Result<PathBuf, Error> {
+    loop {
+        let event = rx.recv()?;
+        match get_changed_path_from_event(event) {
+            Ok(Some(path)) => {
+                message::working("Detected changes...");
+                //wait for cooldown
+                while let Ok(_e) = rx.recv_timeout(cooldown) {
```
ashleygwilliams: convention here would be to just write `Ok(_)`

src/commands/build/wranglerjs/mod.rs
```diff
@@ -39,26 +47,84 @@ pub fn run_build(project: &Project) -> Result<(), failure::Error> {
...
+}
 
-        bundle
-            .write(&wranglerjs_output)
-            .expect("could not write bundle to disk");
+pub fn run_build_and_watch(
+    project: &Project,
+    tx: Option<Sender<()>>,
+) -> Result<(), failure::Error> {
+    let (mut command, temp_file, bundle) = setup_build(project)?;
```
ashleygwilliams: long term we probably want a better solution here ;)

src/commands/build/wranglerjs/mod.rs
```diff
@@ -39,26 +47,84 @@ pub fn run_build(project: &Project) -> Result<(), failure::Error> {
...
+    command.arg("--watch=1");
+
+    info!("Running {:?} in watch mode", command);
+
+    thread::spawn(move || {
+        let _command_guard = util::GuardedCommand::spawn(command);
+
+        let (watcher_tx, watcher_rx) = channel();
+        let mut watcher = watcher(watcher_tx, Duration::from_secs(1)).unwrap();
```
ashleygwilliams: i am worried about these unwraps!

ashleygwilliams: if we can: let's get rid of unwraps- make them errors if at all possible. if not possible: info debugging lines would be very useful. ask yourself- when this panics.. what am i gonna want to know

ashleygwilliams: (this is basically creating an `expect`)

src/commands/build/wranglerjs/mod.rs
```diff
@@ -39,26 +47,84 @@ pub fn run_build(project: &Project) -> Result<(), failure::Error> {
...
+    output: &WranglerjsOutput,
+) -> Result<(), failure::Error> {
+    if output.has_errors() {
+        message::user_error(output.get_errors().as_str());
+        failure::bail!("Webpack returned an error");
     }
+
+    bundle
+        .write(output)
+        .expect("could not write bundle to disk");
```
ashleygwilliams: can this be a `?`

src/commands/publish/preview/upload.rs
```diff
@@ -0,0 +1,159 @@
+use crate::commands::publish;
+use crate::http;
+use crate::settings::global_user::GlobalUser;
+use crate::settings::project::Project;
+use crate::terminal::message;
+use reqwest::Client;
+use serde::Deserialize;
+#[derive(Debug, Deserialize)]
```
ashleygwilliams: id' probably put a newline here

src/main.rs
```diff
@@ -188,10 +200,18 @@ fn run() -> Result<(), failure::Error> {
             None => None,
         };
         commands::init(name, project_type)?;
-    } else if matches.subcommand_matches("build").is_some() {
+    } else if let Some(matches) = matches.subcommand_matches("build") {
```
ashleygwilliams: do we need build --watch?

src/main.rs
```diff
@@ -207,7 +227,12 @@ fn run() -> Result<(), failure::Error> {
             None => None,
         };
 
-        commands::preview(project, user, method, body)?;
+        let watch = match matches.occurrences_of("watch") {
```
ashleygwilliams: match presence of not occurences .. because what if someone types `wrangler preview --watch --watch`

src/commands/publish/preview/mod.rs
xortive: we ignore the `wrangler preview get --watch` and `wrangler preview post --watch` commands as it is unclear what they should do. They will be addressed once we have a command for browserless preview

ashleygwilliams:
closing in favor of #451

PR #253, opened 2019-06-14
feat: allow custom webpack configuration in wrangler.toml

### LINKED ISSUE #246: Allow the user to specify an alternative webpack.config.js file
Came up in the context of #245, a user might want to use a webpack configuration in an alternate location (instead of `webpack.config.js` in cwd).

This implies to add a configuration; I'm not sure if we want to use a flag or an entry in the wrangler.toml.

Implementation detail; our wranglerjs component also supports arguments, it will be a new argument (see `--use-entry` for example).

EverlastingBugstopper:
Fixes #246 

This PR allows wrangler to work with projects like this one: https://github.com/jrf0110/cf-workers-boilerplate


EverlastingBugstopper:
Side note: Rust is so cool. This is my first Rust PR, I would love feedback on what could be better. I'm not sure if my patterns are the best.

src/commands/build/wranglerjs/mod.rs
```diff
@@ -65,6 +66,8 @@ pub fn run_build(
             .to_string();
         command.arg("--no-webpack-config=1");
         command.arg(format!("--use-entry={}", package_main));
+    } else {
+        command.arg(format!("--webpack-config={}", &webpack_config_path.to_str().unwrap().to_string()));
```
xtuc:
```suggestion
        command.arg(format!("--webpack-config={}", &webpack_config_path.to_str()));
```
Would that work? I'm also not a Rust expert

EverlastingBugstopper: It does not

EverlastingBugstopper:
I added another test case to make sure that a specified webpack configuration file with multiple exports will throw an error

src/commands/build/wranglerjs/mod.rs
granjef3: try this. (? operator will return Err() if there is a problem instead of doing a panic like unwrap)

```suggestion
        command.arg(format!("--webpack-config={}", &webpack_config_path.to_str()?.to_string()));
```

EverlastingBugstopper: No that does not work either `the trait `std::error::Error` is not implemented for `std::option::NoneError`

EverlastingBugstopper: I think it's fine the way it is, it should never be an empty path, even if you specify webpack_config="" it will default to "webpack.config.js".

EverlastingBugstopper:
Ran `cargo fmt`

EverlastingBugstopper:
Ran `cargo clippy`

src/commands/build/wranglerjs/mod.rs
granjef3: after some research you could use Option::ok_or() to convert into a Result, then use ? on that result to get what's inside.

EverlastingBugstopper: Why is that necessary?

tests/webpack_multiple_specify_config/webpack.worker.js
```diff
@@ -0,0 +1,5 @@
+module.exports = [
+    { entry: "./a.js" },
+    { entry: "./b.js" }
+  ]
+
```
xtuc:
```suggestion
```

tests/build.rs
```diff
@@ -80,7 +95,21 @@ fn it_fails_with_multiple_webpack_configs() {
...
+fn it_fails_with_multiple_specify_webpack_configs() {
+    let fixture = "webpack_multiple_specify_config";
+    create_temporary_copy(fixture);
+
+    settings! {fixture, r#"
+        type = "Webpack"
+        webpack_config = "webpack.worker.js"
+    "#};
+
+    build_fails_with(fixture, "Multiple webpack configurations are not supported. You can specify a different path for your webpack configuration file in wrangler.toml with the `webpack_config` field");
```
ashleygwilliams: long term... testing on the full error message might make this test a little brittle. i think it's totally fine to keep now, but if we find the error message changes slightly a lot over time... we may just want to test that the error message contains some keywords that we know won't change and are enough to identify it :)

src/commands/build/mod.rs
```diff
@@ -9,7 +9,14 @@ use std::process::Command;
 
 use crate::emoji;
 
-pub fn build(cache: &Cache, project_type: &ProjectType) -> Result<(), failure::Error> {
+pub fn build(cache: &Cache, project: &Project) -> Result<(), failure::Error> {
+    let project_type = &project.project_type;
+    let webpack_config_path = PathBuf::from(
+        &project
+            .webpack_config
+            .clone()
```
ashleygwilliams: ideally we can avoid clones but sometimes it's not ideal for code readability :) that being said i think this is fine for now, we should just audit these generally

ashleygwilliams:
@EverlastingBugstopper do you think you'd be able to work with @victoriabernard92 + @signalnerve  to make a small template that demonstrates using a user created webpack config? i think it'd be great to release this alongside an example

src/commands/build/mod.rs
EverlastingBugstopper: Yeah I didn't really like it, just what the compiler recommended. I'm still a little iffy on how Rust handles borrowing + what is idiomatic.

EverlastingBugstopper:
@ashleygwilliams Yeah I could do that. Would that be a part of this PR or separate?

ashleygwilliams:
@EverlastingBugstopper let's do it in a different one! landing this now :)

PR #257, opened 2019-06-16
clarified intro link in README

tomByrer:
The prior link was labeled 'documentation' which was false; it lead to the landing page.  The real installation docs lead back to this page, so the statement was redundant.  To keep your SEO experts happy, I bumped up the CfW sentence up a section ;)

I also gave the badges a bit more breathing room; looked like 1 big badge.


ashleygwilliams:
thank you very much!

PR #258, opened 2019-06-17
Adds formatting, linting, and testing pre-commit hook

### LINKED ISSUE #254: Add `cargo fmt` and `cargo clippy` to ./git/hooks/pre-commit
Should prevent a lot of CI issues

EverlastingBugstopper:
These hooks will be compiled the first time someone runs `cargo test`

`rusty-hook` https://github.com/swellaby/rusty-hook

Fixes #254


EverlastingBugstopper:
If this is too janky/we don't want the dependency perhaps we could just add a section to the README suggesting these pre-commit hooks. Could also switch these to pre-push and/or add `cargo test`

Cargo.toml
```diff
@@ -34,6 +34,11 @@ flate2 = "1.0.7"
 [dev-dependencies]
 assert_cmd = "0.11.1"
 fs_extra = "1.1.0"
+pre-commit = "0.5.2"
+
+[package.metadata.precommit]
+fmt = "cargo fmt"
+clippy = "cargo clippy"
```
EverlastingBugstopper: If we want other hooks (or if we want to add some arguments to these commands) we can add them easily here. I was initially using `cargo-husky` but their user-hooks implementation wasn't working

EverlastingBugstopper: `precommit` package isn't compatible w/Windows. Trying `rusty-hook`

PR #259, opened 2019-06-17
feat: adds more descriptive subdomain errors

### LINKED ISSUE #207: Give a nicer error message back if a user runs subdomain command with their existing subdomain
I was expecting that I would need to run the new `subdomain` command with my workers.dev subdomain, to register it in the configuration. I did that and got this error back:

```sh
worker $ wrangler subdomain signalnerve
🐌  Registering your subdomain, signalnerve.workers.dev, this could take up to a minute.
Error: ErrorMessage { msg: "⛔ This account already has a registered subdomain. You can only register one subdomainper account. Your subdomain is signalnerve.workers.dev \n Status Code: 409 Conflict\n Msg: {\n  \"result\": null,\n  \"success\": false,\n  \"errors\": [\n    {\n      \"code\": 10036,\n      \"message\": \"workers.api.error.account_has_subdomain\"\n    }\n  ],\n  \"messages\": []\n}\n" }
```

It would be neat to validate the user's existing subdomain against the argument they passed in for the command: if they're the same, we can return something back like "Your domain is already registered!"

EverlastingBugstopper:
Warns users when attempting to add a subdomain w/o account_id in
wrangler.toml

Doesn't show failure when attempting to register the subdomain that
is already associated with their account

Fixes #207

### DIFF
--- a/src/commands/subdomain.rs
+++ b/src/commands/subdomain.rs
@@ -63,6 +63,11 @@ fn subdomain_addr(account_id: &str) -> String {
 }
 
 pub fn subdomain(name: &str, user: &GlobalUser, project: &Project) -> Result<(), failure::Error> {
+    if project.account_id.is_empty() {
+        failure::bail!(
+            "⛔ You must provide an account_id in your wrangler.toml before creating a subdomain!"
+        )
+    }
     let msg = format!(
         "Registering your subdomain, {}.workers.dev, this could take up to a minute.",
         name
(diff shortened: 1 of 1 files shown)


src/commands/subdomain.rs
```diff
@@ -83,32 +86,36 @@ pub fn subdomain(name: &str, user: &GlobalUser, project: &Project) -> Result<(),
         .body(sd_request)
         .send()?;
 
-    let msg;
+    let mut msg = None;
```
EverlastingBugstopper: I'm not sure if this is the best way to do this, pointers much appreciated

EverlastingBugstopper: nvm: figured out how to return early :)

src/commands/subdomain.rs
```diff
@@ -89,12 +94,21 @@ pub fn subdomain(name: &str, user: &GlobalUser, project: &Project) -> Result<(),
...
             let sd = Subdomain::get(account_id, user)?;
-            msg = format!(
-                "⛔ This account already has a registered subdomain. You can only register one subdomain per account. Your subdomain is {}.workers.dev \n Status Code: {}\n Msg: {}",
-                sd,
-                res.status(),
-                res_text,
-            );
+            if sd == name {
+                println!(
+                    "{} Your subdomain {}.workers.dev has already been registered!",
```
ashleygwilliams: can you say more about what this message is trying to communicate? on first glance i read it and it seemed as though it was supposed to be an error message, but the sparkles seemed confusing. i *think* you mean it as an informative message... let's figure out what exactly we're communicating here, would love to hear more about how you thought about it!

EverlastingBugstopper: So this case is for when someone tries to register the subdomain they've already registered. I don't really think we need to return an error message here (which is the current behavior). Perhaps there's a better way to say this...

EverlastingBugstopper: If that's not clear, this message is returned when avery@cloudflare.com has already registered avery.workers.dev and runs `wrangler subdomain avery`

signalnerve: I would read this as "Your subdomain has already been registered _(by someone else)_" – is that the intent here? If not, should it be something like "You've already registered the subdomain xyz.workers.dev"?

signalnerve: I would read this as "Your subdomain has already been registered (by someone else)" – is that the intent here? If not, should it be something like "You've already registered the subdomain xyz.workers.dev"?

EverlastingBugstopper: Yeah so this case is specifically for when the user is trying to register the subdomain that they themselves have already registered. The latter makes sense to me

PR #263, opened 2019-06-21
Add message module for terminal output

### LINKED ISSUE #219: Standardize terminal messaging output function
Rather than calling `println!`, it would be nice to call something like `message.success` with some parameters that standardize the formatting of the output for e.g. `success`, `info`, `error`, etc.

ashleymichal:
fixes #219 

* include semantic commands for success, info, warn, etc.
* alphabetize emoji 🤓 
* replace some remaining inline emoji


ashleygwilliams:
looks like you have an import error and some dead code warnings! lemme know if you need help with those :)

ashleygwilliams:
adding @granjef3 as a reviewer as we'll tap him to finish out this migration :)

src/terminal/message.rs
```diff
@@ -0,0 +1,34 @@
+use super::emoji;
+use console::Emoji;
+
+fn message(label: Emoji, msg: &str) {
+    println!("{}", format!("{0} {1} {0}", label, msg));
```
granjef3:
```suggestion
    println!("{0} {1} {0}", label, msg);
```

src/commands/build/wranglerjs/bundle.rs
```diff
@@ -254,7 +254,7 @@ mod tests {
     fn cleanup(name: String) {
         let current_dir = env::current_dir().unwrap();
         let path = Path::new(&current_dir).join(name);
-        println!("p: {:?}", path);
+        message::info(format!("p: {:?}", path));
```
granjef3: need to `use crate::terminal::message` at the top of this file

src/terminal/emoji.rs
```diff
@@ -13,16 +13,20 @@
...
 pub static DANCERS: Emoji = Emoji("👯 ", "");
+pub static EYES: Emoji = Emoji("👀 ", "");
+pub static FACEPALM: Emoji = Emoji("🤦‍♀️ ", "");
+pub static INBOX: Emoji = Emoji("📥 ", "");
+pub static INFO: Emoji = Emoji("💁‍ ", "");
 pub static MICROSCOPE: Emoji = Emoji("🔬 ", "");
-pub static CRAB: Emoji = Emoji("🦀 ", "");
+pub static SHEEP: Emoji = Emoji("🐑 ", "");
 pub static SLEUTH: Emoji = Emoji("🕵️‍♂️", "");
+pub static SNAIL: Emoji = Emoji("🐌 ", "");
```
granjef3: Unused -- makes clippy complain.

src/terminal/emoji.rs
```diff
@@ -13,16 +13,20 @@
 
 use console::Emoji;
 
-pub static WORKER: Emoji = Emoji("👷 ", "");
-pub static SPARKLES: Emoji = Emoji("✨  ", "");
+pub static CONSTRUCTION: Emoji = Emoji("🚧 ", "");
+pub static CRAB: Emoji = Emoji("🦀 ", "");
 pub static DANCERS: Emoji = Emoji("👯 ", "");
+pub static EYES: Emoji = Emoji("👀 ", "");
+pub static FACEPALM: Emoji = Emoji("🤦‍♀️ ", "");
```
granjef3: Unused -- makes clippy complain.

src/terminal/message.rs
```diff
@@ -0,0 +1,34 @@
...
+
+pub fn info(msg: &str) {
+    message(emoji::INFO, msg);
+}
+
+pub fn success(msg: &str) {
+    message(emoji::SPARKLES, msg);
+}
+
+pub fn warn(msg: &str) {
```
granjef3: Unused -- makes clippy complain.

src/terminal/message.rs
```diff
@@ -0,0 +1,34 @@
...
+    message(emoji::WARN, msg);
+}
+
+pub fn user_error(msg: &str) {
+    message(emoji::EYES, msg);
+}
+
+pub fn service_error(msg: &str) {
```
granjef3: Unused -- makes clippy complain.

ashleymichal: There isn't currently a place where these would be used with a println!, only with failure::bail. I agree keeping them would be ideal, i kinda want the equivalent of an eslintignore here though.

granjef3: there is #[allow(dead_code)]

ashleygwilliams: i do not think we should commit dead code. i think the pattern here is clear enough that someone can add. if there's a space in the code to use it, put it in that one spot.

it's very hard to keep track of and clean up the allow dead code attributes

PR #267, opened 2019-06-24
Add an authenticated HTTP client

### LINKED ISSUE #238: Add X-Auth in http base client
After merging https://github.com/cloudflare/wrangler/pull/236 I noticed that the X-Auth headers are also common, we could add it to the default headers.

Electroid:
Resolves #238. Any requests to the Cloudflare API will now use `auth_client`. Tested and works.

### DIFF
--- a/src/commands/publish/mod.rs
+++ b/src/commands/publish/mod.rs
@@ -47,7 +47,7 @@ pub fn create_kv_namespaces(user: &GlobalUser, project: &Project) -> Result<(), 
         project.account_id,
     );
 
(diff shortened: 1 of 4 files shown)


xtuc:
> Also added some operating system information to the user agent

I don't think we need that, I would prefer removing it.

Also could we replace the `client()` in favor of the `auth_client()` completely?

Electroid:
Will do on the agent. For the client, there are some requests, like the playground, where we don't need auth and probably shouldn't send the headers.

PR #271, opened 2019-06-24
Check wrangler.toml before doing publish steps

### LINKED ISSUE #197: Validate wrangler.toml before Fetching Deps
On a fresh project , invoking `wrangler publish` will fetch dependencies (first) and then error out if `account_id` (or other vals) in `wrangler.toml` are not set.

This could be frustrating on slower networks, or even if you just waited to see it fail.

Suggestion: validate the required fields in `wrangler.toml` first.

### LINKED ISSUE #211: wrangler.toml firing wrong error messages on publish
**wrangler.toml** has `zone_id` and `account_id` missing `routes`

`Error: ErrorMessage { msg: "You must provide a zone_id in your wrangler.toml before publishing!" }`

**wrangler.toml** has  `account_id` `routes` missing `zone_id`

`Error: ErrorMessage { msg: "⛔ There was an error featching your project\'s routes.\n Status Code: 400 Bad Request\n Msg: {\"success\":false,\"errors\":[{\"code\":7003,\"message\":\"Could not route to \\/zones\\/workers\\/routes, perhaps your object identifier is invalid?\"},{\"code\":7000,\"message\":\"No route for that URI\"}],\"messages\":[],\"result\":null}" }`

granjef3:
This PR verifies the presence of `account_id` and `name` in wrangler.toml when calling `wrangler publish`
and additionally verifies the presence of `zone_id` and `route` when calling `wrangler publish --release`, before actually doing any steps towards said command (important for users on slow internet connections, as some of these steps involve fetching deps).

Fixes #197.
Probably fixes #211.


src/commands/publish/mod.rs
```diff
@@ -23,6 +23,36 @@ use crate::settings::project::{Project, ProjectType};
...
+    if let Some(mf) = missing_fields {
+        failure::bail!(
+            "Check your wrangler.toml, it is missing {} which is required to publish!",
+            mf
+        )
+    }
+
+    if let Some(mf) = missing_fields_release {
+        failure::bail!(
+            "Check your wrangler.toml, it is missing {} which is required to publish a release!",
```
signalnerve: This phrasing feels a little awkward to me, but I'm also not sure how best to re-phrase it...

ashleygwilliams: i'd probably say something like "Your wrangler.toml is missing `X` field. This is is required to publish to <domain/subdomain>."

You'll probably want to switch the message on what they are publishing to... (zone or zoneless). these have different req's

You might also add the caution that they may have been wanting to publish with or without `--release` if the things they have in their wrangler.toml suggest that

src/commands/publish/mod.rs
```diff
@@ -23,6 +23,36 @@ use crate::settings::project::{Project, ProjectType};
 
 pub fn publish(user: &GlobalUser, project: &Project, release: bool) -> Result<(), failure::Error> {
     info!("release = {}", release);
+
+    let missing_fields = match (project.account_id.is_empty(), project.name.is_empty()) {
+        (true, true) => Some("an account_id and name"),
+        (true, false) => Some("an account_id"),
+        (false, true) => Some("a name"),
+        (false, false) => None,
+    };
```
xtuc: Alternatively we could use a list to store all the missing fields and join them in the error message (omiting the `and` but I doesn't matter IMO).

ashleygwilliams: i think i would prefer this strategy as i think the code will be slightly more readable.

granjef3:
My naive implementation here does not currently address #197  -- this logic belongs in `settings/project.rs` or `main.rs`.

Right now where it is in `commands/publish/mod.rs` a build will run (and fetch build deps) before verifying the `wrangler.toml`. The only way to verify `wrangler.toml` before wrangler does any work is:

- when it's first read in `settings/project.rs` (calling `Project::new` with some sort of "required fields")
- immediately after in main.rs. (just copy/paste the code I've written in `commands/publish/mod.rs` and put it after Project::new).

ashleymichal:
@granjef3 , let's discuss tomorrow, but I looking at `main.rs`, the logic in `command::build` is "shared", if you will, among several commands. perhaps it makes sense to extract that to a lower level function that could be called by many commands (build, publish, preview), allowing you to encapsulate the validation inside the preview call while still calling it before the build step.

granjef3:
Resolved the issue I was dealing with before:

- used a list that gets appended (I'd like to display this better in the error message but unsure as to how)
- Removed injecting cache throughout the code (it gets memoized in install/mod.rs now)
- moved the build call from main.rs to inside publish/preview
- all so that we can call validate before build, and have publish own its validator.

ashleymichal:
The only thing i would do differently here is push the build logic down a layer of abstraction, so it is below rather than level with the things that call it. That's a code org thing.

The only thing I am unsure of, and that prevents me from reviewing this, is my ignorance of the lazy static, er, constant? which smells like an addition to global scope to my untrained eye.

granjef3:
I believe `lazy_static!` is file local. normal static constants in rust must be known at compile time, the lazy_static! macro encapsulate some typing/`unsafe` magic so they can be initialized at runtime. I don't think it is a code smell in this use case as it is not being used to do communication or anything like that, just to avoid performing the Cache::new operation more than once as the original code did.

I think the code organization thing is a valid point. Build is more of an "operation" or "step" than a command. Curious to see @ashleygwilliams opinion on this.

ashleygwilliams:
i will take a quick look and give a review- but otherwise i think we should land this

PR #272, opened 2019-06-25
Refactor wranglerjs

granjef3:
- move dependency installing from build/mod.rs to wranglerjs/mod.rs
 - refactor wranglerjs internally by separating out command creation and
   build state handling such that a `run_build_watch` can easily be added without duplicating lots of code.

I kept this very scope-limited, this is the bare minimum needed to be able to call wranglerjs from more than one place without having to do a bunch of setup beforehand.

### DIFF
--- a/src/commands/build/wranglerjs/mod.rs
+++ b/src/commands/build/wranglerjs/mod.rs
@@ -33,10 +33,11 @@ pub fn run_build(cache: &Cache, project: &Project) -> Result<(), failure::Error>
     info!("Running {:?}", command);
 
     let status = command.status()?;
-    let output = fs::read_to_string(temp_file.clone()).expect("could not retrieve ouput");
-    fs::remove_file(temp_file)?;
 
     if status.success() {
+        let output = fs::read_to_string(temp_file.clone()).expect("could not retrieve ouput");
+        fs::remove_file(temp_file)?;
+
         let wranglerjs_output: WranglerjsOutput =
             serde_json::from_str(&output).expect("could not parse wranglerjs output");
 
@@ -59,6 +60,7 @@ pub fn run_build(cache: &Cache, project: &Project) -> Result<(), failure::Error>
         message::success(&msg);
         Ok(())
     } else {
+        fs::remove_file(temp_file)?;
         failure::bail!("failed to execute `{:?}`: exited with {}", command, status)
     }
 }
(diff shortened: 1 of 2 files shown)


src/commands/build/wranglerjs/mod.rs
```diff
@@ -17,32 +17,75 @@ use std::iter;
...
-    webpack_config_path: PathBuf,
-    bundle: &Bundle,
-) -> Result<WranglerjsOutput, failure::Error> {
+pub fn run_build(cache: &Cache, project: &Project) -> Result<(), failure::Error> {
+    let (mut command, temp_file, bundle) = setup_build(cache, project)?;
+
+    info!("Running {:?}", command);
+
+    let status = command.status()?;
+    info!("it ran");
```
xtuc: I'm not sure this log is very useful?

src/commands/build/wranglerjs/mod.rs
```diff
@@ -17,32 +17,75 @@ use std::iter;
...
     let mut command = Command::new(node);
+    let wranglerjs_path = install(cache).expect("could not install wranglerjs");
     command.arg(wranglerjs_path);
+
+    //put path to our wasm_pack as env variable so wasm-pack-plugin can utilize it
+    let wasm_pack_path = install::install("wasm-pack", "rustwasm", cache)?.binary("wasm-pack")?;
     command.env("WASM_PACK_PATH", wasm_pack_path);
 
-    // create temp file for special {wranglerjs} IPC.
+    //create a temp file to communicate with wranglerjs
```
xtuc: IPC stands for Inter-process communication, we are using a file between these two processes. I feel like we should keep that comment. We can rephrase it if it's not clear to you.

src/commands/build/wranglerjs/mod.rs
granjef3: My bad, this is from debugging and I missed it.
```suggestion
```

src/commands/build/wranglerjs/mod.rs
granjef3:
```suggestion
    // create a temp file for IPC with the wranglerjs process
```

src/commands/build/wranglerjs/mod.rs
```diff
@@ -17,32 +17,74 @@ use std::iter;
...
-    webpack_config_path: PathBuf,
-    bundle: &Bundle,
-) -> Result<WranglerjsOutput, failure::Error> {
+pub fn run_build(cache: &Cache, project: &Project) -> Result<(), failure::Error> {
+    let (mut command, temp_file, bundle) = setup_build(cache, project)?;
+
+    info!("Running {:?}", command);
+
+    let status = command.status()?;
+    let output = fs::read_to_string(temp_file.clone()).expect("could not retrieve ouput");
```
xtuc: Could you move this in the success branch?

xtuc: I wouldn't have moved the temp_file deletion too but I guess that's fine.

xtuc:
Just for the record, status: update branch and merge

Issue #242, opened 2019-06-12
don't recommend installing node using brew


ashleygwilliams:
we should tell folks to use nvm

ashleymichal:
closed by #283

PR #283, opened 2019-06-28
Better node install documentation

granjef3:
Many system package managers install node in a way that breaks `npm install -g` for wrangler due to permissions errors. Using `sudo` to get past this is currently broken since `npm` will run our post-install script as `nobody`, and even if it ran as `root` users couldn't access `root`'s home directory.

This PR provides better documentation on how to install node (using a version manager) so users aren't confused by this issue.

I would appreciate feedback on the exact language used here.


xtuc:
Closes https://github.com/cloudflare/wrangler/issues/242

README.md
```diff
@@ -151,11 +151,9 @@ Wrangler can be installed both through [npm](https://www.npmjs.com/get-npm) and
 
 ### Using `npm`:
 
-1. If you don't already have npm on your machine, install it. On OSX with [Homebrew](https://brew.sh/) that looks like:
+1. If you don't already have npm on your machine, install it using npm's recommended method: [a version manager](https://www.npmjs.com/get-npm).
```
ashleymichal:
```suggestion
1. If you don't already have npm on your machine, install it using [npm's recommended method](https://www.npmjs.com/get-npm).
```

README.md
```diff
@@ -151,11 +151,9 @@ Wrangler can be installed both through [npm](https://www.npmjs.com/get-npm) and
...
+    If you have already installed npm with `brew` or `apt`, it is possible you will run into an `EACCES` error (related to how many system packagers install node) while installing wrangler. You can either uninstall that version of node and use the npm recommended install method, or use `cargo` instead.
 
-    ```
-    $ brew install nodejs
-    ```
     
 1. Install Wrangler by running:
```
ashleymichal: the numbering gets weird here; should be corrected somehow. likely would be corrected by making the bit about system package installers into a footnote

README.md
```diff
@@ -151,11 +151,9 @@ Wrangler can be installed both through [npm](https://www.npmjs.com/get-npm) and
 
 ### Using `npm`:
 
-1. If you don't already have npm on your machine, install it. On OSX with [Homebrew](https://brew.sh/) that looks like:
+1. If you don't already have npm on your machine, install it using npm's recommended method: [a version manager](https://www.npmjs.com/get-npm).
+    If you have already installed npm with `brew` or `apt`, it is possible you will run into an `EACCES` error (related to how many system packagers install node) while installing wrangler. You can either uninstall that version of node and use the npm recommended install method, or use `cargo` instead.
```
ashleymichal: this makes for weirdness; let's make this into a footnote after the two steps

README.md
granjef3: This is definitely better, commiting.

README.md
granjef3: Github still renders it fine, although the markdown numbers don't increment, the rendered output does. Will move the bit about system installers "inside" point #1 like the openssl one is.

README.md
granjef3: Will move the bit about system installers "inside" point #1 like the openssl one is.

README.md
ashleymichal: i can't be the only one who opens readmes in my text editor locally

granjef3: Should be a separate PR as the entire doc is written with this numbering

ashleymichal:
bonus points if you instruct the user as to how to learn where their install is and what it was installed with...

README.md
```diff
@@ -151,12 +151,10 @@ Wrangler can be installed both through [npm](https://www.npmjs.com/get-npm) and
 
 ### Using `npm`:
 
-1. If you don't already have npm on your machine, install it. On OSX with [Homebrew](https://brew.sh/) that looks like:
+1. If you don't already have npm on your machine, install it using [npm's recommended method](https://www.npmjs.com/get-npm).
```
ashleygwilliams: we should specify that this is using a node version manager and list/link some here. i think linking out to npm is good but i don't want to force people to click that link to get to where they need to go.

README.md
```diff
@@ -151,12 +151,10 @@ Wrangler can be installed both through [npm](https://www.npmjs.com/get-npm) and
...
+
+    If you have already installed npm with `brew` or `apt`, it is possible you will run into an `EACCES` error (related to how many system packagers install node) while installing wrangler. You can either uninstall that version of node and use the npm recommended install method, or use `cargo` instead.
```
ashleygwilliams: alternatively we can point people to the releases tab. they can manually download the binary for their machine and place it where they'd like

README.md
ashleygwilliams: i think it's important to explain that an OS package manager will install node and it's global node_modules dir in a way that is only accessible to root. using the version manager, it will scope the node installation to the user and as a result will eliminate the need for root permissions

README.md
granjef3: I am concerned about recommending installing a raw binary that can't update itself

README.md
granjef3: I had this before: 

```
If you don't already have npm on your machine, install it using npm's recommended method: [a version manager](https://www.npmjs.com/get-npm).
```

And reverted it at @ashleymichal suggestion as I felt it was clearer without it, and the npm docs are fairly comprehensive wrt installing node.

ashleygwilliams: the first thing you see when you scroll down the get-npm docs (which fwiw i wrote, many years ago) is this:

Screen Shot 2019-06-28 at 3 35 25 PM

we don't want folks to do that. we need to make it clear that the version manager strategy is the one to pick.

README.md
ashleygwilliams: i hear you. i suggest it because its an option for people who are really stuck. in my workshop, there was one person who had to do it this way. since it exists i think it . is worth mentioning, as well as the building from source. 

we would call this a "manual install" and this is a common pattern for open source software.

README.md
granjef3: ahhh I see. Will fix this

README.md
granjef3:
```suggestion
    If you have already installed npm with a package manager, it is possible you will run into an `EACCES` error (related to how many system packagers install npm) while installing wrangler. You can either uninstall that version of npm and use the npm recommended install method (a version manager), use `cargo` instead, or download a binary from [our releases page ](https://github.com/cloudflare/wrangler/releases) and put it on your PATH.
```

granjef3: @ashleymichal Look good to you?

ashleymichal: it's gotten way dense. it looks like it's worth calling out "manual install" in its own section after cargo. then all you have to say here is "you can either uninstall and reinstall as instructed above, or use an alternate wrangler installation method", call out nvm above (now i see why that is valuable, my bad for the earlier comment), and

README.md
```diff
@@ -151,12 +151,10 @@ Wrangler can be installed both through [npm](https://www.npmjs.com/get-npm) and
 
 ### Using `npm`:
 
-1. If you don't already have npm on your machine, install it. On OSX with [Homebrew](https://brew.sh/) that looks like:
+1. If you don't already have npm on your machine, install it using [npm's recommended method](https://www.npmjs.com/get-npm).
+
+    If you have already installed npm with a package manager, it is possible you will run into an `EACCES` error (related to how many system packagers install npm) while installing wrangler. You can either uninstall npm and reinstall using the npm recommended install method (a version manager), or one of our other install methods.
```
ashleymichal:
```suggestion
    If you have already installed npm with a package manager, it is possible you will run into an `EACCES` error while installing wrangler. This is related to how many system packagers install npm. You can either uninstall npm and reinstall using the npm recommended install method (a version manager), or one of our other install methods.
```

README.md
```diff
@@ -188,7 +186,13 @@ Wrangler can be installed both through [npm](https://www.npmjs.com/get-npm) and
     ```
     $ brew install openssl@1.1
     ```
-    
+
+### Manual Install:
```
ashleymichal: 👍

ashleygwilliams:
merge at will!

PR #285, opened 2019-07-01
Use serde for metadata

xtuc:
This change adds proper construction of the worker metadata, previously
it was an error-prone string.

`bindings.rs` is used to build certain binding type, currently only
wasm_module is supported.

### DIFF
--- a/src/commands/build/wranglerjs/bundle.rs
+++ b/src/commands/build/wranglerjs/bundle.rs
@@ -9,6 +9,8 @@ use std::path::{Path, PathBuf};
 use log::info;
 
 use crate::commands::build::wranglerjs::output::WranglerjsOutput;
+use crate::settings::binding::Binding;
+use crate::settings::metadata;
 use crate::terminal::message;
 
 // Directory where we should write the {Bundle}. It represents the built
@@ -50,8 +52,9 @@ impl Bundle {
 
         script_file.write_all(script.as_bytes())?;
 
+        let metadata = create_metadata(self).expect("could not create metadata");
         let mut metadata_file = File::create(self.metadata_path())?;
-        metadata_file.write_all(create_metadata(self).as_bytes())?;
+        metadata_file.write_all(metadata.as_bytes())?;
 
         // cleanup {Webpack} dist, if specified.
         if let Some(dist_to_clean) = &wranglerjs_output.dist_to_clean {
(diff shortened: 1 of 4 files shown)


cwndrws:
This looks pretty good to me, except could we move this out of the `wranglerjs` module, as it will be used for all types of workers?

xtuc:
Good point @cwndrws, I moved the metadata and bindings into settings.

src/commands/build/wranglerjs/bundle.rs
```diff
@@ -50,8 +52,9 @@ impl Bundle {
 
         script_file.write_all(script.as_bytes())?;
 
+        let metadata = create_metadata(self).expect("could not create metadata");
         let mut metadata_file = File::create(self.metadata_path())?;
-        metadata_file.write_all(create_metadata(self).as_bytes())?;
+        metadata_file.write_all(metadata.as_bytes())?;
```
cwndrws: I know this didn't originate from this PR, but why do we write the metadata to a file?

ashleymichal:
probably not now, but settings is ripe for some extractions.

src/commands/build/wranglerjs/bundle.rs
granjef3: I think its useful if you want to use wrangler as a "build-only" tool

cwndrws: But metadata isn't used unless you're publishing to Cloudflare,  so we support building with wrangler and then publishing with curl or something?

granjef3: Maybe as part of a CI process where the CI handles upload because wrangler doesn't support complex config for handling staging environments etc. I think it's useful for all "files" that we upload for multipart to end up on disk somewhere for this reason.

ashleymichal: does it stay on disk? or is it cleaned up later?

src/commands/build/wranglerjs/bundle.rs
```diff
@@ -109,31 +112,20 @@ pub fn create_prologue() -> String {
...
-    } else {
-        r#"
-            {
-                "body_part": "script"
-            }
-        "#
-        .to_string()
+        bindings.push(Binding::new_wasm_module(
+            bundle.get_wasm_binding(),
+            bundle.get_wasm_binding(),
```
ashleymichal: twice?

xtuc: Once is the name and the other the multipart part.

src/commands/build/wranglerjs/bundle.rs
xtuc: There was some back and forth with the metadata file lately, see #276. The file is store in ./worker as part of the worker "bundle", it is persistent.

I'm sure some people will commit their worker, it make sense to have a consistent state including the metadata, it's useful for the preview and debugging too.

xtuc:
@ashleymichal 
> probably not now, but settings is ripe for some extractions.

I don't feel strong about the directory, feel free to suggest something else.

ashleymichal:
> @ashleymichal
> 
> > probably not now, but settings is ripe for some extractions.
> 
> I don't feel strong about the directory, feel free to suggest something else.

I think it's just fine for now, it's sort of a "needs design" thing that will be relatively painless to refactor after the fact. will approve

src/settings/binding.rs
```diff
@@ -0,0 +1,14 @@
...
+
+#[derive(Serialize, Debug)]
+#[serde(tag = "type")]
+pub enum Binding {
+    #[allow(non_camel_case_types)]
+    wasm_module { name: String, part: String },
+}
+
+impl Binding {
+    pub fn new_wasm_module(name: String, part: String) -> Binding {
```
ashleymichal: this seems odd; why would we want a wasm-specific method on the binding trait? or am i reading this wrong?

xtuc: The idea is too add for each binding type a method (like a factory, constructor), see https://github.com/cloudflare/wrangler/pull/215/files#diff-deeb786858380f08f8f4d40680620a9a if it makes more sense.

ashleymichal: ok that makes sense, thank you.

granjef3:
> @ashleymichal
> 
> > probably not now, but settings is ripe for some extractions.
> 
> I don't feel strong about the directory, feel free to suggest something else.

I think bundle deserves its own module, so there is a home within the code for "things that wrangler writes out at the end of a build"

ashleygwilliams:
let's land . this . after an approving review from @ashleymichal

PR #286, opened 2019-07-01
Change global config perm

xtuc:
Avoid the global user config to be system readable, restrict any access
to the current user and only in read only (600).

Merge https://github.com/cloudflare/wrangler/pull/239 first.

### DIFF
--- a/src/commands/config.rs
+++ b/src/commands/config.rs
@@ -1,7 +1,21 @@
 use crate::terminal::message;
 use std::fs;
+use std::fs::File;
+#[cfg(not(target_os = "windows"))]
+use std::os::unix::fs::PermissionsExt;
+use std::path::PathBuf;
 
 use crate::settings::global_user::{get_global_config_dir, GlobalUser};
+
+// set the permissions on the dir, we want to avoid that other user reads to
+// file
+#[cfg(not(target_os = "windows"))]
+pub fn set_file_mode(file: &PathBuf) {
+    File::open(&file)
+        .unwrap()
+        .set_permissions(PermissionsExt::from_mode(0o600))
+        .expect("could not set permissions on file");
+}
 
 pub fn global_config(email: &str, api_key: &str) -> Result<(), failure::Error> {
     let s = GlobalUser {
@@ -16,6 +30,10 @@ pub fn global_config(email: &str, api_key: &str) -> Result<(), failure::Error> {
 
     let config_file = config_dir.join("default.toml");
     fs::write(&config_file, &toml)?;
+
+    // set permissions on the file
+    #[cfg(not(target_os = "windows"))]
+    set_file_mode(&config_file);
 
     message::success(&format!(
         "Successfully configured. You can find your configuration file at: {}",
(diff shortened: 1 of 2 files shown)


src/commands/config.rs
```diff
@@ -33,3 +57,21 @@ pub fn global_config(email: &str, api_key: &str) -> Result<(), failure::Error> {
...
+
+    #[test]
+    fn it_creates_file_with_perms() {
+        let dir = env::temp_dir().join("wrangler_test_config_dir");
+        global_config("email", "api_key", Some(dir.clone())).expect("could not create file");
+        assert_eq!(dir.exists(), true);
+
+        let metadata = dir.metadata().unwrap();
+        let permissions = metadata.permissions();
+        assert_eq!(permissions.mode(), CONFIG_PERM);
```
xtuc: Currently this fails with:
```
  left: `16893`,
 right: `256`', src/commands/config.rs:74:9
```

It's 40755 in octal which is not what I set :man_shrugging:

src/commands/config.rs
```diff
@@ -1,27 +1,51 @@
...
+#[cfg(not(target_os = "windows"))]
+use std::os::unix::fs::PermissionsExt;
+use std::path::{Path, PathBuf};
 
 use crate::emoji;
+use crate::install::target;
 use crate::settings::global_user::GlobalUser;
 
-pub fn global_config(email: &str, api_key: &str) -> Result<(), failure::Error> {
+static CONFIG_PERM: u32 = 0o400;
```
cwndrws: People may want to edit their config. Probably better to make this `600`, right?

xtuc: That's a good point, I believe I already did it too.

src/commands/config.rs
xtuc: That said, It seems that permissions don't apply.

tests/config.rs
```diff
@@ -0,0 +1,62 @@
...
+    assert_eq!(
+        config,
+        r#"email = "a"
+api_key = "b"
+"#
+    );
+
+    // check dir permissions (but not on windows)
+    if !cfg!(target_os = "windows") {
+        let mut command = Command::new("stat");
```
xtuc:
```rust
let metadata = config_file.clone().metadata().unwrap();
let permissions = metadata.permissions();
assert_eq!(permissions.mode(), 0o600);
```

The API returned odd modes, that's why I used `stat`.

EverlastingBugstopper:
Could you explain a bit more what this is doing? Having a hard time grokking from title + comment. Does this have to do with how wrangler is installed?

xtuc:
sent on gchat.

tests/config.rs
```diff
@@ -0,0 +1,62 @@
...
+    let config = fs::read_to_string(&config_file).expect("could not read config");
+    assert_eq!(
+        config,
+        r#"email = "a"
+api_key = "b"
+"#
+    );
+
+    // check dir permissions (but not on windows)
+    if !cfg!(target_os = "windows") {
```
cwndrws: What do we do on Windows?

xtuc: I don't know, cc @steveklabnik @ashleygwilliams?

xtuc:
@ashleygwilliams https://github.com/cloudflare/wrangler/pull/286/files#diff-ba20ca15ebf8c5833a1500e22a42dadc

Issue #280, opened 2019-06-27
Always use multipart upload


ashleymichal:
On upload, we have some conditional logic around what kind of request to make to the Workers API based on whether or not we are sending bindings, however the service will handle an empty list of bindings just fine, so let's reduce our (future) API surface area by just always composing a multi-part request for worker uploads!

PR #292, opened 2019-07-02
add some tests

xtuc:
targeting #215


xtuc:
I would suggest to merge this PR into #215 and merge #215 as well. So we can do the bundle refactor (#293) more easily.

Cargo.toml
```diff
@@ -39,3 +39,4 @@ fs_extra = "1.1.0"
 
 [features]
 vendored-openssl = ['openssl/vendored']
+slow_tests = []
```
ashleymichal: what is this for?

src/commands/build/mod.rs
```diff
@@ -1,20 +1,64 @@
...
 use std::env;
+use std::fs;
+use std::fs::File;
+use std::io::prelude::*;
 use std::path::PathBuf;
 use std::process::Command;
 
 use crate::terminal::message;
 
+fn write_kv_metadata(maybe_kv_namespaces: &Option<Vec<KvNamespace>>) -> Result<(), failure::Error> {
```
ashleymichal: is there any important difference between a `None` and a `Vec<KvNamespace>` of length 0?

tests/build.rs
```diff
@@ -136,6 +136,73 @@ fn it_builds_with_webpack_wast() {
...
+
+    let metadata = fs::read_to_string(fixture_out_path(fixture).join("metadata.json"))
+        .expect("could not read metadata");
+    assert_eq!(metadata, r#"{"body_part":"script","bindings":[{"type":"kv_namespace","name":"bind","namespace_id":"id"}]}"#);
+    cleanup(fixture);
+}
+
+#[cfg(feature = "slow_tests")]
+#[test]
+fn it_builds_with_kv_metadata_rust() {
```
ashleymichal: is there also a test for `it_builds_with_wasm_bindings_rust`? because reading the code it looks like maybe no?

Cargo.toml
xtuc: The test for rust is very slow, so I wanted to exclude it when running locally. It's enabled in the CI.

src/commands/build/mod.rs
xtuc: The thing is that the vector is not present until you specify the key in the config. In the webpack backend I have a unwrap_or_else approach, feel free to indicate which one is prefered.

tests/build.rs
xtuc: Yes I agree, we don't yet.

ashleymichal:
the problem with this PR is that it puts metadata generation in the build step, whereas in a normal rustwasm project you will find metadata generation in the publish step; which means that in practice actually you will write kv namespace bindings and then write over them immediately and end up with just your wasm binding. I think we need to actually allow the multipart form data step to be its own thing; additionally there should not be a step "write kv namespace bindings", but rather a step that takes a project and adds all bindings at once.

xtuc:
> however you will find metadata generation in the publish step

Metadata generation is back at build time #276.

> and then write over them immediately and end up with just your wasm binding

Rust and JavaScript doesn't have metadata auto generation, this PR is introducing it for KV.

> I think we need to actually allow the multipart form data step to be its own thing

I added a mention about the multipart from in https://github.com/cloudflare/wrangler/issues/293. I don't know if it's related to this PR.

> perhaps each project needs to be able to handle serializing itself.

This is currently the case but there no need to, that's why #280 was made.

xtuc:
TODO: keep the tests but remove the kv_namespace logic

xtuc:
Only adds a test, merging in feat-better-kv now.

Issue #240, opened 2019-06-11
document using npm with a version manager to avoid permissions errors


kentonv:
https://workers.cloudflare.com says to use `npm install -g @cloudflare/wrangler` to install, but on Linux that doesn't work because only root can perform global installs. The usual thing to do is put `sudo` in front of it, but that doesn't work either, producing the log shown below.

It appears the problem is that Wrangler's post-install script is trying to put stuff in the user's home directory. Fundamentally, `install -g` shouldn't be doing any per-user stuff; the whole point is to install for all users. The permission error seems to occur because [when `npm install` is run as root, it actually runs the post-install script as `nobody`](https://docs.npmjs.com/misc/scripts#user). This seems... bizarre, since I'm not sure what a post-install script could usefully do when running as `nobody`, but there you go. In any case, `homedir()` still returns `/root` (despite the active user being `nobody`), and of course `nobody` cannot access `/root`, hence the permissions error.

But even if the post-install script did run as root, it seems that Wrangler would end up installing files into the root user's home directory, where they would be inaccessible to any other user and thus not very useful.

Non-global installs appear work, with the caveat that the wrangler binary gets placed in `$HOME/.wrangler` with no versioning, so separate projects depending on different versions of Wrangler will interfere with each other. IMO it would be better to place the binary somewhere under the module's installation location in `node_modules`.

```
kenton@zero:~$ sudo npm install -g @cloudflare/wrangler
/usr/bin/wrangler -> /usr/lib/node_modules/@cloudflare/wrangler/run-wrangler.js

> @cloudflare/wrangler@1.0.2 postinstall /usr/lib/node_modules/@cloudflare/wrangler
> node install-wrangler.js

fs.js:115
    throw err;
    ^

Error: EACCES: permission denied, mkdir '/root/.wrangler'
    at mkdirSync (fs.js:753:3)
    at Object.<anonymous> (/usr/lib/node_modules/@cloudflare/wrangler/install-wrangler.js:60:3)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:742:12)
[...]

PR #298, opened 2019-07-02
[WIP][RFC] Use Homebrew instead of NPM

granjef3:
Using npm to distribute our binary is prone to issues (#240) since misconfigured npm installs are common. Fixing this opens a whole can of worms, due to the way npm treats postinstall scripts. Making this work would require making our own cross-platform installer for wrangler and maintaining it.

Homebrew does a nice job of handling this, works on Linux, macOS, and Windows (for WSL users), and will also install our dependencies (node and openssl@1.1) for the user. Users who prefer `nvm` can just put `nvm`'s node in front of `brew` node in their PATH. Users who dislike homebrew can use our manual install method for now.

While we evaluate the pros/cons of creating our own installer, I believe this is a potential solution to provide a consistent cross-platform solution for users.

My proposed brew formula `cf-wrangler`:

```
class CfWrangler < Formula
  desc "Management tool for Cloudflare Workers"
  homepage "https://github.com/cloudflare/wrangler/"
  url "https://github.com/cloudflare/wrangler/archive/v1.0.0.tar.gz"
  sha256 "f3cfd04c0d69372fa1bb09e8e23200e7da7988cd9767f107daa73771a380d096"

  depends_on "rust" => :build
  depends_on "node"
  depends_on "openssl@1.1"

  def install
    system "cargo", "install", "--root", prefix, "--path", "."
  end

  test do
    system "wrangler", "generate"
    system "cd", "worker"
    system "wrangler", "build"
  end
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,10 +11,10 @@
 
 You have many options to install wrangler!
 
-### Using `npm`
+### Using [Homebrew](https://brew.sh/)
 
 ```
-npm i @cloudflare/wrangler -g
+brew install cf-wrangler
 ```
 
 ### Using `cargo`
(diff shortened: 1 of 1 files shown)


xtuc:
Could we remove the npm package entirely then?

steveklabnik:
As far as I'm concerned, recommending only homebrew is a non-starter for Windows. Its usage is extremely tiny, and asking people to install a whole package management system in order to install our thing will cause people to bounce off.

ashleygwilliams:
Homebrew is first and foremost understood in the community as a MacOS operating system package manager. This tool needs to support Windows and Linux. I do not think that the burden of asking folks to install a node version manager for correct behavior (has multiple other benefits to them) is more than the burden of asking a Windows or Linux user to use brew (has little to no other benefits and support on Windows is only via WSL, which is a non starter). As a result, we cannot take this path at this time.

I am going to close this RFC. I would recommend that we move forward with the standalone installer and focus our design efforts on improving that experience.

PR #299, opened 2019-07-03
kv-namespace bindings

xtuc:


xtuc:
Why this PR has been closed?

ashleymichal:
Thanks for asking, I should have included a proper explanation:

We stepped back and worked out a design to solve the problem, including identifying a few refactors that will be beneficial. Because of the scope of the design we came up with, and the steps we identified, we felt it would be easier to start off of the master branch, using these two PRs as reference.

Issue #293, opened 2019-07-02
RFC: expand usage of the Bundle


xtuc:
## Problem

In #215 #292 we want to generate the bindings for every backend, regardless of their implementations. For wranglerjs I introduced the notion of the bundle; representing the backend's output and a worker.

## Suggestion

We should move the bundle struct and base impl out of  the `webpack` backend. It contains the following information: path to the build dir, path to the script, path to the wasm, has a wasm module?. 

The backend should call write before handing the struct to wrangler, so that it just has to upload the bundle.

Conceptually the result of the backend computation is a bundle.

### Creating the metadata file

The metadatas (including bindings) are not included into the bundle, that said the bundle gives information about it. Wrangler should write the metadata and the bindings.

### Creating the multipart form

This is not part of the bundle, but rather Wrangler's logic once it gets the bundle from a backend.

granjef3:
I would like to see this result in the bundle being passed as an output from the build command, such that the publish and preview commands would take it as an argument

cwndrws:
I believe bundle is a webpack-specific term, so we probably shouldn't overload it if we're going to use it as a top-level type that contains everything that makes up a worker.

xtuc:
I believe other teams already uses WorkerBundle, does it make sense to use it? or Worker?

granjef3:
> I believe bundle is a webpack-specific term, so we probably shouldn't overload it if we're going to use it as a top-level type that contains everything that makes up a worker.

This is a good point too -- maybe we can do a s/bundle/worker or other similar name? Would be nice to distinguish between your worker as in the code before compile, and a worker as in the thing we upload

cwndrws:
I had started working in this direction and have a struct called `Worker`. It contains a script path and a list of bindings. My goal was to have some sort of API client that took a `Worker` no matter what type and uploaded it.

cwndrws:
Here's the `Worker` type I was talking about: 

```rust
use reqwest::multipart;
use serde::Serialize;

pub struct Worker {
    script_path: String,
    bindings: Vec<Binding>,
}

impl Worker {
    pub fn multipart(&self) -> Result<multipart::Form, failure::Error> {
        let mut form = multipart::Form::new().file("script", &self.script_path)?;
        let binding_parts = binding_parts(&self.bindings)?;
        for (name, part) in binding_parts {
            form = form.part(name, part);
        }
        Ok(form)
    }
}

fn binding_parts<'a>(
[...]

PR #300, opened 2019-07-03
[WIP] refactor: WorkerBundle

xtuc:
refactor: WorkerBundle

This change introduce the notion of the WorkerBundle. Conceptually it
represents result of the backend (javascript, rust or webpack) computation.

The abstraction provide generalization for creating metadata and
uploading the worker to the API using a multipart form.

To maintain compatiblity with the rust backend, the WorkerBundle can be
contructed with a file in `metadata_path` property. We will use the file
directly instead of generating the metadata from the bindings.

For easier testing WorkerBundle allow to pass a `out_root` property
which will be used as base directory when persisting the bundle files.

Fixes https://github.com/cloudflare/wrangler/issues/293


src/commands/build/wranglerjs/output.rs
```diff
@@ -77,3 +125,119 @@ mod tests {
         assert_eq!(wranglerjs_output.wasm_size(), "3 bytes");
     }
 }
+
```
cwndrws: Probably shouldn't check in commented out tests

src/commands/build/mod.rs
```diff
@@ -1,17 +1,29 @@
+// FIXME: move
```
cwndrws: It's not clear what needs to happen here.

src/worker_bundle.rs
```diff
@@ -0,0 +1,107 @@
...
+use crate::settings::metadata;
+use crate::terminal::message;
+
+const WORKER_BUNDLE_OUT: &str = "./worker";
+
+#[derive(Debug)]
+pub struct WorkerBundle {
+    // out: String,
+    pub script_path: PathBuf,
+    pub wasm_path: Option<PathBuf>,
```
cwndrws: It seems like the wasm stuff should get owned by a binding rather than being top-level attributes of a WorkerBundle.

src/commands/build/wranglerjs/output.rs
```diff
@@ -20,33 +29,72 @@ pub struct WranglerjsOutput {
...
-    pub fn wasm_size(&self) -> String {
-        let size = self.wasm.to_owned().unwrap().len();
-        match NumberPrefix::decimal(size as f64) {
-            Standalone(bytes) => format!("{} bytes", bytes),
-            Prefixed(prefix, n) => format!("{:.0} {}B", n, prefix),
-        }
-    }
+// We inject some code at the top-level of the Worker; called {prologue}.
+// This aims to provide additional support, for instance providing {window}.
+pub fn create_prologue() -> String {
```
cwndrws: This seems out of scope for this change.

src/commands/build/mod.rs
xtuc: We can move the krate and package code into build because it's captured by WorkerBundle.

src/commands/build/wranglerjs/output.rs
xtuc: I just moved the code from the bundle.rs file, that I deleted in favor of the shared WorkerBundle.

src/worker_bundle.rs
xtuc: Your approach make sense to me now. I was hesitant to modify the binding struct because it doesn't strictly match with the online bindings anymore, but I guess it's fine.

ashleymichal:
closing in favor of incoming pr

xtuc:
What was the issue with this PR?

ashleymichal:
Thanks for asking, I should have included a proper explanation:

We stepped back and worked out a design to solve the problem, including identifying a few refactors that will be beneficial. Because of the scope of the design we came up with, and the steps we identified, we felt it would be easier to start off of the master branch, using these two PRs (this and #299 ) as reference.

PR #309, opened 2019-07-08
Use rust template for rust project

### LINKED ISSUE #277: Cannot generate new Rust project
From a clean install via `cargo install wrangler` I cannot have wrangler generate a rust based project.

Have tried:

* `wrangler generate` => webpack project (as expected)
* `wrangler generate foo` => webpack project (in `/foo`, as expected)
* `wrangler generate foo --type=rust` => webpack project
* `wrangler generate foo --type="rust"` => webpack project
* `wrangler generate --type="rust"` => webpack project

None of these seems to result in a rust project being generated. Started out from the [Cloudflare tutorial](https://developers.cloudflare.com/workers/webassembly/tutorial/), but it made no mention of `--type` but found that in this repos docs.

Is this expected behaviour? Am I being an idiot?

```
~/src/tmp $ wrangler -V
👷  ✨   wrangler 1.0.0

~/src/tmp $ rustup show
Default host: x86_64-apple-darwin

installed toolchains
--------------------

stable-x86_64-apple-darwin
nightly-x86_64-apple-darwin (default)

installed targets for active toolchain
--------------------------------------

wasm32-unknown-unknown
x86_64-apple-darwin

active toolchain
----------------

nightly-x86_64-apple-darwin (default)
rustc 1.35.0-nightly (3750348da 2019-04-08)

~/src/tmp $ cargo --version
cargo 1.35.0-nightly (6f3e9c367 2019-04-04)

~/src/tmp $ node --version
v11.4.0
```

On OSX via HyperTerm

EverlastingBugstopper:
This PR will use the template at https://github.com/cloudflare/rustwasm-worker-template when `—type rust` is passed to `wrangler generate`

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -15,6 +15,7 @@ mod installer;
 mod settings;
 mod terminal;
 
+use crate::settings::project::ProjectType;
 use terminal::emoji;
 
 fn main() -> Result<(), failure::Error> {
@@ -167,12 +168,19 @@ fn main() -> Result<(), failure::Error> {
     } else if let Some(matches) = matches.subcommand_matches("generate") {
         let name = matches.value_of("name").unwrap_or("worker");
         let project_type = match matches.value_of("type") {
-            Some(s) => Some(settings::project::ProjectType::from_str(&s.to_lowercase())?),
+            Some(s) => Some(ProjectType::from_str(&s.to_lowercase())?),
             None => None,
         };
-        let template = matches
-            .value_of("template")
-            .unwrap_or("https://github.com/cloudflare/worker-template");
+
+        let default_template = "https://github.com/cloudflare/worker-template";
+        let template = matches.value_of("template").unwrap_or(match project_type {
+            Some(ref pt) => match pt {
+                ProjectType::Rust => "https://github.com/cloudflare/rustwasm-worker-template",
+                _ => default_template,
+            },
+            _ => default_template,
+        });
+
         info!(
             "Generate command called with template {}, and name {}",
             template, name


src/main.rs
```diff
@@ -167,12 +168,19 @@ fn main() -> Result<(), failure::Error> {
     } else if let Some(matches) = matches.subcommand_matches("generate") {
         let name = matches.value_of("name").unwrap_or("worker");
         let project_type = match matches.value_of("type") {
-            Some(s) => Some(settings::project::ProjectType::from_str(&s.to_lowercase())?),
+            Some(s) => Some(ProjectType::from_str(&s.to_lowercase())?),
             None => None,
```
ashleymichal: i know it's not part of this PR, but based on the deeper layers of this code, it should be possible to always return just a bare `ProjectType` rather than an `Option` here; the `generate` command should also be capable of taking this. That would get rid of the nested match below and clean this up. iirc, the default would be `webpack`

src/main.rs
```diff
@@ -167,12 +168,19 @@ fn main() -> Result<(), failure::Error> {
...
-            .unwrap_or("https://github.com/cloudflare/worker-template");
+
+        let default_template = "https://github.com/cloudflare/worker-template";
+        let template = matches.value_of("template").unwrap_or(match project_type {
+            Some(ref pt) => match pt {
+                ProjectType::Rust => "https://github.com/cloudflare/rustwasm-worker-template",
+                _ => default_template,
+            },
+            _ => default_template,
+        });
```
ashleymichal: cleaning up the above would allow you to define the default template as a single match against project type, and then put that in the `unwrap_or()`:

```
let default_template = match project_type {
    ProjectType::Rust => "https://github.com/cloudflare/rustwasm-worker-template",
    _ => "https://github.com/cloudflare/worker-template";
}

let template = matches.value_of("template").unwrap_or(default_template);
```

src/main.rs
ashleymichal: reviewed the code again. turns out this refactor depends on the incoming solution for #315.

PR #313, opened 2019-07-09
Consolidate multipart

cwndrws:
This contains a some refactors and consolidation of all PUT bodies to use multipart regardless of project type.


src/commands/publish/mod.rs
```diff
@@ -1,33 +1,30 @@
...
 
 pub fn publish(user: &GlobalUser, project: &Project, release: bool) -> Result<(), failure::Error> {
     info!("release = {}", release);
 
     validate_project(project, release)?;
-    commands::build(&project)?;
+
+    let worker = project.worker()?;
+
     create_kv_namespaces(user, &project)?;
```
ashleymichal: we should likely remove this, it's unofficially dead code.

src/settings/project/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+            }
+        };
+    }
+    Ok(Resource::WasmModule(WasmModule {
+        path: wasm_path.to_string(),
+        binding: binding,
+    }))
+}
+
+pub fn get_project_config() -> Result<Project, failure::Error> {
```
ashleymichal: this no longer needs to be a public function; users of this crate should call `Project::new` instead.

src/commands/publish/mod.rs
```diff
@@ -1,33 +1,26 @@
-mod krate;
 pub mod preview;
 mod route;
+pub mod worker_bundle;
```
granjef3: I feel like `worker_bundle` should be a member of the new `workers` crate you have created, maybe under a `publish` section as you have done with the old `commands::build` logic.

ashleymichal: i had a similar thought. considering landing this pr and making the additional refactor separately so we can also get kv config landed.

cwndrws:
The comment got deleted, but I'll respond anyways. I agree that the WorkerBundle stuff might not be in the right place, but the reason I don't like it alongside the Worker type, is that it's an implementation detail of publish/preview not the concept of a Worker.

src/workers/build/mod.rs
```diff
@@ -0,0 +1,151 @@
+mod krate;
+
+use std::fs;
+use std::io::BufReader;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+use super::{Resource, Script, WasmModule, Worker};
```
granjef3: I think WorkerBundle could become `/publish/mod.rs` that has `impl` to return the multipart form from a `Worker`, and get rid of the separate data structure

src/commands/publish/mod.rs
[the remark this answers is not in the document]
granjef3: > The comment got deleted, but I'll respond anyways. I agree that the WorkerBundle stuff might not be in the right place, but the reason I don't like it alongside the Worker type, is that it's an implementation detail of publish/preview not the concept of a Worker.

context: I was imagining a future where the workers crate represents the workers API, and you could build/upload/preview a `Worker`

src/commands/publish/worker_bundle.rs
```diff
@@ -0,0 +1,113 @@
+use crate::workers::{Resource, Worker};
+use reqwest::multipart;
+use serde::Serialize;
+
+#[derive(Debug, Serialize)]
```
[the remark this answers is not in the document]
cwndrws: No I don't think we need it here. Will fix.

src/commands/publish/mod.rs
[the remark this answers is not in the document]
cwndrws: I would only want to do that if we were to move all of the preview/publish logic into this crate. The worker bundle is an implementation detail of those operations and shouldn't belong in the more abstract business logic part of the code. I think a better place for this to go would be when/if we pull out a module that contains an API client.

src/commands/publish/worker_bundle.rs
```diff
@@ -0,0 +1,113 @@
...
+pub struct WorkerBundle {
+    name: String,
+    script_name: String,
+    script_path: String,
+    bindings: Vec<Binding>,
+}
+
+impl WorkerBundle {
```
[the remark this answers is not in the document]
cwndrws: The idea is one (`Worker`) is a domain object that matches the mental model of what a worker is and how wrangler deals with it. The other (`WorkerBundle`) is an implementation detail of how we have to interact with the API. We should use `Worker` in all possible places and only use `WorkerBundle` when we need to send requests to the API.

src/commands/publish/worker_bundle.rs
```diff
@@ -0,0 +1,113 @@
...
+#[derive(Serialize, Debug)]
+pub struct MetaData<'a> {
+    body_part: String,
+    bindings: &'a [Binding],
+}
+
+// setting the serde tag to type makes it so the "type" key in the resulting json object is the
+// same as the enum member name.
+#[derive(Serialize, Debug)]
+#[serde(tag = "type")]
```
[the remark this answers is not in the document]
cwndrws: What I would need to do this would be `#[serde(rename_all = "snake_case")]`, which doesn't exist as far as I know.

src/workers/build/mod.rs
cwndrws: I disagree, separating the domain entity of a Worker from the implementation of our API requests provides value.

cwndrws: I do think it would make sense to further improve the organization of the code, though.

ashleymichal:
we need a rebase, and I'd like to give @ashleygwilliams the day (Friday) to review this, but other than that I think this is good.

src/commands/publish/mod.rs
```diff
@@ -43,46 +36,10 @@ pub fn publish(user: &GlobalUser, project: &Project, release: bool) -> Result<()
...
-                }
-            }
-            info!("Namespace '{}' exists now", namespace)
-        }
-    }
-    Ok(())
-}
-
-fn publish_script(
+fn publish_worker(
```
ashleygwilliams: we don't ever publish anything that is not a worker- publish vs publish worker here is not clear

src/commands/publish/worker_bundle.rs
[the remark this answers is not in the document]
ashleygwilliams: i would prefer renames to the clippy allows here and it should be possible.

src/commands/publish/worker_bundle.rs
```diff
@@ -0,0 +1,113 @@
...
+    wasm_module(WasmBinding),
+    #[allow(non_camel_case_types)]
+    kv_namespace(KVBinding),
+}
+
+impl From<Resource> for Binding {
+    fn from(resource: Resource) -> Binding {
+        match resource {
+            Resource::WasmModule(wasm) => Binding::wasm_module(WasmBinding {
+                path: wasm.path.clone(),
```
ashleygwilliams: are the clones necessary?

src/main.rs
```diff
@@ -14,6 +14,7 @@ mod install;
 mod installer;
 mod settings;
 mod terminal;
+mod workers;
```
ashleygwilliams: this should be singular in my opinion

src/workers/build/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+            wranglerjs::run_build(project)?;
+        }
+    }
+
+    let worker = match project.project_type {
+        ProjectType::Rust => wasm_worker(&project.name)?,
+        ProjectType::Webpack => webpack_worker(&project.name)?,
+        ProjectType::JavaScript => js_worker(&project.name)?,
+    };
+    // add other resoureces
```
ashleygwilliams: this comment is confusing to me

src/workers/build/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    let script_path = "./worker/generated/script.js";
+    let wasm_path = format!("./pkg/{}_bg.wasm", name);
+
+    let wasm_resource = wasm_resource(wasm_path)?;
+
+    Ok(Worker {
+        name: project_name.to_string(),
+        script: Script {
+            name: "script".to_string(),
```
ashleygwilliams: is this ever not hardcoded? i'd prefer to make it a default than put it in the code like this

src/workers/build/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+    let script_path = "./worker/script.js";
+    let mut worker = Worker {
+        name: project_name.to_string(),
+        script: Script {
+            name: "script".to_string(),
+            path: script_path.to_string(),
+        },
+        resources: Vec::new(),
+    };
+    let wasm_path = "./worker/module.wasm";
```
ashleygwilliams: is this always hardcoded? we should make it default or a const

src/workers/build/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+fn wasm_resource(wasm_path: String) -> Result<Resource, failure::Error> {
+    let metadata_path = "./worker/metadata_wasm.json";
+    let mut binding = "wasm".to_string();
+
+    if Path::new(metadata_path).exists() {
+        let file = fs::File::open(metadata_path)?;
+        let reader = BufReader::new(file);
+        let data: serde_json::Value = serde_json::from_reader(reader)?;
+        binding = match data["bindings"][0]["name"].as_str() {
+            Some(s) => s.to_string(),
```
ashleygwilliams: i would probably just use ? here for now

src/workers/build/mod.rs
```diff
@@ -0,0 +1,151 @@
+mod krate;
```
ashleygwilliams: this is file is getting long, we may want to split it up a bit

src/workers/build/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+
+fn build_generated_dir() -> Result<(), failure::Error> {
+    let dir = "./worker/generated";
+    if !Path::new(dir).is_dir() {
+        fs::create_dir("./worker/generated")?;
+    }
+    Ok(())
+}
+
+fn command(args: &[&str], binary_path: PathBuf) -> Command {
```
ashleygwilliams: don't fully understand why this is here,as it relates to how wrangler calls commands

src/commands/publish/mod.rs
cwndrws: Would `publish_bundle` be better?

src/main.rs
cwndrws: It was at one point and I think @ashleymichal wanted it to be plural, but I can't remember why. Maybe she could chime in?

src/workers/build/mod.rs
cwndrws: This is where we'll add KV namespaces once we have that configuration in the `Project` struct. I can clarify this comment or remove it.

src/workers/build/mod.rs
cwndrws: I'll make this a default instead. This was more or less a direct port of what the previous code was doing.

src/workers/build/mod.rs
cwndrws: I think a larger refactor needs to happen where we don't make so many assumptions about where built files end up or at a minimum isolate those assumptions to a build module or something. This would make it so calling build would return all the information we need to know about the built files instead of assuming it like here.

src/workers/build/mod.rs
cwndrws: It was in the build command module, but was moved here as it's more or less business logic for building workers and shouldn't be in a command module. It's just here because this is the only file that references it. Do you have a preferred location?

ashleymichal: let's create a top-level `utils` folder (not my favorite convention but appropriate here) and export it from there.

src/main.rs
ashleymichal: the idea was to distinguish based on the platform (Workers is the name of the platform/product) in direct contrast to, say, KV, which is another platform Wrangler will interact with. @ashleygwilliams and I discussed this morning and there is a suggested structure of project that may help.

src/commands/publish/worker_bundle.rs
cwndrws: Just the one on the first `wasm.binding`, since it's used twice. Removing the others. Can't take references here since `resource` is consumed by this function. I can change this to move away from the `From` trait to not consume if you think that would be cleaner.

src/workers/build/mod.rs
ashleymichal: yes; either build needs to return built asset paths, or asset paths need to be part of configuration (whether exposed or not).

src/workers/build/mod.rs
cwndrws: Wouldn't I have to implement `From<NoneError>`? is that possible? I don't own the definition of `NoneError` or `failure::Error`, so I don't think I can implement either `From` or `Into` for this. Maybe I'm misunderstanding something.

src/commands/publish/worker_bundle.rs
ashleygwilliams: nah, per in air convo the one clone is totally fine :)

src/workers/build/mod.rs
ashleygwilliams: i'm up for the clarification or removing it- up to you!

ashleymichal:
closing in favor of #329

Issue #323, opened 2019-07-16
Support webpack.config.js exporting functions and promises


third774:
According to the [webpack docs](https://webpack.js.org/configuration/configuration-types/) `webpack.config.js` should also be able to export a function that returns a config object, a promise that resolves to a config object, or a function that returns a promise of a config object.

I've found the function pattern to be useful for being able to easily compose various flags when doing local development with a `webpack.config.js` that looks like this:

```js
require("dotenv").config();

const webpack = require("webpack");
const BundleAnalyzerPlugin = require("webpack-bundle-analyzer")
  .BundleAnalyzerPlugin;

const DEFAULTS = {
  runtime: "webworker"
};

module.exports = (env = DEFAULTS, argv = { mode: "production" }) => {
  console.log("cliEnv", env);
  const { mode } = options;
  const mergedEnv = {
    ...DEFAULTS,
    ...env
  };
  return {
    mode,
    entry: "./src/index.ts",
    output: {
      path: __dirname,
      filename: "index.js"
    },
    module: {
      rules: [{ test: /\.(ts|js)$/, use: "babel-loader" }]
    },
    resolve: {
      extensions: [".ts", ".js"]
    },
    target: mergedEnv.runtime,
    externals: [],
    plugins: [
      mergedEnv.analyze && new BundleAnalyzerPlugin(),
      new webpack.EnvironmentPlugin({
        RUNTIME: mergedEnv.runtime,
        OMDB_API_KEY: process.env.OMDB_API_KEY,
        SPREADSHEET_ID: process.env.SPREADSHEET_ID
      })
    ].filter(Boolean)
  };
};
```

And a `package.json` that looks like this:

```json
{
  "name": "movie-night-data-worker",
  "version": "1.0.0",
  "description": "A worker for pulling movie data from a Google Sheet and the OMDb API",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1",
    "format": "prettier --write '**/*.{js,css,json,md}'",
    "build": "webpack --mode=production --env.runtime=webworker",
    "dev": "webpack --mode=development --watch --env.runtime=node",
    "analyze": "webpack --mode=development --watch --env.analyze --env.runtime=webworker"
  },
  "author": "Kevin Kipp <kevin.kipp@gmail.com>",
  "license": "MIT",
  "devDependencies": {
    "husky": "^2.1.0",
    "prettier": "^1.17.0",
[...]

PR #325, opened 2019-07-17
feat #323: Allow fn & promise from webpack config

third774:
PR for https://github.com/cloudflare/wrangler/issues/323


ashleygwilliams:
@xtuc can you please review this PR?

tests/webpack_function_promise_config_js/webpack.config.js
```diff
@@ -0,0 +1 @@
+module.exports = () => new Promise(res => res({ entry: "./index.js" }));
```
xtuc:
```suggestion
module.exports = () => Promise.resolve({ entry: "./index.js" ));
```

tests/webpack_promise_config_js/webpack.config.js
```diff
@@ -0,0 +1 @@
+module.exports = new Promise(res => res({ entry: "./index.js" }));
```
xtuc:
```suggestion
module.exports = Promise.resolve({ entry: "./index.js" });
```

tests/build.rs
```diff
@@ -42,6 +42,51 @@ fn it_builds_with_webpack_single_js() {
...
+    let fixture = "webpack_function_config_js";
+    create_temporary_copy(fixture);
+
+    settings! {fixture, r#"
+        type = "Webpack"
+    "#};
+
+    build(fixture);
+    assert!(fixture_out_path(fixture).join("script.js").exists());
+    assert!(fixture_out_path(fixture).join("metadata.json").exists());
```
xtuc: I would remove the metadata.json checks because it's not relevant here and I believe it has been removed in master or PRs.

xtuc:
Could you please add the `env` and `argv` params from the doc https://webpack.js.org/configuration/configuration-types/#exporting-a-function?

third774:
Thanks for the feedback! I'll see what I can find on the webpack front.

third774:
Would this mean we would allow for CLI arguments used on the `wrangler` cli to be passed to webpack? Or maybe filter down to a list of webpack specific flags?

For example, if I run `wrangler build --env.mode="dev"`, should I expect to see that in my `env` object?

xtuc:
That's a good question. I don't think we should support all the webpack-cli arguments in wrangler. I would pass an empty object. Would it make sense to you?

Also note that in #220 we are going to support passing an environment, it would be great if webpack could get access to it.

third774:
I think passing an empty object for `env` makes sense for now. Once https://github.com/cloudflare/wrangler/issues/220 is complete, it could definitely make sense to make that available as well.

Additionally, there could be some defaults that get set and merged with [webpack-merge](https://www.npmjs.com/package/webpack-merge). I believe `webpack-cli` does this under the hood for providing the `mode` along with a few other things.

wranglerjs/index.js
```diff
@@ -3,93 +3,100 @@ const { join } = require("path");
...
-  console.error("Error: " + msg);
-  process.exit(1);
-  return new Error("error");
-}
-
-const rawArgs = process.argv.slice(2);
-const args = rawArgs.reduce((obj, e) => {
-  if (e.indexOf("--") === -1 && e.indexOf("=") === -1) {
-    throw error("malformed arguments");
+(async function() {
```
xtuc: Just for the record, I'm not a fan of the async iife, but I think it's the best option for now.

xtuc:
The PR looks good to me, I would like to wait for #220, so we can pass it into the `env` object and make it feature complete.

wranglerjs/index.js
EverlastingBugstopper: Really? That's interesting! I'd love to hear your thoughts 😄

xtuc: The main difference is that it's not top-level code anymore, which hurts readability a in my opinion. Once we have Top-Level-Await support it will look a lot nicer.

There are a few technical differences too but they don't matter here, just a personal pref. I'm not blocking on it.

EverlastingBugstopper: image
this is the first thing i see when i google 👀

xtuc: Rich meant the proposal from 2 years ago and from a bundler percpective. Check the 2019 edit.

EverlastingBugstopper: oh well would you look at that 😋 just saw the headline before coming in to the office - this is why it's important to read the content!

third774:
@xtuc — awesome! Is there anything I need to do to fix the CI tests?

tests/build.rs
```diff
@@ -41,6 +41,48 @@ fn it_builds_with_webpack_single_js() {
     cleanup(fixture);
 }
 
+#[test]
+fn it_builds_with_webpack_function_config_js() {
+    let fixture = "webpack_function_config_js";
```
xtuc:
```suggestion
    let fixture = "webpack_function_promise_config_js";
```

tests/build.rs
```diff
@@ -41,6 +41,48 @@ fn it_builds_with_webpack_single_js() {
...
+    "#};
+
+    build(fixture);
+    assert!(fixture_out_path(fixture).join("script.js").exists());
+    cleanup(fixture);
+}
+
+#[test]
+fn it_builds_with_webpack_promise_config_js() {
+    let fixture = "webpack_promise_config_js";
```
xtuc:
```suggestion
    let fixture = "webpack_function_promise_config_js";
```

xtuc:
@third774 I fixed them. I think we should wait for https://github.com/cloudflare/wrangler/issues/385 now.

xtuc:
@third774 let's maybe not wait for env to land. I went ahead a merged this change, it will be release soon.

PR #327, opened 2019-07-17
Better error printing

granjef3:
The default `Termination` impl for `Result` uses `Debug` to print any `Result::Err` returned by `main`. This PR overrides that behavior by moving the contents of main into a wrapper function that we can print the result of ourselves using `Display` instead.
This will print errors in a more human readable fashion.

### DIFF
--- a/src/terminal/emoji.rs
+++ b/src/terminal/emoji.rs
@@ -25,6 +25,6 @@ pub static SLEUTH: Emoji = Emoji("🕵️‍♂️", "");
 pub static SPARKLES: Emoji = Emoji("✨  ", "");
 pub static SWIRL: Emoji = Emoji("🌀 ", "");
 pub static UP: Emoji = Emoji("🆙 ", "");
-pub static WARN: Emoji = Emoji("⛔ ", "");
+pub static WARN: Emoji = Emoji("⚠️", "");
 pub static WAVING: Emoji = Emoji("👋 ", "");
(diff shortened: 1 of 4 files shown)


src/terminal/emoji.rs
```diff
@@ -25,6 +25,6 @@ pub static SLEUTH: Emoji = Emoji("🕵️‍♂️", "");
 pub static SPARKLES: Emoji = Emoji("✨  ", "");
 pub static SWIRL: Emoji = Emoji("🌀 ", "");
 pub static UP: Emoji = Emoji("🆙 ", "");
-pub static WARN: Emoji = Emoji("⛔ ", "");
+pub static WARN: Emoji = Emoji("⚠️", "");
```
EverlastingBugstopper: change to ⚠

granjef3: it is, github can't unicode

granjef3:
https://github.com/tismith/exitfailure might be a better solution

src/terminal/emoji.rs
EverlastingBugstopper: You need to add a space after the emoji here to match the style of the rest. After that, this looks good to me!

granjef3:
```suggestion
pub static WARN: Emoji = Emoji("⚠️ ", "");
```

PR #328, opened 2019-07-17
Always use multipart upload

### LINKED ISSUE #280: Always use multipart upload

ashleymichal:
Closes #280 

Previously we would conditionally use multipart based on project type, but since all project types can have bindings, and since multipart works with or without bindings, this PR standardizes toward doing just that.

There is one ugly thing here that I would like advice on handling better. Commented inline.

### DIFF
--- a/src/commands/publish/mod.rs
+++ b/src/commands/publish/mod.rs
@@ -8,7 +8,7 @@ use package::Package;
 
 use log::info;
 
-use reqwest::multipart::Form;
+use reqwest::multipart::{Form, Part};
 use std::collections::HashMap;
 use std::fs;
 use std::path::Path;
@@ -92,31 +92,12 @@ fn publish_script(
 
     let client = http::auth_client(user);
 
-    let project_type = &project.project_type;
-    let mut res = match project_type {
-        ProjectType::Rust => {
-            info!("Rust project detected. Publishing...");
-            client
-                .put(&worker_addr)
-                .multipart(build_multipart_script()?)
-                .send()?
-        }
-        ProjectType::JavaScript => {
-            info!("JavaScript project detected. Publishing...");
-            client
-                .put(&worker_addr)
-                .header("Content-Type", "application/javascript")
-                .body(build_js_script()?)
-                .send()?
-        }
-        ProjectType::Webpack => {
-            info!("Webpack project detected. Publishing...");
-            client
-                .put(&worker_addr)
-                .multipart(build_webpack_form()?)
-                .send()?
-        }
-    };
+    let script_upload_form = build_script_upload_form(project)?;
+
+    let mut res = client
+        .put(&worker_addr)
+        .multipart(script_upload_form)
+        .send()?;
 
     if res.status().is_success() {
         message::success("Successfully published your script.");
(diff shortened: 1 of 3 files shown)


src/commands/publish/mod.rs
```diff
@@ -177,9 +178,22 @@ fn make_public_on_subdomain(project: &Project, user: &GlobalUser) -> Result<(),
     Ok(())
 }
 
-fn build_js_script() -> Result<String, failure::Error> {
+fn build_js_script() -> Result<Form, failure::Error> {
     let package = Package::new("./")?;
-    Ok(fs::read_to_string(package.main()?)?)
+    let script_path = package.main()?;
+    let metadata_json = r#"{"body_part":"script","bindings":[]}"#;
```
ashleymichal: this is terrible. will add PR just to address this mess.

granjef3: honestly considering this doesn't need to be dynamic for the single file js upload, I don't think this is heinous at all

src/commands/publish/mod.rs
```diff
@@ -177,9 +178,22 @@ fn make_public_on_subdomain(project: &Project, user: &GlobalUser) -> Result<(),
...
+    let form = Form::new()
+        .file("script", script_path)
+        .unwrap_or_else(|_| panic!("{} not found. Did you rename your js files?", script_path));
+
+    form.part("metadata", metadata);
+
+    Ok(form)
 }
 
 fn build_multipart_script() -> Result<Form, failure::Error> {
```
granjef3: can we make this work for webpack + rust project types?

ashleymichal: probably yes. the subject of this pr is to consistently use multipart upload. subsequent PRs will focus on unifying the way the multipart form is constructed to be more flexible.

granjef3:
Approving since combining those two functions isn't in scope for this PR

src/commands/publish/script_upload_form.rs
```diff
@@ -0,0 +1,112 @@
...
+    let bindgen_js: String = fs::read_to_string(bindgen_js_path)?.parse()?;
+
+    let worker_js: String = fs::read_to_string("./worker/worker.js")?.parse()?;
+    let js = format!("{} {}", bindgen_js, worker_js);
+
+    fs::write("./worker/generated/script.js", js.as_bytes())?;
+    Ok(())
+}
+
+fn build_webpack_form() -> Result<Form, failure::Error> {
```
xtuc: With the WorkerBundle/Worker abstraction I was able to generate the same form for every backend. All have `script` and accept a `bindinds` array (can be empty).
All the specific paths were configured by each backend.

ashleymichal: that's my next pr 😄

ashleymichal:
my only outstanding hesitation for pushing that big green button is this:

because the API uses the format of the PUT request (is it multipart?) to decide whether to clear bindings, in the existing code there is a potential edge case wherein a user could have a js type project where you configure bindings in the UI and update their JavaScript from Wrangler without overwriting their bindings. Thus, this is a potential breaking change for users who have figured this out.

src/commands/publish/mod.rs
ashleymichal: "is terrible" was likely hyperbolic, you're right. felt ooky but yah the one case is just fine.

gabbifish:
@ashleymichal could we use the "inherit" binding type to preserve already-uploaded bindings via the UI?

gabbifish:
oh whoops! my comment was too late. Something to think about for a future PR though?

EverlastingBugstopper:
@gabbifish inheritance still requires that you specify the name of the bindings to inherit so I don't think that solves this

gabbifish:
@EverlastingBugstopper ugh schucks that is true. Perhaps we could GET the preexisting worker binding from the CF API, get the names of the bindings to inherit, and then set all those bindings to type `inherit`, but that is potentially messy logic.

ashleymichal:
@gabbifish i have a thing for this... https://wiki.cfops.it/display/EW/Protect+Overwrite (coming soon as an RFC here!)

PR #329, opened 2019-07-18
feat: unify upload form building

ashleymichal:
This PR introduces structs that are eerily similar to #300 , with the main difference being that the bulk of the changes target the publish process, since that's where the multipart form needs to be built. Hopefully it is clear how this can be extended to add KV Namespaces as Bindings. If it is not, feedback is appreciated!

Further work could be done to make the job of building the Assets to be the responsibility of the Project struct, but this is already pretty big. The best thing happening here is that the form is being built based on the Project, (type for now), which leaves room for that info to come from project config (i.e. project.kv_namespaces).

Steps taken:

- WranglerJS build step no longer writes metadata.json file
- Publish step no longer looks for metadata.json file in the filesystem (another "breaking" change)
- Publish step generates form data based on Project
- Data structures representing the parts of a Project we want to upload (`Assets` and its children), and how to translate them into the different parts of a Form (`Files` and `Bindings`).
- all project types call the same form building logic, "assets" are configured per project.


src/commands/publish/upload_form/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+
+pub fn build_script_upload_form(project: &Project) -> Result<Form, failure::Error> {
+    let project_type = &project.project_type;
+    match project_type {
+        ProjectType::Rust => {
+            info!("Rust project detected. Publishing...");
+            let name = krate::Krate::new("./")?.name.replace("-", "_");
+            // TODO: move into build?
+            build_generated_dir()?;
+            concat_js(&name)?;
```
ashleymichal: these two steps feel like they belong in the build step, however for now i'm leaving these here with a todo, can create an issue unless they should stay here.

src/commands/publish/upload_form/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+
+            let wasm_module = WasmModule {
+                path: format!("./pkg/{}_bg.wasm", name).to_string(),
+                filename: "wasm".to_string(),
+                binding: "wasmprogram".to_string(),
+            };
```
ashleymichal: we build up a wasm module based on the template values in the `metadata_wasm.json` from the default rust template.

src/commands/publish/upload_form/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+            let bundle = Bundle::new();
+
+            let script = Script {
+                path: bundle.script_path(),
+            };
+
+            let wasm_module = WasmModule {
+                path: bundle.wasm_path(),
+                filename: bundle.get_wasm_binding(),
+                binding: bundle.get_wasm_binding(),
```
ashleymichal: likewise with webpack projects we copy the logic for writing the `metadata.json` from the `build/wranglerjs` module

src/commands/publish/upload_form/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+        ProjectType::Webpack => {
+            info!("Webpack project detected. Publishing...");
+            // FIXME(sven): shouldn't new
+            let bundle = Bundle::new();
+
+            let script = Script {
+                path: bundle.script_path(),
+            };
+
+            let wasm_module = WasmModule {
```
ashleymichal: this block should be conditional based on "bundle.has_wasm"...

src/commands/publish/upload_form/mod.rs
```diff
@@ -0,0 +1,145 @@
...
+            info!("JavaScript project detected. Publishing...");
+            let package = Package::new("./")?;
+
+            let script = Script {
+                path: package.main()?,
+            };
+
+            let assets = ProjectAssets {
+                script,
+                wasm_modules: Vec::new(),
```
ashleymichal: javascript projects don't (currently) have wasm_modules, so we just pass an empty vec here.

src/commands/publish/upload_form/mod.rs
```diff
@@ -0,0 +1,145 @@
...
+fn add_files(mut form: Form, assets: &ProjectAssets) -> Result<Form, failure::Error> {
+    for file in assets.files() {
+        form = form.file(file.name, file.path)?;
+    }
+
+    Ok(form)
+}
+
+fn add_metadata(mut form: Form, assets: &ProjectAssets) -> Result<Form, failure::Error> {
+    let metadata_json = serde_json::json!(&Metadata {
```
ashleymichal: use of the macro here may be inappropriate, idk...

src/commands/publish/upload_form/script.rs
```diff
@@ -0,0 +1,15 @@
...
+
+#[derive(Debug)]
+pub struct Script {
+    pub path: String,
+}
+
+impl ToFile for Script {
+    fn to_file(&self) -> File {
+        File {
+            name: "script".to_string(),
```
ashleymichal: this is always named `script` in the file upload

src/commands/publish/upload_form/wasm_module.rs
```diff
@@ -0,0 +1,27 @@
...
+pub struct WasmModule {
+    pub path: String,
+    pub filename: String,
+    pub binding: String,
+}
+
+impl ToBinding for WasmModule {
+    fn to_binding(&self) -> Binding {
+        let name = self.filename.clone();
+        let part = self.binding.clone();
```
ashleymichal: these are the worst... every time i think i know which is which i start to doubt. I will double check and add comments, as these names are super generic.

src/commands/publish/upload_form/mod.rs
```diff
@@ -0,0 +1,145 @@
...
+            info!("Rust project detected. Publishing...");
+            let name = krate::Krate::new("./")?.name.replace("-", "_");
+            // TODO: move into build?
+            build_generated_dir()?;
+            concat_js(&name)?;
+
+            let wasm_module = WasmModule {
+                path: format!("./pkg/{}_bg.wasm", name).to_string(),
+                filename: "wasm".to_string(),
+                binding: "wasmprogram".to_string(),
```
ashleymichal: these are backwards.

src/commands/publish/upload_form/script.rs
```diff
@@ -0,0 +1,15 @@
+use super::file::{File, ToFile};
+
+#[derive(Debug)]
+pub struct Script {
```
ashleymichal: likely a struct for this is overkill, i got carried away. nice to have, but `ProjectAssets.files()` could just as easily construct a File with just a `script_path` for now.

src/commands/publish/upload_form/wasm_module.rs
```diff
@@ -0,0 +1,29 @@
+use super::binding::{Binding, ToBinding};
+use super::file::{File, ToFile};
+
+#[derive(Debug)]
+pub struct WasmModule {
```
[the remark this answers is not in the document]
ashleymichal: the bindings module that sven put in the settings folder actually handles that pretty nicely.

ashleymichal:
> the `To<Type>` pattern in here smells of a reimplementation of https://doc.rust-lang.org/std/convert/trait.Into.html that clones for you. I don't love this pattern as it is hiding clones behind a function call, and I wonder if having a top level type to represent a resource, and a binding type that implements `From<Resource>` with options in the necessary places would be better. Not necessarily reimplementing pull/#313, but just taking those parts from it.

what's tricky about the From trait is that it consumes the first resource, but in the case of wasm modules i actually need two things: a file and a binding. perhaps the wording "ToThing" is misleading? and/or the trait is overkill and we could just duck type? 🤷‍♀

src/commands/publish/upload_form/wasm_module.rs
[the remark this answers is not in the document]
granjef3: 👍

ashleymichal:
rebased onto master. no merge conflicts.

src/commands/publish/upload_form/file.rs
```diff
@@ -0,0 +1,5 @@
+#[derive(Debug)]
+pub struct File {
```
ashleygwilliams: we should evaluate what we're doing here! we should probably use a rust builtin... but the "name" here seems like it needs sorting

src/commands/publish/upload_form/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+mod project_assets;
+mod wasm_module;
+
+use log::info;
+
+use reqwest::multipart::{Form, Part};
+use std::fs;
+use std::path::Path;
+
+use crate::commands::build::wranglerjs::Bundle;
```
ashleygwilliams: switch this to `use crate::commands::build::wranglerjs` sot hat when `Bundle` appears we know it's `wranglerjs::Bundle`

src/commands/publish/upload_form/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+        }
+        ProjectType::JavaScript => {
+            info!("JavaScript project detected. Publishing...");
+            let package = Package::new("./")?;
+
+            let script_path = package.main()?;
+
+            let assets = ProjectAssets {
+                script_path,
+                wasm_modules: Vec::new(),
```
ashleygwilliams: is there a reason we should pass an empty vector? This feels like a None!

src/commands/publish/upload_form/project_assets.rs
```diff
@@ -0,0 +1,37 @@
+use super::binding::Binding;
+use super::file::File;
+use super::wasm_module::WasmModule;
+
+#[derive(Debug)]
+pub struct ProjectAssets {
+    pub script_path: String,
+    pub wasm_modules: Vec<WasmModule>,
```
ashleygwilliams: deafult to none?

src/commands/publish/upload_form/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+            let bundle = Bundle::new();
+
+            let script_path = bundle.script_path();
+
+            let mut wasm_modules = Vec::new();
+
+            if bundle.has_wasm() {
+                let wasm_module = WasmModule {
+                    path: bundle.wasm_path(),
+                    filename: bundle.get_wasm_binding(),
```
ashleygwilliams: so let's make filename/binding the same in Webpack as RustWasm and always point to the same thing - either wasm or wasmprogram - we can file another issue for it

src/commands/publish/upload_form/mod.rs
ashleygwilliams: should be fine, might make errors weird but a lot of the values here are hardcoded so not likely to throw

src/commands/publish/upload_form/wasm_module.rs
```diff
@@ -0,0 +1,29 @@
...
+        let part = self.filename.clone();
+
+        Binding::new_wasm_module(name, part)
+    }
+}
+
+impl WasmModule {
+    pub fn file(&self) -> File {
+        File {
+            name: self.filename.clone(),
```
ashleygwilliams: we should sort if these clones are necessary

src/commands/publish/upload_form/wasm_module.rs
```diff
@@ -0,0 +1,29 @@
...
+    pub path: String,
+    pub filename: String,
+    pub binding: String,
+}
+
+// `name` corresponds to the binding used in the worker js
+// `part` corresponds to the name given to the file in the upload form
+impl WasmModule {
+    pub fn binding(&self) -> Binding {
+        let name = self.binding.clone();
```
ashleygwilliams: we should sort if these clones are necessary

src/commands/publish/upload_form/project_assets.rs
ashleymichal: idk now, i think this was a thing i said while we were CRing together and it doesn't buy much.

src/commands/publish/upload_form/mod.rs
ashleymichal: https://github.com/cloudflare/wrangler/issues/358

src/commands/publish/upload_form/wasm_module.rs
ashleymichal: tell me more 😬

src/commands/publish/upload_form/file.rs
ashleymichal: got this out, getting what we need from the paths for all the things.

src/commands/publish/upload_form/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+    let metadata = Part::text((metadata_json).to_string())
+        .file_name("metadata.json")
+        .mime_str("application/json")?;
+
+    form = form.part("metadata", metadata);
+
+    Ok(form)
+}
+
+fn filename_from_path(path: &String) -> String {
```
ashleygwilliams: this should return a `Result<String, failure::Error>` and the `unwraps` should be replaced with `?`

ashleymichal: turns out the unwraps were unwrapping `Option` types, rather than `Result` types. do you think this fn should give an error if it gets back a `None`, bc I'm inclined to feel the caller should. maybe give ProjectAssets a `new` method that checks the path for a non-empty filename?

steveklabnik: I'd probably make this return `Option<String>` in that case, and have the caller adapt the `None` to its own kind of error. I think there's good arguments both ways; I'm interested to hear what @ashleygwilliams says

ashleymichal: @steveklabnik i'm inclined to do something like this, the reasoning being that `filename_from_path` is kind of just a utility that i feel does not need to be opinionated about what a valid filename is, as it doesn't imply an application of the filename, whereas i could check either before constructing the ProjectAssets struct, or inside the constructor.

PR #333, opened 2019-07-19
feat kv config

ashleymichal:
This PR starts us off on grabbing the correct info from `wrangler.toml` for KVNamespace support. Two breaking changes to note:

- as of this feature, we will no longer provision namespaces on behalf of the user. We will revisit that feature once we implement KV subcommands.
- KV Namespace now takes both an `id` and a `binding` name in the toml per design.
- Wrangler includes KV Namespace Bindings in upload requests

~Next PR to include logic for including KV Bindings in upload requests 🎉~ Pushed up KV Bindings upload here...

API Error handling to come.

### DIFF
--- a/src/settings/project/kv_namespace.rs
+++ b/src/settings/project/kv_namespace.rs
@@ -3,18 +3,18 @@ use std::fmt;
 use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
-pub struct KVNamespace {
+pub struct KvNamespace {
     pub id: String,
     pub binding: String,
 }
 
-impl fmt::Display for KVNamespace {
+impl fmt::Display for KvNamespace {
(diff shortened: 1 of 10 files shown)


src/settings/project/kv_namespace.rs
```diff
@@ -0,0 +1,21 @@
+use std::fmt;
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct KVNamespace {
```
steveklabnik: This should be `KvNamespace` https://rust-lang-nursery.github.io/api-guidelines/naming.html

> In UpperCamelCase, acronyms and contractions of compound words count as one word: use Uuid rather than UUID, Usize rather than USize or Stdin rather than StdIn. In snake_case, acronyms and contractions are lower-cased: is_xid_start.

PR #334, opened 2019-07-19
feat: Support KV Namespace Configuration

### LINKED ISSUE #273: Refactor Bindings to support all binding types
Wrangler should be able to support all binding types not just wasm and kv. In order to support future bindings and make bindings generation simpler, we should try to treat all bindings the same (kv, wasm, secrets, etc.).

ashleymichal:
Builds on #329 (merge that first)

Closes #273 #92

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,9 +25,9 @@ cargo install wrangler
 
 ## 🎙️ Commands
 
-  - ### 👯 `generate` 
+  - ### 👯 `generate`
     Scaffold a project, including boilerplate for a Rust library and a Cloudflare Worker.
-    You can pass a name and template to this command optionally. 
+    You can pass a name and template to this command optionally.
 
     ```
     wrangler generate <name> <template> --type=["webpack", "javascript", "rust"]
@@ -61,7 +61,7 @@ cargo install wrangler
     By default, `publish` will make your worker available at `<project-name>.<subdomain>.workers.dev`.
     To disable publishing to your workers.dev subdomain, set `private = true` in your `wrangler.toml`.
     This setting prevents the `publish` command from making your worker publicly available. To
-    explicitly enable deployment to `<project-name>.<subdomain>.workers.dev`, you can set `private = false`. 
+    explicitly enable deployment to `<project-name>.<subdomain>.workers.dev`, you can set `private = false`.
 
     To use this command, you'll need to have the following keys in your `wrangler.toml`:
 
@@ -143,6 +143,17 @@ There are two types of configuration that `wrangler` uses: global user and per p
         This key is optional if you are using a workers.dev subdomain and is only required for `publish --release`.
     - `webpack_config`: This is the path to the webpack configuration file for your worker. This is optional and
         defaults to `webpack.config.js`
+    - `[[kv-namespaces]]`: These specify any [Workers KV](TODO) namespaces you want to access from
+        inside your Worker. Each namespace you include should have an entry in your wrangler.toml that includes:
+
+        - `binding`: the name you want to bind to in your script
+        - `id`: the namespace_id assigned to your kv namespace [upon creation](https://workers.cloudflare.com/docs/reference/storage/writing-data/).
+            e.g. (per namespace):
+        ``` toml
+        [[kv-namespaces]]
+        binding = "foo"
+        id = "0f2ac74b498b48028cb68387c421e279"
+        ```
 
 ## ⚓ Installation
 
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ There are two types of configuration that `wrangler` uses: global user and per p
         This key is optional if you are using a workers.dev subdomain and is only required for `publish --release`.
     - `webpack_config`: This is the path to the webpack configuration file for your worker. This is optional and
         defaults to `webpack.config.js`
-    - `[[kv-namespaces]]`: These specify any [Workers KV](TODO) namespaces you want to access from
+    - `[[kv-namespaces]]`: These specify any [Workers KV](https://workers.cloudflare.com/docs/reference/storage/) namespaces you want to access from
         inside your Worker. Each namespace you include should have an entry in your wrangler.toml that includes:
 
         - `binding`: the name you want to bind to in your script
--- a/README.md
+++ b/README.md
@@ -147,13 +147,14 @@ There are two types of configuration that `wrangler` uses: global user and per p
         inside your Worker. Each namespace you include should have an entry in your wrangler.toml that includes:
 
         - `binding`: the name you want to bind to in your script
-        - `id`: the namespace_id assigned to your kv namespace [upon creation](https://workers.cloudflare.com/docs/reference/storage/writing-data/).
+        - `id`: the namespace_id assigned to your kv namespace upon creation.
             e.g. (per namespace):
         ``` toml
         [[kv-namespaces]]
         binding = "foo"
         id = "0f2ac74b498b48028cb68387c421e279"
         ```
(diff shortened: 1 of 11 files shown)


ashleymichal:
rebased onto master. no merge conflicts

src/settings/binding.rs
```diff
@@ -5,10 +5,16 @@ use serde::Serialize;
 pub enum Binding {
     #[allow(non_camel_case_types)]
     wasm_module { name: String, part: String },
+    #[allow(non_camel_case_types)]
```
cwndrws: You could use camel case serde renames instead of silencing the warning.

src/commands/publish/upload_form/project_assets.rs
```diff
@@ -0,0 +1,43 @@
...
+            let binding = wm.binding();
+            bindings.push(binding);
+        }
+        for kv in &self.kv_namespaces {
+            let binding = kv.binding();
+            bindings.push(binding);
+        }
+
+        bindings
+    }
```
steveklabnik: I really want to write this as

```rust
pub fn bindings(&self) -> Vec<Binding> {
    self.wasm_modules
        .iter()
        .chain(self.kv_namespaces.iter())
        .map(|e| e.binding())
        .collect()    
}
```

but, that doesn't work since they're different types.

You *could* write

```rust
pub fn bindings(&self) -> Vec<Binding> {
    self.wasm_modules
        .iter()
        .map(|m| m.binding())
        .chain(self.kv_namespaces.iter().map(|n| n.binding())
        .collect()    
}
```

but I think this is stretching the limits of Too Many Iterators :)

Another option is

```rust
    let mut bindings: Vec<_> = wasm_modules.iter().map(|m| m.binding()).collect();
    bindings.extend(kv_namespaces.iter().map(|n| n.binding()));
    bindings
```

... but after considering all of this, i think the original code is probably the most clear in this case. I guess.

src/commands/publish/upload_form/wasm_module.rs
```diff
@@ -0,0 +1,29 @@
...
+impl WasmModule {
+    pub fn binding(&self) -> Binding {
+        let name = self.binding.clone();
+        let part = self.filename.clone();
+
+        Binding::new_wasm_module(name, part)
+    }
+}
+
+impl WasmModule {
```
steveklabnik: any reason these are in two separate `impl` blocks?

src/settings/project/kv_namespace.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub binding: String,
+}
+
+impl fmt::Display for KvNamespace {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "id: {}, binding: {}", self.id, self.binding)
+    }
+}
+
+impl std::cmp::PartialEq for KvNamespace {
```
steveklabnik: i *believe* this is the same as if you derived it, is that true? if it's not, might be worth a comment explaining why this is impl'd manually

src/settings/project/mod.rs
```diff
@@ -96,14 +62,23 @@ impl Project {
...
+        config_path.push("./wrangler.toml");
+
+        get_project_config(config_path)
+    }
+
+    pub fn kv_namespaces(&self) -> Vec<KvNamespace> {
+        match &self.kv_namespaces {
+            Some(kv) => kv.clone(),
+            None => Vec::new(),
+        }
```
steveklabnik: this can be written

```rust
self.kv_namespaces.cloned().unwrap_or_else(Vec::new)
```

src/settings/project/mod.rs
```diff
@@ -96,14 +62,23 @@ impl Project {
...
     }
 }
 
-pub fn get_project_config() -> Result<Project, failure::Error> {
+fn get_project_config(config_path: PathBuf) -> Result<Project, failure::Error> {
```
steveklabnik: usually, functions take a `&Path` instead of a `PathBuf`, and it looks like `config_path` was previously a `&Path`; this should probably change to that

src/commands/publish/upload_form/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+use wasm_module::WasmModule;
+
+use super::{krate, Package};
+
+pub fn build_script_upload_form(project: &Project) -> Result<Form, failure::Error> {
+    let project_type = &project.project_type;
+    let kv_namespaces = project.kv_namespaces();
+    match project_type {
+        ProjectType::Rust => {
+            info!("Rust project detected. Publishing...");
```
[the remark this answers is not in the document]
ashleymichal: for sure; i think at some point i'd love the whole thing to be pushed up, but for right now i'm kinda down with this (slightly more verbose) approach, as it's not terribly dense and not much needs to be abstracted away.

src/commands/publish/upload_form/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    Ok(())
+}
+
+fn concat_js(name: &str) -> Result<(), failure::Error> {
+    let bindgen_js_path = format!("./pkg/{}.js", name);
+    let bindgen_js: String = fs::read_to_string(bindgen_js_path)?.parse()?;
+
+    let worker_js: String = fs::read_to_string("./worker/worker.js")?.parse()?;
```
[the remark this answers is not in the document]
ashleymichal: this function is copy-pasted from the original file, but in essence these hard-coded values are based on templates. Not ideal, but changing this is currently out of scope.

src/commands/publish/upload_form/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+    let js = format!("{} {}", bindgen_js, worker_js);
+
+    fs::write("./worker/generated/script.js", js.as_bytes())?;
```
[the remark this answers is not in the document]
ashleymichal: same

src/commands/publish/upload_form/project_assets.rs
ashleymichal: 😂

src/commands/publish/upload_form/wasm_module.rs
ashleymichal: mmmmm nope.

src/settings/project/kv_namespace.rs
ashleymichal: mmm good point. i'll have a look.

src/settings/project/mod.rs
ashleymichal: i hope you meant `.clone()` bc `.cloned()` gave me many complaints.

steveklabnik: no, i meant cloned: https://doc.rust-lang.org/std/option/enum.Option.html#method.cloned

steveklabnik: image

cloned goes from `Option<&T>` to `Option<T>`. But you already have an `Option<T>`. `clone` is right here, and I'm wrong.

Thanks, compiler!

signalnerve:
@ashleymichal can you add some details about what this PR is, so that I can document it in the CHANGELOG when it's merged?

README.md
```diff
@@ -143,6 +143,17 @@ There are two types of configuration that `wrangler` uses: global user and per p
         This key is optional if you are using a workers.dev subdomain and is only required for `publish --release`.
     - `webpack_config`: This is the path to the webpack configuration file for your worker. This is optional and
         defaults to `webpack.config.js`
+    - `[[kv-namespaces]]`: These specify any [Workers KV](https://workers.cloudflare.com/docs/reference/storage/) namespaces you want to access from
```
ashleygwilliams: we should explicitly call out that users are expected to make their namespace in the UI

ashleymichal: added commit for this.

PR #336, opened 2019-07-21
Fixes #278 - Check for rust before installing wasm-pack

### LINKED ISSUE #278: wranglerjs needs to properly forward wasm-pack errors to the user
If you try to build a rust+wasm project, after installing wrangler through npm, you get "OS error 2" instead of "hey, you need to install rust"

xprazak2:
I tried to pull in [mocktopus](https://github.com/CodeSandwich/Mocktopus), but it is only available for nightly rust, so I did not go through with it. Let me know about your preference for mocking in tests.

### DIFF
--- a/src/commands/build/wranglerjs/mod.rs
+++ b/src/commands/build/wranglerjs/mod.rs
@@ -64,7 +64,7 @@ pub fn run_build(project: &Project) -> Result<(), failure::Error> {
 
 //setup a build to run wranglerjs, return the command, the ipc temp file, and the bundle
 fn setup_build(project: &Project) -> Result<(Command, PathBuf, Bundle), failure::Error> {
-    for tool in &["node", "npm"] {
+    for tool in &["node", "npm", "rustc"] {
         env_dep_installed(tool)?;
     }
 


src/commands/build/wranglerjs/mod.rs
```diff
@@ -62,6 +62,10 @@ pub fn run_build(project: &Project) -> Result<(), failure::Error> {
     }
 }
 
+fn rust_present() -> bool {
+    which::which("rustc").is_ok()
+}
+
 //setup a build to run wranglerjs, return the command, the ipc temp file, and the bundle
 fn setup_build(project: &Project) -> Result<(Command, PathBuf, Bundle), failure::Error> {
     for tool in &["node", "npm"] {
```
xtuc: I believe you can add `rustc` here and it will check for its exisiance in the env

xprazak2: You are right, updated.

granjef3:
Hey there, thanks for the PR! `wasm-pack` actually checks for the presence of rust and we want to use that behavior instead of duplicating it in wrangler. We appreciate the work you've done here, but we are going to close this PR in favor of forwarding the `wasm-pack` error. Feel free to open a new WIP PR if you wish to implement that solution!

Issue #337, opened 2019-07-22
Dockerize wrangler


mrpatrick:
### Suggestion
* It would be great to have a dockerized version of this program that is self-contained
* include an option to for build-required files (i.e. cache the template requirements in the image so they don't have to be downloaded everytime you call `build`
* include node version option

gabbifish:
Hi @mrpatrick! Thanks for submitting a feature request + suggested PR :).

Before we consider adding a new feature to Wrangler, we need to think about how it fits into the Wrangler user experience. A couple of questions with respect to your feature:
1. How does this dockerized version of Wrangler improve Wrangler setup and usage? What pain points does it address?
2. What is the intended workflow for using this dockerized Wrangler binary?
3. How would this dockerized distribution of wrangler be distributed? Would folks need to always build the container themself? Would this be an image folks would use from the docker registry? How will the dockerized wrangler binary be maintained and updated with new point releases?

Again, thanks so much for reaching out to us, and we're looking forward to hear more of your decision-making.

mrpatrick:
Hi @gabbifish - thanks for considering.  See inline below

> 1. How does this dockerized version of Wrangler improve Wrangler setup and usage? What pain points does it address?

Doesn't rely on npm/node (or knowledge on using node/npm), os libraries, software updates, etc.  You just need docker - that's it.  This makes "installing" easy (i.e. I can use this program even though I don't have node/npm on my laptop).  It always makes updates easier and doesn't rely on 3rd party libraries / software other than docker (i.e. no conflicts / issues with updates).  it's fully self contained.  Docker is well established platform and most CLI's and services have an official docker version.  This would help wrangler become more widely accepted and provide greater availability / visibility. 

> 2. What is the intended workflow for using this dockerized Wrangler binary?

Suggested workflow is documented here: https://github.com/cloudflare/wrangler/pull/338

This is of course how I'm using it and others may have better suggestions / improvements to this.

> How would this dockerized distribution of wrangler be distributed? Would folks need to always build the container themself? Would this be an image folks would use from the docker registry? 

I would suggest providing both methods (as keeping with docker best practices and standards).  If you take a look at my PR, it's just adding 2 files: Dockerfile, .dockerignore to the repo.  That allows other to build the image as they see fit as well as provide an docker image via docker registry like I'm currently doing via our company repo:

https://hub.docker.com/r/avatarnewyork/wrangler

[...]

PR #338, opened 2019-07-22
Dockerize wrangler

mrpatrick:
This commit "dockerizes" wrangler #337 .  Included are build args that include default values below.  You can pull a working example here: `docker pull avatarnewyork/wrangler`

### Build image
`docker build -t wrangler --build-arg version=12.6.0-alpine INCLUDE_WORKER_TEMPLATE=https://github.com/cloudflare/worker-template .`

### Configuration
create a directory if it doesn't already exist where you can mount the wrangler global config (if you plan on publishing to your own domain) assuming `/Users/me` is your home dir

``` bash
mkdir -p /Users/me/.wrangler/config
```

#### generate
Use the current directory to create a project called myproject (see generate command below for all options)

`docker run --rm -ti -v /Users/me/.wrangler/config:/home/node/.wrangler/config -v $(pwd):/home/node/wrangler wrangler generate myproject`

#### build / preview / publish
move into your project directory and run any of the other commands (see commands below for all options)

* `docker run --rm -ti -v /Users/me/.wrangler/config:/home/node/.wrangler/config -v $(pwd):/home/node/wrangler wrangler build`
* `docker run --rm -ti -v /Users/me/.wrangler/config:/home/node/.wrangler/config -v $(pwd):/home/node/wrangler wrangler preview`
* `docker run --rm -ti -v /Users/me/.wrangler/config:/home/node/.wrangler/config -v $(pwd):/home/node/wrangler wrangler publish`


Dockerfile
```diff
@@ -0,0 +1,19 @@
+ARG version=12.6.0-alpine
+FROM node:$version
+
+ARG INCLUDE_WORKER_TEMPLATE=https://github.com/cloudflare/worker-template
```
Electroid: This is nice, but I wonder if this should be included in the official image. What's the use case for running the default template in a container?

mrpatrick: @Electroid Originally, I left it out, however after actually using the built image, it became impractical to use as the build/preview/publish commands take a long time to run (since it will have to download / install the components every execution.  Including the template here, "caches" those files so the commands are fast (I also thought it was inline with the project in that it assumes this template unless you specify another one as part of the config).  Just a note, I'm a novice when it comes to Node 
 so there is probably a much better / more elegant way to handle this - def open to suggestions.

Exadra37:
Why was this close?

This would be a great addition to have.

Using wrangler means I have to install all its dependencies in my dev machine and solve conflicts. It's an headache I need to go hover each time I need to use wrangler, that could be avoided by package all together in a working solution, aka conflict free with whatever you already have installed in your machine. Docker is a very good way of doing it so.

EverlastingBugstopper:
Hey @Exadra37 - this was closed because there wasn't any activity for a while. I'll reopen here but we have no immediate plans to manage and distribute a Docker image.

PR #353, opened 2019-07-23
Don't break `wrangler preview` when kv namespaces present

ashleymichal:
the preview service as we use it today does not ignore kv namespace bindings, but throws an error when sent a script with those bindings. we don't want to take away a user's ability to use preview, for one, and we also want to give a helpful warning when they do.

### DIFF
--- a/src/commands/publish/upload_form/mod.rs
+++ b/src/commands/publish/upload_form/mod.rs
@@ -19,16 +19,19 @@ use wasm_module::WasmModule;
 
 use super::{krate, Package};
 
-pub fn build_script_upload_form(
+pub fn build_script_upload_form(project: &Project) -> Result<Form, failure::Error> {
+    build_script_upload_form_with_kv(project, project.kv_namespaces())
+}
+
+pub fn build_script_upload_form_no_kv(project: &Project) -> Result<Form, failure::Error> {
+    build_script_upload_form_with_kv(project, Vec::new())
+}
+
+fn build_script_upload_form_with_kv(
     project: &Project,
-    include_kv: bool,
+    kv_namespaces: Vec<kv_namespace::KvNamespace>,
 ) -> Result<Form, failure::Error> {
     let project_type = &project.project_type;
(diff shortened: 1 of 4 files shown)


src/commands/publish/upload_form/mod.rs
```diff
@@ -19,9 +19,17 @@ use wasm_module::WasmModule;
 
 use super::{krate, Package};
 
-pub fn build_script_upload_form(project: &Project) -> Result<Form, failure::Error> {
+pub fn build_script_upload_form(
+    project: &Project,
+    include_kv: bool,
+) -> Result<Form, failure::Error> {
```
steveklabnik: so, boolean parameters are generally considered to be a bad style. I actually really *like* what you did above, with

```rust
    let include_kv = false;
```

rather than just passing `false` and `true`. But we can do even better!

```rust
// note this isn't pub!
fn build_script_upload_form_with_namespaces(
    project: &Project,
    kv_namespaces: Vec<NamespaceTypeOrWhatever>,
) -> Result<Form, failure::Error> {
   // previous body of build_script_upload_form goes here
   //
   // except you don't need to set up the kv_namespaces variable
}


pub fn build_script_upload_form(
    project: &Project,
) -> Result<Form, failure::Error> {
    build_script_upload_form_with_namespaces(project, Vec::new())
}

pub fn build_script_upload_form_with_kv(
    project: &Project,
    include_kv: bool,
) -> Result<Form, failure::Error> {
     build_script_upload_form_with_namespaces(project, project.kv_namespaces())
}
```

I do not know how @ashleygwilliams feels about this kind of refactoring, but I generally prefer it. That being said, boolean parameters are not *the worst*, so if she prefers this code as-is, I don't think it's the end of the world.

ashleymichal: the boolean parameter did indeed feel Gross. I like this better, ty.

ashleymichal: implemented this, except called out the "weird" case which is building without kv.

PR #360, opened 2019-07-26
[WIP] - Use `cloudflare-rs` for Cloudflare API calls

ashleymichal:
# Proposal

The Argo Tunnel team has put together an open source Rust library for interacting with Cloudflare's v4 API endpoints. It handles wrapping reqwest, constructing http requests to the various endpoints' specifications, and parsing / extracting both API success and API failure responses. Currently there is only partial support for WorkersKV (mostly untested), but extending the library is relatively straight forward in terms of pattern matching, though the Rust can be daunting to a beginner.

This PR illustrates one possible approach to wrapping and using this library in Wrangler.

To give a sense of the work involved in adopting this as our API client, I've linked the various artifacts of my process so far below:

## Contributing to `cloudflare-rs`
A proposal PR I made to the tunnel team about how to structure service modules to better represent usage: 
https://github.com/cloudflare/cloudflare-rs/pull/2
The meat of the proposal is in the Readme and the directory structure (specifically the zone module which has been broken out into a root with the domain data structures and files containing the individual endpoints)

## Proposed WorkersKV module (`cloudflare-rs` module)
An active PR applying the above module structure to the existing WorkersKV module, as well as fixing a bug in the endpoint: https://github.com/cloudflare/cloudflare-rs/pull/7

**Note**: it is expected that this PR does not pass CI, due to the outstanding PR against the cloudflare-rs repo that updates the CreateNamespace endpoint.


src/http/kv/create_namespace.rs
```diff
@@ -0,0 +1,32 @@
...
+    let client = new_api_client(user);
+
+    let res = client
+        .request(&workerskv::CreateNamespace {
+            account_identifier: &project.account_id,
+            params: workerskv::CreateNamespaceParams { title },
+        })
+        .unwrap_or_else(|failure| match failure {
+            APIFailure::Error(_, errors) => {
+                panic!("{}", api_errors_to_message(errors));;
```
ashleymichal: panics should be replaced by our own errors, these are for illustration purposes only.

src/http/kv/create_namespace.rs
```diff
@@ -0,0 +1,32 @@
...
+            }
+            APIFailure::Invalid(error) => {
+                panic!("{}", error);
```
ashleymichal: these panics are for illustrative purposes, i'd prefer to return our own errors.

src/http/kv/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+        }
+    }
+}
+
+// TODO: Map these to Wrangler-relevant messages? or an Enum?
+fn api_errors_to_message(errors: Vec<APIError>) -> String {
+    let mut messages = Vec::new();
+    for error in errors {
+        match error.code {
+            // https://api.cloudflare.com/#workers-kv-namespace-errors
```
ashleymichal: these error messages are literally just mapped from the linked docs; technically these are exactly the same text as we would pull out of the `message` part of the error response, but here is where we could translate them to wrangler-specific messages, including actionable feedback to the user

src/http/mod.rs
```diff
@@ -1,3 +1,5 @@
+pub mod kv;
```
ashleymichal: it felt appropriate to namespace the http mods based on the services we use, i.e. workers and kv, so you might call, in this case, `kv::create_namespace("title")`.

src/http/kv/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+impl From<GlobalUser> for Credentials {
+    fn from(user: GlobalUser) -> Credentials {
+        Credentials::User {
+            key: user.api_key,
+            email: user.email,
+        }
+    }
+}
+
+fn new_api_client(user: GlobalUser) -> HTTPAPIClient {
```
granjef3: nit incoming: We can extend the HTTPAPIClient type and implement From<GlobalUser> for it. `new_api_client` takes ownership of `user` so the semantics are the same.

I would suggest the pattern of `use cloudflare::HTTPAPIClient as CFAPIClient` if we did this, and then `CFAPIClient::from(user)` which _feels_ kinda nice

src/commands/publish/mod.rs
```diff
@@ -166,6 +126,21 @@ fn validate_project(project: &Project, release: bool) -> Result<(), failure::Err
         missing_fields.push("name")
     };
 
+    match &project.kv_namespaces {
```
steveklabnik: the form 

```
match foo {
    Some(bar) => { stuff }
    None => {}
```

can be written more clearly as

```
foo.map(|bar| { stuff })
```

src/http/kv/create_namespace.rs
```diff
@@ -0,0 +1,32 @@
+use cloudflare::apiclient::APIClient;
```
steveklabnik: I guess I should file a bug upstream, but this should be `ApiClient`

src/commands/publish/mod.rs
ashleymichal: this wasn't intended to be part of this pr, it just crept in when the target branch got rebased.

src/http/kv/create_namespace.rs
sssilver: We're happy to rename the types

src/http/kv/create_namespace.rs
```diff
@@ -0,0 +1,32 @@
...
+use crate::settings::project::Project;
+
+use super::{api_errors_to_message, new_api_client, KvNamespace};
+
+pub fn create_namespace(
+    user: GlobalUser,
+    project: &Project,
+    title: String,
+) -> Result<KvNamespace, failure::Error> {
+    let client = new_api_client(user);
```
sssilver: It feels to me like wrapping `create_namespace` with an identically named function that has different arguments more pertinent to your abstractions ("project", "user") to be unnecessary boilerplate. There's very little substance in the implementation of this function. So I wouldn't do that.

But if I were doing that, I'd probably pass the `client` as an argument, to help make this more testable and less rigid.

And that would take away even more substance, making me question my choice of wrapping it even more.

src/http/kv/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+use crate::settings::global_user::GlobalUser;
+
+impl From<GlobalUser> for Credentials {
+    fn from(user: GlobalUser) -> Credentials {
+        Credentials::User {
+            key: user.api_key,
+            email: user.email,
+        }
+    }
+}
```
sssilver: ❤️

src/http/kv/mod.rs
sssilver: Not sure how I feel about this. `From`'s semantics imply conversion. Although admittedly isomorphism is not a part of the definition, it doesn't feel right to say that we're _converting_ a user into an API client. A constructor to me feels more appropriate here.

src/http/kv/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+
+fn new_api_client(user: GlobalUser) -> HTTPAPIClient {
+    let credentials = Credentials::from(user);
+
+    HTTPAPIClient::new(credentials)
+}
+
+#[derive(Debug)]
+pub struct KvNamespace {
```
sssilver: The API client has no type for this?

sssilver: Ah you're doing the extra layer of abstraction here as well. IMO unnecessary.

ashleymichal:
closing in favor of #392

PR #370, opened 2019-07-30
nit: Make binding name capitalized in readme

xortive:
this is a style thing, but since bindings are a global I think they should be capitalized (to match other snippets that use them as well)


ashleymichal:
i mean, sure? this is likely a larger style question, but the only constraints on bindings is that they be acceptable js symbols

xortive:
I was going for consistency with the kv docs, unsure about wider JS ecosystem

https://workers.cloudflare.com/docs/reference/storage/reading-data/

PR #382, opened 2019-08-02
Migrate straggler emojis to terminal::emoji

EverlastingBugstopper:
This PR changes a few remaining instances where warngler was using hard-coded emojis for messages rather than `terminal::emoji`. There are two instances where this changes ⛔ to ⚠️, but the rest is simply a change in implementation.

### DIFF
--- a/src/commands/publish/route.rs
+++ b/src/commands/publish/route.rs
@@ -1,6 +1,7 @@
 use crate::http;
 use crate::settings::global_user::GlobalUser;
 use crate::settings::project::Project;
+use crate::terminal::emoji;
 use reqwest::header::CONTENT_TYPE;
 use serde::{Deserialize, Serialize};
 
@@ -27,10 +28,10 @@ impl Route {
         {
             failure::bail!("You must provide a zone_id in your wrangler.toml before publishing!");
         }
-
+        let msg_config_error = format!("{} Your project config has an error, check your `wrangler.toml`: `route` must be provided.", emoji::WARN);
         Ok(Route {
             script: Some(project.name.to_string()),
-            pattern: project.route.clone().expect("⚠️ Your project config has an error, check your `wrangler.toml`: `route` must be provided.").to_string(),
+            pattern: project.route.clone().expect(&msg_config_error),
         })
     }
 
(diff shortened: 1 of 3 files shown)


src/commands/publish/route.rs
```diff
@@ -70,7 +71,8 @@ fn get_routes(user: &GlobalUser, project: &Project) -> Result<Vec<Route>, failur
 
     if !res.status().is_success() {
         let msg = format!(
-            "⛔ There was an error fetching your project's routes.\n Status Code: {}\n Msg: {}",
+            "{} There was an error fetching your project's routes.\n Status Code: {}\n Msg: {}",
+            emoji::WARN,
```
gabbifish: I think we're losing the ⛔ emoji here; we should probably add a new emoji to https://github.com/cloudflare/wrangler/blob/master/src/terminal/emoji.rs so we still have it on hand. Perhaps emoji::ERROR?

src/commands/publish/route.rs
```diff
@@ -97,7 +99,8 @@ fn create(user: &GlobalUser, project: &Project, route: &Route) -> Result<(), fai
 
     if !res.status().is_success() {
         let msg = format!(
-            "⛔ There was an error creating your route.\n Status Code: {}\n Msg: {}",
+            "{} There was an error creating your route.\n Status Code: {}\n Msg: {}",
```
gabbifish: same as above; we probably want to use a new emoji::ERROR here to preserve this emoji.

src/commands/subdomain.rs
```diff
@@ -60,9 +60,10 @@ fn subdomain_addr(account_id: &str) -> String {
 
 pub fn subdomain(name: &str, user: &GlobalUser, project: &Project) -> Result<(), failure::Error> {
     if project.account_id.is_empty() {
-        failure::bail!(
-            "⛔ You must provide an account_id in your wrangler.toml before creating a subdomain!"
-        )
+        failure::bail!(format!(
```
gabbifish: Same emoji:ERROR comment as above.

src/commands/publish/route.rs
EverlastingBugstopper: We decided to change that emoji a while back but missed these instances: https://github.com/cloudflare/wrangler/pull/327/files#diff-2e8c637bda89431d960c0efd861afbf9L28

EverlastingBugstopper: Losing the ⛔is intentional to make our warning error messages consistent across the project

gabbifish: Gotcha! LGTM then :)

Issue #385, opened 2019-08-02
Add support for multiple environments


EverlastingBugstopper:
EDIT: A lot of the stuff in this top level is outdated - you can browse through the [Environments milestone](https://github.com/cloudflare/wrangler/milestone/12) for a clearer picture :)

This issue was born out of RFC #220  👶 

# 👉 Current Situation

`wrangler publish --release` only supports a single configuration. It would be very useful to compose a `wrangler.toml` that allows a project to be deloyed across multiple zones/accounts.

# 🙏 Proposed Usage

I think it would be good to keep this change backwards compatible with existing `wrangler.toml` configurations, and also allow a new configuration type. The following would all be valid `wrangler.toml` files.

## 👷 Current Configuration Style

```
name = "worker"
type = "webpack"
zone_id = "12345abcd"
private = false
account_id = "12345abcd"
route = "https://example.com"
``` 

This configuration would allow you to `wrangler publish` for zoneless and `wrangler publish --release` for zoned. This behavior should remain unchanged with this update.

## 👷 Add a staging environment to an existing configuration

```
name = "worker"
type = "webpack"
zone_id = "12345abcd"
private = false
account_id = "12345abcd"
route = "https://example.com"

[env.staging]
name = "staging-worker"
type = "webpack"
zone_id = "12345abcd"
private = false
account_id = "12345abcd"
route = "https://staging.example.com"
``` 

This configuration would allow you to deploy zoneless with `wrangler publish`, `wrangler publish --release` would deploy the configuration at the top, and `wrangler publish --release --env staging` would deploy the worker to their staging environment.

## 👷Configuration that requires explicitly specifying environment 

```
[env.prod]
name = "worker"
type = "webpack"
zone_id = "12345abcd"
private = false
account_id = "12345abcd"
route = "https://example.com"

[env.staging]
name = "staging-worker"
type = "webpack"
zone_id = "12345abcd"
private = false
account_id = "12345abcd"
route = "https://staging.example.com"
```

This configuration would allow you to publish zoneless with `wrangler publish`, and `wrangler publish --release` would ask you to specify an environment to deploy to. You could run `wrangler publish --release --env staging`, `wrangler publish --release --env prod`, and `wrangler publsh --release --env staging --env prod`.

# 🤢 Problematic Configurations

## ⚠️  Same name, different routes

```
[env.prod]
name = "worker"
type = "webpack"
zone_id = "12345abcd"
private = false
account_id = "12345abcd"
[...]

PR #386, opened 2019-08-02
Environments

### LINKED ISSUE #357: Unclear messaging of wrangler publish
my `wrangler.toml`looked like:

```
name = "request"
type = "webpack"
zone_id = "ef47aa23.."
private = true
account_id = "95e06.."
route = "https://vsnginx.workers-tooling.cf"
```

 from running `wrangler publish` I got:
```
⬇️   Installing wranglerjs...
✨   Built successfully, built project size is 1 KiB. ✨
✨   Successfully published your script. ✨
✨   Success! Your worker was successfully published. ✨
```
my intention of course was to run `wrangler publish --release` which I have done several times just missed the flag 🤦‍♀ , but we can work on better messaging. Here are some ideas:
- change output of wrangle publish to say what it's actually publishing e.g. ` Your script named request was successfully published on blah.v.workers.dev and https://nginx.my.com/`
- when publishing without the `--release` flag send some info. e.g. `To publish on a zone other than workers.dev, you can include the --release flag along with a zone_id and route in your wrangler.toml`
- if zone tag and route are included with private set to true maybe assume `--release` ? that's tricky though because one could accidentally push a route

EverlastingBugstopper:
This PR will add the feature described in #385. 

*TL;DR:* Wrangler users want to be able to push their same workers project to multiple places, perhaps they have a production, staging, and dev worker. This PR allows them to do so with the introduction of the `--env` flag and some changes to `wrangler.toml`. Most notably, `private` has been deprecated, `workers_dot_dev` has been added as a boolean field, and you can specify new environments with `[env.foo]`. To read more, please see [this document](https://github.com/cloudflare/wrangler/blob/avery/environments/docs/environments.md).

# Contains

- Adds tests for building with multiple environments
- Separate deprecated tomls in fixture directory
- parse multiple environments from `wrangler.toml`
- deprecate `--release`
- pass desired environment via `--env` or `-e` for `wrangler publish` from command line
- add `workers_dot_dev` to `wrangler.toml` 
- define default behaviors at top of `wrangler.toml` and inherit throughout each environment
- renames `Project` to `Target`, the reasoning behind this is that `Project` seems to describe an entire codebase, when what we're really doing is passing around an object that specifies where the worker is deployed to -- seems like `Target` is a more suitable name.
- documentation for environments feature

## Completed issues
Fixes #357, #385, #388


src/settings/project/mod.rs
```diff
@@ -13,43 +13,293 @@ use std::path::{Path, PathBuf};
...
+        let wdd_failure = format!(
+            "{} Your environment should only include `workers_dot_dev` or `route`",
+            emoji::WARN
+        );
+
+        // TODO: deprecate --release, remove warnings and parsing
+        // switch wrangler publish behavior to act the same at top level
+        // and environments
+        // brace yourself, this is hairy
+        let workers_dot_dev = match environment {
```
gabbifish: In general, I think this hairy section could be greatly simplified by using `if let` syntax instead of many nested matches: https://doc.rust-lang.org/rust-by-example/flow_control/if_let.html

PR #391, opened 2019-08-03
refactor: move emoji formatting into concrete fns

ashleymichal:
doing this for some more flexibility in new message function formats. merging into master but this is relevant to KV commands work #392


EverlastingBugstopper:
This is great! Do you eventually want to deprecate external usage of `message` altogether? Perhaps we should open an issue for updating instances of `message` elsewhere in the codebase to use these new functions.

ashleymichal:
@EverlastingBugstopper there shouldn't be any external use of the `message` fn, as it is not public. everywhere you see `message` outside of this file it is used as a namespace. The purpose of this module is to centralize and standardize how we present information to the user in the terminal. the purpose of the refactor is to just move more of the formatting concerns into the module functions themselves rather than assume that every type of message is surrounded by a specific emoji.

PR #392, opened 2019-08-03
feat create namespace

### LINKED ISSUE #340: Add KV Command: Create Namespace
As a Workers KV user, I should be able to add a namespace to my KV store from Wrangler by typing

``` sh
$ wrangler kv add myNamespace
```

ashleymichal:
Closes #340 . Includes commit from #391 (merge first). 

This is dependent on an update to the cloudflare-rs library per this issue: https://github.com/cloudflare/cloudflare-rs/issues/10.

Interacting with this code looks a little something like this:
On successful creation:
``` sh
$ wrangler kv create "my namespace"
🌀  Creating namespace with title "my namespace"
✨  Success: WorkersKVNamespace {
    id: "f7b02e7fc70443149ac906dd81ec1791",
    title: "my namespace",
}
```

On API error (in this case duplicate kv namespace):
``` sh
$ wrangler kv create "my namespace"
🌀  Creating namespace with title "my namespace"
⚠️  Error 10014: a namespace with this account ID and title already exists
🕵️‍♂️  Run `wrangler kv list` to see your existing namespaces with IDs

```


Cargo.toml
```diff
@@ -13,6 +13,7 @@ build = "build.rs"
 atty = "0.2.11"
 binary-install = "0.0.3-alpha"
 clap = "2.32.0"
+cloudflare = { git = "https://github.com/cloudflare/cloudflare-rs.git" }
```
steveklabnik: is there a reason for this? this will block publishing of `wrangler` to crates.io. https://crates.io/crates/cloudflare does appear to exist.

src/commands/kv/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+        10010 | 10011 | 10012 | 10013 | 10014 | 10018 => {
+            "Run `wrangler kv list` to see your existing namespaces with IDs"
+        }
+        10009 | 10020 => "Run `wrangler kv list <namespaceID>` to see your existing keys", // key errors
+        // TODO: link to more info
+        // limit errors
+        10022 | 10023 | 10024 | 10030 => "See documentation",
+        // TODO: link to tool for this
+        // legacy namespace errors
+        10021 | 10035 | 10038 => "Consider moving this namespace",
```
steveklabnik: i think this message is a bit confusing, but I don't have a great idea for a replacement right this second

src/main.rs
```diff
@@ -45,6 +45,19 @@ fn run() -> Result<(), failure::Error> {
         .author("ashley g williams <ashley666ashley@gmail.com>")
         .setting(AppSettings::ArgRequiredElseHelp)
         .setting(AppSettings::DeriveDisplayOrder)
+        .subcommand(
+            SubCommand::with_name("kv")
+                .about(&*format!(
+                    "{} Interact with your Workers KV Store",
```
steveklabnik: I would say "Interact with Workers KV"; this isn't wrong, but feels out of place to me. Can't really say why...

src/main.rs
```diff
@@ -45,6 +45,19 @@ fn run() -> Result<(), failure::Error> {
...
+                    emoji::KV
+                ))
+                .subcommand(
+                    SubCommand::with_name("add")
```
steveklabnik: we decided to go with "create", didn't we?

src/main.rs
```diff
@@ -233,6 +246,15 @@ fn run() -> Result<(), failure::Error> {
             .expect("The subdomain name you are requesting must be provided.");
 
         commands::subdomain(name, &user, &project)?;
+    } else if let Some(kv_matches) = matches.subcommand_matches("kv") {
+        match kv_matches.subcommand() {
+            ("add", Some(add_matches)) => {
```
steveklabnik: (will need to be changed here too)

src/terminal/emoji.rs
```diff
@@ -18,10 +18,11 @@ pub static DANCERS: Emoji = Emoji("👯 ", "");
 pub static EYES: Emoji = Emoji("👀 ", "");
 pub static INBOX: Emoji = Emoji("📥 ", "");
 pub static INFO: Emoji = Emoji("💁‍ ", "");
+pub static KV: Emoji = Emoji("🗂️ ", "");
```
steveklabnik: 💯

Cargo.toml
ashleygwilliams: it exists https://crates.io/crates/cloudflare

and as called out in the PR body this is blocked on a PR to that ;)

steveklabnik: Ah, I read it as "depends on the library" but not "depends on the specific release of the library :)"

ashleymichal: good catch tho, i should make an issue to, by the end of development on this feature, point at crates.io and not the git repo.

src/main.rs
ashleymichal: yes, but that was as of this morning. new commit incoming soon.

src/main.rs
ashleymichal: I can take the verbiage from the zoneless dashboard: 
```Store application data in the Cloudflare network and access your key-value pairs from Workers.```

ashleymichal: this is also not a PR against master, so, everything in this feature branch is open to change before being released.

of course it is always nicer to iron things out early and not have to think about them any more.

src/commands/kv/mod.rs
ashleymichal: still room for change before release.

Cargo.toml
ashleymichal: file issue: #401

src/main.rs
steveklabnik: cool, let's not stress about it too much then :)

Issue #393, opened 2019-08-04
Import @hapi/joi in Worker cause error


hbinduni:
Hi, i use `webpack` and try to import `@hapi/joi` like this:

```
import {config, Endpoint, S3} from 'aws-sdk';
import moment from 'moment';
import Joi from '@hapi/joi';
```

It build successfully but in preview, it return error:

`Error: Error("expected value", line: 2, column: 1)`

If i remove `import Joi` part, everything goes well. Can you help to **fixed** this?

<img width="376" alt="Screen Shot 2019-08-05 at 03 10 41" src="https://user-images.githubusercontent.com/1378410/62428848-aa995700-b730-11e9-8be1-bcef296c8e52.png">

below my `webpack.config.js`:

```
module.exports = {
  entry: __dirname + '/index.js',
  mode: 'production',
  optimization: {
    minimize: false,
  },
  performance: {
    hints: false,
  },
  devtool: 'cheap-module-source-map',
  node: {
    net: 'empty',
    tls: 'empty',
    dns: 'empty',
  },
  output: {
    path: __dirname + '/dist',
    publicPath: 'dist',
    filename: 'worker.js',
  },
};
```

I put another post in [cloudflare community for workers](https://community.cloudflare.com/t/import-hapi-joi-in-worker-cause-error/104415).

xtuc:
Thanks for the report. Could you please run the same command with a detailed stacktrace: `RUST_BACKTRACE=1 wrangler preview` and post it here?

hbinduni:
this is stacktrace i get:

```
RUST_BACKTRACE=1 wrangler preview
⬇️ Installing wasm-pack...
⬇️ Installing wranglerjs...
✨   Built successfully.
Error: Error("expected value", line: 2, column: 1)

stack backtrace:
   0: backtrace::backtrace::trace::ha005f61e3cfbd07b (0x1050a1afd)
   1: backtrace::capture::Backtrace::new_unresolved::h339da43cb4cb2a6a (0x1050a0c67)
   2: failure::backtrace::internal::InternalBacktrace::new::h2f8d45be066dd26d (0x1050a04b8)
   3: <failure::backtrace::Backtrace as core::default::Default>::default::h174bed90f39d5e6b (0x1050a06b4)
   4: wrangler::commands::publish::preview::preview::h65c0824e07e11a23 (0x104d26e99)
   5: wrangler::main::hef4cb33fcad6dad6 (0x104d3b7c3)
[...]

PR #394, opened 2019-08-05
Improve JSON errors debuggability

xtuc:
feat: improve JSON errors debuggability

As seen in https://github.com/cloudflare/wrangler/issues/393 previously
JSON decoding errors are obscure.

This changes improve the debugging in three cases:
  - The `package.json` file fails to decode; now emits a clearer error.
  - The wranglerjs backend returns an invalid JSON; now preserves the
  output file for further investigation.
    The console doesn't print the output file location by default, you
    need to pass `RUST_LOG=info` while running `wrangler build` and
    search for `--output-file=FILE` argument passed to wranglerjs.
  - The preview service returns invalid JSON; now emits a clearer error
    and full output can be seen using `RUST_LOG=info`


src/commands/build/wranglerjs/mod.rs
```diff
@@ -54,10 +53,10 @@ pub fn run_build(project: &Project) -> Result<(), failure::Error> {
             wranglerjs_output.project_size()
         );
 
+        fs::remove_file(temp_file)?;
         message::success(&msg);
         Ok(())
     } else {
-        fs::remove_file(temp_file)?;
```
ashleymichal: if we are preserving this for forensic purposes, should we output its location?

xtuc: I assume that "forensic purposes" means post-moderm analysis, in that case yes. The location is available when using `RUST_LOG=info`. I would prefer that the user don't know about it.

PR #399, opened 2019-08-05
Fix `wrangler config` for systems with non-unix EOL

### LINKED ISSUE #389: Wrangler can panic if global user configuration contains non printable characters
These lines will read until `\n`, which means on windows `email` and `api_key` will have `\r` at the end
https://github.com/cloudflare/wrangler/blob/1841a16b03435bf92a5766b4f36b8d748466265a/src/main.rs#L156
https://github.com/cloudflare/wrangler/blob/1841a16b03435bf92a5766b4f36b8d748466265a/src/main.rs#L158

which will cause a panic at these lines since `HeaderValue::from_str` requires its input to consist of solely printable ASCII characters.
https://github.com/cloudflare/wrangler/blob/1841a16b03435bf92a5766b4f36b8d748466265a/src/http.rs#L37
https://github.com/cloudflare/wrangler/blob/1841a16b03435bf92a5766b4f36b8d748466265a/src/http.rs#L38

xortive:
`wrangler config` was not properly truncating whitespace from the end of user input, resulting in a panic when trying to use `wrangler publish` as wrangler would try to create an HTTP header with invalid characters. Now, wrangler will properly truncate extra whitespace (including `\r`) from the end of input to `wrangler config`

Fixes #389

### DIFF
--- a/src/commands/config.rs
+++ b/src/commands/config.rs
@@ -17,11 +17,8 @@ pub fn set_file_mode(file: &PathBuf) {
         .expect("could not set permissions on file");
 }
 
-pub fn global_config(email: &str, api_key: &str) -> Result<(), failure::Error> {
-    let s = GlobalUser {
-        email: email.to_string(),
-        api_key: api_key.to_string(),
-    };
+pub fn global_config(email: String, api_key: String) -> Result<(), failure::Error> {
+    let s = GlobalUser { email, api_key };
 
     let toml = toml::to_string(&s)?;
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -153,11 +153,13 @@ fn run() -> Result<(), failure::Error> {
 
     if let Some(_matches) = matches.subcommand_matches("config") {
         println!("Enter email: ");
-        let email: String = read!("{}\n");
+        let mut email: String = read!("{}\n");
+        email.truncate(email.trim_end().len());
         println!("Enter api key: ");
-        let api_key: String = read!("{}\n");
+        let mut api_key: String = read!("{}\n");
+        api_key.truncate(api_key.trim_end().len());
 
-        commands::global_config(&email, &api_key)?;
+        commands::global_config(email, api_key)?;
     } else if let Some(matches) = matches.subcommand_matches("generate") {
         let name = matches.value_of("name").unwrap_or("worker");
         let project_type = match matches.value_of("type") {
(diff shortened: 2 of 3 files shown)


src/commands/config.rs
```diff
@@ -17,11 +17,8 @@ pub fn set_file_mode(file: &PathBuf) {
         .expect("could not set permissions on file");
 }
 
-pub fn global_config(email: &str, api_key: &str) -> Result<(), failure::Error> {
-    let s = GlobalUser {
-        email: email.to_string(),
-        api_key: api_key.to_string(),
-    };
+pub fn global_config(email: String, api_key: String) -> Result<(), failure::Error> {
```
xortive: while I was in here, I fixed the ownership semantics of global_config to remove a useless string copy.

src/main.rs
```diff
@@ -153,11 +153,13 @@ fn run() -> Result<(), failure::Error> {
 
     if let Some(_matches) = matches.subcommand_matches("config") {
         println!("Enter email: ");
-        let email: String = read!("{}\n");
+        let mut email: String = read!("{}\n");
+        email.truncate(email.trim_end().len());
```
xortive: the truncate dance here is to avoid allocating an extra string, and to keep email as an owned String instead of a string slice

tests/config.rs
```diff
@@ -5,15 +5,24 @@ use std::io::prelude::*;
...
+#[test]
+fn it_generates_the_config_windows_eol() {
+    generate_config_with("\r\n");
+}
+
+fn generate_config_with(eol: &str) {
     let fake_home_dir = env::current_dir()
         .expect("could not retrieve cwd")
-        .join(".it_generates_the_config");
+        .join(format!(".it_generates_the_config_{}", random_chars(5)));
```
xortive: because `cargo test` is multithreaded, I had to add this random_char thing to the end of the home directory name to keep the threads from deleting each other's home dir. We should probably enforce this for all tests that create directories in the future.

xortive:
Also: wanted review on the actual PR title and message content, trying out formatting these such that they work well when copy-pasted into a changelog.

src/main.rs
```diff
@@ -153,11 +153,13 @@ fn run() -> Result<(), failure::Error> {
...
         println!("Enter api key: ");
-        let api_key: String = read!("{}\n");
+        let mut api_key: String = read!("{}\n");
+        api_key.truncate(api_key.trim_end().len());
```
EverlastingBugstopper: 😻

PR #407, opened 2019-08-07
Add section to readme about Vendored OpenSSL

### LINKED ISSUE #361: Build fails on Linux due to OpenSSL dependencies
I tried to build Wrangler on Ubuntu and had to install `pkg-config` and `libssl-dev`, I believe this means that vendored openSSL isn't working correctly? do we need openssl as a direct dependency if we don't use it directly?

xortive:
Wrangler has some [external dependencies](https://docs.rs/openssl/0.10.24/openssl/#automatic) when installing on linux, and we didn't document the `--features vendored-openssl` flag to allow those without `pkg-config` and `libssl-dev` installed to install wrangler.

Closes #361

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,12 @@ Wrangler can be installed both through [npm](https://www.npmjs.com/get-npm) and 
     cargo install wrangler
     ```
 
+    Installing wrangler on linux requires some [OpenSSL-related packages](https://docs.rs/openssl/0.10.24/openssl/#automatic) to be installed. If you don't want to deal with this, you can use vendored OpenSSL.
+
+    ```
+    cargo install wrangler --features vendored-openssl
+    ```
+
 ### Manual Install:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -190,6 +190,12 @@ Wrangler can be installed both through [npm](https://www.npmjs.com/get-npm) and
     cargo install wrangler
     ```
 
+    Installing wrangler on linux requires some [OpenSSL-related packages](https://docs.rs/openssl/0.10.24/openssl/#automatic) to be installed. If you don't want to deal with this, you can use vendored OpenSSL.
+
+    ```
+    cargo install wrangler --features vendored-openssl
```
EverlastingBugstopper: do we have option to do this w/npm?

xortive: this isn't necessary for npm, npm distributes built binaries that already utilize vendored openssl for linux.

PR #414, opened 2019-08-08
Fix capitalization in README

Electroid:
Very small change, but noticed we don't capitalize workers at the top of the README.


EverlastingBugstopper:
Thanks @Electroid!

PR #419, opened 2019-08-09
feat(npm-installer): install current version, not latest

### LINKED ISSUE #418: Wrangler Installer should point to a specific version of Wrangler
Related to #380.

ashleygwilliams:
fixes #418


ashleymichal:
use function that takes tag. later step can get the tag from Cargo.toml and pass to function.

also output error message on request failure, rather than throwing.

PR #427, opened 2019-08-12
Nits: fix clippy warnings

xortive:
Cleaned up some clippy warnings that got past CI.

### DIFF
--- a/src/commands/publish/package.rs
+++ b/src/commands/publish/package.rs
@@ -38,7 +38,7 @@ impl Package {
 
         let package_json: String = fs::read_to_string(manifest_path.clone())?.parse()?;
         let package: Package = serde_json::from_str(&package_json)
-            .expect(&format!("could not parse {:?}", manifest_path));
+            .unwrap_or_else(|_| panic!("could not parse {:?}", manifest_path));
 
         Ok(package)
     }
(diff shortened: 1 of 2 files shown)


src/terminal/message.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(dead_code)]
```
xortive: normally would be against this type of hammer, but I think this is an appropriate nail.

EverlastingBugstopper: i'm confused is it a hammer or a nail 😉

src/commands/publish/package.rs
```diff
@@ -38,7 +38,7 @@ impl Package {
 
         let package_json: String = fs::read_to_string(manifest_path.clone())?.parse()?;
         let package: Package = serde_json::from_str(&package_json)
-            .expect(&format!("could not parse {:?}", manifest_path));
+            .unwrap_or_else(|_| panic!("could not parse {:?}", manifest_path));
```
gabbifish: Out of curiosity, why did `expect` cause these warnings?

xortive: https://rust-lang.github.io/rust-clippy/master/index.html#expect_fun_call

PR #429, opened 2019-08-12
feat: authenticate calls to preview service when possible

### LINKED ISSUE #423: Preview throws Server Error: 502 Bad Gateway
#### Environment

* operating system: macOS
* output of `rustc -V`:
* output of `node -v`:
* output of `wrangler -V`:

```
➜ rustc -V
rustc 1.35.0 (3c235d560 2019-05-20)

➜ node -v                           
v10.16.0

➜ RUST_LOG=info RUST_BACKTRACE=1 ./node_modules/.bin/wrangler --version
👷 ✨   wrangler 1.1.1
```

#### Steps to reproduce

Run `wrangler preview`

Clean build and display of preview.

```
➜ RUST_LOG=info RUST_BACKTRACE=1 ./node_modules/.bin/wrangler preview                    
[2019-08-11T22:01:13Z INFO  wrangler] Getting project settings
[2019-08-11T22:01:13Z INFO  wrangler::commands::build::wranglerjs] skipping npm install because node_modules exists
[2019-08-11T22:01:13Z INFO  wrangler::install] prebuilt artifact https://github.com/cloudflare/wrangler/releases/download/v1.1.1/wranglerjs-v1.1.1.tar.gz
⬇️ Installing wranglerjs...
[2019-08-11T22:01:13Z INFO  wrangler::commands::build::wranglerjs] wranglerjs downloaded at: "/Users/glenn/Library/Caches/.wrangler/wranglerjs-5b6f918ed4519a00"
[2019-08-11T22:01:13Z INFO  wrangler::commands::build::wranglerjs] skipping npm install because node_modules exists
[2019-08-11T22:01:13Z INFO  wrangler::install] prebuilt artifact https://github.com/rustwasm/wasm-pack/releases/download/v0.8.1/wasm-pack-v0.8.1-x86_64-apple-darwin.tar.gz
⬇️ Installing wasm-pack...
[2019-08-11T22:01:13Z INFO  wrangler::commands::build::wranglerjs] Running "/Users/glenn/.nvm/versions/node/v10.16.0/bin/node" "/Users/glenn/Library/Caches/.wrangler/wranglerjs-5b6f918ed4519a00" "--output-file=/var/folders/cg/q568zb_17fv6qx4r2lpl3g0w0000gn/T/.wranglerjs_outputwZuMa" "--wasm-binding=wasmprogram" "--no-webpack-config=1" "--use-entry=/Users/glenn/src/github.com/MYPROJ/MYPROJ-api-cloudflare/index.js"
✨   Built successfully, built project size is 22 KiB.
[2019-08-11T22:01:15Z INFO  wrangler::commands::publish::upload_form] Webpack project detected. Publishing...
[2019-08-11T22:01:15Z INFO  wrangler::commands::publish::upload_form] Form { boundary: "145b6e631ed14205892b639eb78c6655", parts: [("metadata", Part { value: Body { kind: b"{\"bindings\":[{\"name\":\"MYPROJ_API\",\"namespace_id\":\"REDACTED\",\"type\":\"kv_namespace\"}],\"body_part\":\"script\"}" }, mime: Some("application/json"), file_name: Some("metadata.json"), headers: {} }), ("script", Part { value: Body { kind: Reader { length: 66475 } }, mime: Some("application/javascript"), file_name: Some("script.js"), headers: {} })] }
Error: https://cloudflareworkers.com/script: Server Error: 502 Bad Gateway
stack backtrace:
   0: backtrace::backtrace::trace::h22584d25fbba1826 (0x10cfeb72e)
   1: backtrace::capture::Backtrace::new_unresolved::hfc9f84de61204b0b (0x10cfea818)
   2: failure::backtrace::internal::InternalBacktrace::new::hba46b32ce7838972 (0x10cfea029)
   3: <failure::backtrace::Backtrace as core::default::Default>::default::h24970bd4cf762182 (0x10cfea225)
   4: wrangler::commands::publish::preview::preview::h38fb112a3adb7a96 (0x10ccc775b)
   5: wrangler::run::h685661c5e5cdd2ad (0x10ccde593)
   6: wrangler::main::h29514711cbac4433 (0x10ccdbe75)
   7: std::rt::lang_start::{{closure}}::ha67eafa477c8744a (0x10ccdef56)
   8: std::panicking::try::do_call::h8037d9f03e27d896 (0x10d0fe8b8)
   9: __rust_maybe_catch_panic (0x10d1082ef)
  10: std::rt::lang_start_internal::hc8e69e673740d4ae (0x10d0ff39e)
  11: main (0x10ccded49)

➜ RUST_LOG=info RUST_BACKTRACE=1 ./node_modules/.bin/wrangler --version
👷 ✨   wrangler 1.1.1

```

FYI, running `wrangler build`, or `wrangler publish` runs cleanly and works as expected.

ashleymichal:
Today Wrangler uses the basic service on cloudflareworkers.com to run its preview command. This PR utilizes the authenticated endpoints, which unlocks desirable features such as kv binding support for wrangler preview specifically.

This PR also includes a refactor to Project, which moves the `validate_project` function out of publish and makes it a public method on Project. In hindsight, preview could probably have used it as a function? It could go either way.

Another refactor makes the preview_address a static &str const, as it is always the same, hard coded value.

Finally, I unwrapped the method passed as an arg and passed it unwrapped.

Closes #423 and #431


src/commands/publish/preview/mod.rs
```diff
@@ -18,7 +19,63 @@ struct Preview {
...
+        project.account_id, project.name
+    );
+
+    let client = http::auth_client(user);
+
+    commands::build(&project)?;
+
+    let script_upload_form = publish::build_script_upload_form(project)?;
+
+    log::info!("💩");
```
steveklabnik: 🤣

ashleymichal: lol i gotta stop asking for review on WIP PRs

src/commands/publish/preview/mod.rs
```diff
@@ -18,7 +19,63 @@ struct Preview {
...
+    let res = client
+        .post(&create_address)
+        .multipart(script_upload_form)
+        .send()?
+        .error_for_status();
+
+    let text = &res?.text()?;
```
steveklabnik: So, I don't think that this is an *inherently* bad line, but it is a little punctuation heavy. I'd move the first `?` to be on line 44 above, and then move the `&` to the call on line 50.

ashleymichal:
> Instead of splitting the auth and no_auth previews up, keeping one preview function with a function that performs the upload and returns the script Id (and encapsulates the auth_client, checking for user config, etc stuff) will make this much easier to merge with the livereload branch

I am glad that is the case, because that's approximately the refactor i was headed toward!

src/commands/publish/preview/mod.rs
```diff
@@ -19,35 +20,30 @@ struct Preview {
...
 
 pub fn preview(
-    project: &Project,
     method: Result<HTTPMethod, failure::Error>,
     body: Option<String>,
 ) -> Result<(), failure::Error> {
-    let create_address = "https://cloudflareworkers.com/script";
-
-    let client = http::client();
+    let project = Project::new()?;
```
xortive: Why aren’t we passing project in anymore?

src/commands/publish/preview/mod.rs
```diff
@@ -19,35 +20,30 @@ struct Preview {
...
     commands::build(&project)?;
 
-    let script_upload_form = publish::build_script_upload_form(project)?;
-
-    let res = client
-        .post(create_address)
-        .multipart(script_upload_form)
-        .send()?
-        .error_for_status();
+    let user = match GlobalUser::new() {
```
xortive: Use Result::ok()

src/commands/publish/preview/mod.rs
```diff
@@ -69,6 +65,36 @@ pub fn preview(
...
+    let script_upload_form = publish::build_script_upload_form(project)?;
+
+    let mut res = client
+        .post(&create_address)
+        .multipart(script_upload_form)
+        .send()?
+        .error_for_status()?;
+
+    let text = &res.text()?;
+    log::info!("Response from preview: {:?}", text);
```
gabbifish: Are we sure we want to log the response from preview before or after we parse it with serde? I would think we'd want folks to see the response after parsing succeeds?

src/commands/publish/preview/mod.rs
ashleymichal: my reason isn't enough to change the overall pattern. swapping back.

src/commands/publish/preview/mod.rs
```diff
@@ -69,6 +60,36 @@ pub fn preview(
...
+        Some(_user) => format!(
+            "https://api.cloudflare.com/client/v4/accounts/{}/workers/scripts/{}/preview",
+            project.account_id, project.name
+        ),
+        None => "https://cloudflareworkers.com/script".to_string(),
+    };
+
+    let script_upload_form = publish::build_script_upload_form(project)?;
+
+    let mut res = client
```
gabbifish: I'm guessing this is where headers logic (for the passing of email and the api token) will live :)

gabbifish: We'll probably need two cases of upload call--one for auth, one for not?

gabbifish: Oooh, does the cf apiclient create allow us to call the v4 api here (endpoint https://api.cloudflare.com/client/v4/accounts/{}/workers/scripts/{}/preview)? I think that would be super useful to use.

src/commands/publish/preview/mod.rs
ashleymichal: anywhere in the code you see `log::info!()`, we are planting debug logs. they don't show unless users pass `RUST_LOG=info` as an env var.

src/commands/publish/preview/mod.rs
ashleymichal: the only differences are in the url we use, and in the auth headers. auth headers are set in the client (preview constructs either an auth client or a regular client), and the url is set above (create_address). If the naming is confusing tho we should fix that.

ashleymichal: > Oooh, does the cf apiclient crate allow us to call the v4 api here (endpoint [https://api.cloudflare.com/client/v4/accounts/{}/workers/scripts/{}/preview](https://api.cloudflare.com/client/v4/accounts/%7B%7D/workers/scripts/%7B%7D/preview))? I think that would be super useful to use.

i think i know what you're asking; if by that you mean https://github.com/cloudflare/cloudflare-rs, eventually i want to add workers api support there as well and make it our de-facto v4 api client, but i'm not there yet.

gabbifish: ^ You're correct about what I'm asking! We can save a migration to the cloudflare-rs client for later; this could be worth mentioning in a `//todo` comment. I also now see how the headers are set in the client that's passed into upload_to_preview--I think we can leave that untouched for now. :)

ashleymichal: the cloudflare-rs client migration was one of my Q3 goals; i'm inclined to write an issue for it rather than a comment.

ashleymichal:
> Just realized after testing on my machine that this currently doesn't account for the case where there is a global user config, but account_id isn't configured. This also breaks tests locally if one has a global user config, since the tests for preview don't configure account_id

oo good point. mine probably pass because i have account id as an environment variable as well. I'll try to update the tests, but also, our tests should maybe set/unset env rather than taking it from execution env? dunno. gonna have to think on that.

src/commands/publish/mod.rs
```diff
@@ -20,7 +20,7 @@ use crate::terminal::message;
 pub fn publish(user: &GlobalUser, project: &Project, release: bool) -> Result<(), failure::Error> {
     info!("release = {}", release);
 
-    validate_project(project, release)?;
+    project.validate(release)?;
```
EverlastingBugstopper: 🔥

src/commands/publish/preview/mod.rs
```diff
@@ -10,63 +10,158 @@ use uuid::Uuid;
 
 use crate::commands;
 use crate::http;
+use crate::settings::global_user::GlobalUser;
 use crate::settings::project::Project;
 use crate::terminal::message;
+use reqwest::Client;
+
+// Using this instead of just `https://cloudflareworkers.com` returns just the worker response to the CLI
+const PREVIEW_ADDRESS: &str = "https://00000000000000000000000000000000.cloudflareworkers.com";
```
EverlastingBugstopper: this endpoint gets me every time

src/commands/publish/preview/mod.rs
```diff
@@ -10,63 +10,158 @@ use uuid::Uuid;
...
+    user: Option<GlobalUser>,
+    method: HTTPMethod,
     body: Option<String>,
 ) -> Result<(), failure::Error> {
-    let create_address = "https://cloudflareworkers.com/script";
+    let client: Client;
+
+    let preview = match &user {
+        Some(user) => {
+            log::info!("Running in Auth'd mode");
```
EverlastingBugstopper: don't think it matters but i think i would just say `authenticated` instead of `auth'd` - i know what it means, just seems like a `grep authenticated` would be easier to remember than a `grep Auth'd`

src/commands/publish/preview/mod.rs
```diff
@@ -10,63 +10,158 @@ use uuid::Uuid;
...
+
+            project.validate(false)?;
+
+            commands::build(&project)?;
+            client = http::auth_client(&user);
+
+            authenticated_upload(&client, &project)?
+        }
+        None => {
+            log::info!("Running in Un-auth'd mode");
```
EverlastingBugstopper: same here - `un-auth'd` seems clunky

src/commands/publish/preview/mod.rs
```diff
@@ -10,63 +10,158 @@ use uuid::Uuid;
...
+
+            // KV namespaces are not supported by the preview service unless you authenticate
+            // so we omit them and provide the user with a little guidance. We don't error out, though,
+            // because there are valid workarounds for this for testing purposes.
+            if project.kv_namespaces.is_some() {
+                message::warn("KV Namespaces are not supported in unauthenticated mode");
```
EverlastingBugstopper: `unauthenticated mode` seems opaque to me, maybe `You can't preview your script with KV namespaces without configuring wrangler to use your Cloudflare credentials` or something similar?

src/commands/publish/preview/mod.rs
```diff
@@ -10,63 +10,158 @@ use uuid::Uuid;
...
+            commands::build(&project)?;
+            client = http::client();
+
+            unauthenticated_upload(&client, &project)?
+        }
+    };
+
+    let worker_res = call_worker(&client, preview, method, body)?;
+
+    let msg = format!("Your worker responded with: {}", worker_res);
```
EverlastingBugstopper: Does this need a ✨ emoji?

src/commands/publish/preview/mod.rs
```diff
@@ -10,63 +10,158 @@ use uuid::Uuid;
...
         .post(create_address)
         .multipart(script_upload_form)
         .send()?
-        .error_for_status();
+        .error_for_status()?;
 
-    let text = &res?.text()?;
-    log::info!("Response from preview: {:?}", text);
+    let text = &res.text()?;
+    log::info!("Response from preview: {:#?}", text);
```
EverlastingBugstopper: 🔥

src/commands/publish/preview/mod.rs
```diff
@@ -10,63 +10,158 @@ use uuid::Uuid;
...
 
-    let p: Preview =
+    let preview: Preview =
```
EverlastingBugstopper: 🔥

src/settings/project/mod.rs
```diff
@@ -71,6 +71,65 @@ impl Project {
...
+        };
+        if self.name.is_empty() {
+            missing_fields.push("name")
+        };
+
+        match &self.kv_namespaces {
+            Some(kv_namespaces) => {
+                for kv in kv_namespaces {
+                    if kv.binding.is_empty() {
+                        missing_fields.push("kv-namespace binding")
```
EverlastingBugstopper: `kv-namespace` -> `kv-namespaces` to match .toml?

src/commands/publish/preview/mod.rs
ashleymichal: maybe "without credentials"? i follow up with the help right after.

src/commands/publish/preview/mod.rs
ashleymichal: i mean idk, this is how it is today.

src/commands/publish/preview/mod.rs
ashleymichal: yeah this only mattered bc the response is embedded in a v4 struct when it goes through the API

src/commands/publish/preview/mod.rs
EverlastingBugstopper: yeah that works

src/commands/publish/preview/mod.rs
EverlastingBugstopper: oh actually it has the worker emojis.. i wonder if that gets added elsewhere

```
👷  Your worker responded with: Hello worker! 👷 
```

src/commands/publish/preview/mod.rs
```diff
@@ -10,63 +10,158 @@ use uuid::Uuid;
...
+            authenticated_upload(&client, &project)?
+        }
+        None => {
+            log::info!("GlobalUser not set, running without authentication");
+
+            // KV namespaces are not supported by the preview service unless you authenticate
+            // so we omit them and provide the user with a little guidance. We don't error out, though,
+            // because there are valid workarounds for this for testing purposes.
+            if project.kv_namespaces.is_some() {
+                message::warn("KV Namespaces are not supported without setting API credentials");
```
gabbifish: Nice!

src/commands/publish/preview/mod.rs
```diff
@@ -10,63 +10,158 @@ use uuid::Uuid;
...
+    let text = &res.text()?;
+    log::info!("Response from preview: {:#?}", text);
+
+    let response: V4ApiResponse =
+        serde_json::from_str(text).expect("could not create a script on cloudflareworkers.com");
+
+    Ok(Preview::from(response.result))
+}
+
+fn unauthenticated_upload(client: &Client, project: &Project) -> Result<Preview, failure::Error> {
```
gabbifish: I think splitting uploads into two separate functions (auth and unauth) was the right call.

ashleymichal:
> Breaking changes for newly generated projects. You shouldn't need to use authenticated preview w/o KV
> 
> ```shell
> ❯ wrangler generate
> 🌀  🐑  Generating a new webpack worker project with name 'worker'...
> 🔧   Creating project called `worker`...
> ✨   Done! New project created /Users/averyharnish/Documents/work/wrangler/worker
> ❯ cd worker
> ❯ wrangler preview
> Error: Your wrangler.toml is missing the field ["account_id"] which is required to publish to your subdomain!
> ```

The quick fix to this (and the general complaint that one has to populate their account id) would be to tell folks to set the appropriate env var. But i'm open to other thoughts, we could potentially fallback to unauthenticated and warn about it the way we do with kv.

EverlastingBugstopper:
This might just be a problem with `wrangler config` but if I `wrangler config` and just press enter twice it stores my credentials as empty strings. When I run `wrangler preview` with no `account_id` it falls back to unauthenticated. Fine. When I put in my account id into the toml, I get this
```bash
$ wrangler preview
✨   Built successfully, built project size is 517 bytes.
Error: https://api.cloudflare.com/client/v4/accounts/e04d3d27a6829775c8ef937145d7f3ec/workers/scripts/worker/preview: Client Error: 400 Bad Request
```
yuck

EverlastingBugstopper:
I also don't get a warning about empty config + falling back to unauthenticated if I don't have a global config set up

```bash
❯ rm /Users/averyharnish/.wrangler/config/default.toml 
❯ wrangler preview 
✨   Built successfully, built project size is 517 bytes.
👷  GET https://00000000000000000000000000000000.cloudflareworkers.com
👷  Your worker responded with: Hello worker!
❯ wrangler whoami
Error: ⚠️  Your global config has an error, run `wrangler config`: missing field `email`
```

ashleymichal:
> I also don't get a warning about empty config + falling back to unauthenticated if I don't have a global config set up
> 
> ```shell
> ❯ rm /Users/averyharnish/.wrangler/config/default.toml 
> ❯ wrangler whoami
> Error: ⚠️  Your global config has an error, run `wrangler config`: missing field `email`
> ❯ wrangler preview 
> ✨   Built successfully, built project size is 517 bytes.
> 👷  GET https://00000000000000000000000000000000.cloudflareworkers.com
> 👷  Your worker responded with: Hello worker!
> ```
> 
> This one I think we should probably address here

I'll check this out today.

Issue #384, opened 2019-08-02
Rust projects are not minified


EverlastingBugstopper:
![image](https://user-images.githubusercontent.com/9408157/62386044-4aca6280-b51c-11e9-8cf8-5188ffa11ded.png)

image description: screenshot of unminified JavaScript for the default Rust project generated by wrangler

ashleygwilliams:
would be great to "fix" my "concat" function while solving this ;)

gabbifish:
Follow-up as I work on this: Do we want to use unminified JS for the preview, and minified JS purely for publication? I think this would be the right choice--you never know if folks want to poke around the concatenated js code in preview mode.

ashleymichal:
> Follow-up as I work on this: Do we want to use unminified JS for the preview, and minified JS purely for publication? I think this would be the right choice--you never know if folks want to poke around the concatenated js code in preview mode.

I'd agree, except that there are some scripts that will go well over the size limit when not minified, and the preview service enforces the same limits. It might be that we give folks an option? or use environments to allow for custom webpack configs that do or do not minify js.

gabbifish:
^ I like ashley's suggestion of making this an option; perhaps we can make minification the default and allow people to specify unminified code for debugging (this logic could live in the config, for example).

If we were to make this something users can control, are we okay with widening the scope of this PR? Or is this veering into bike shedding territory?

EverlastingBugstopper:
We already allow users to specify their own webpack configuration with `webpack_config` in `wrangler.toml` :)

gabbifish:
^ I think webpack_config is only used for Webpack projects, not Rust ones. Let's shelve the option-setting functionality to another issue; I'll focus on just getting basic minification to work :)

EverlastingBugstopper:
It looks like we aren't going to get to this for 1.2.0 which is fine. Removing the milestone but let's keep the issue open and return when we have a better game plan

gabbifish:
We are going to close this ticket in favor of having users specify minification options via webpack configs. For Rust workers in particular, users can depend on wasm-pack-plugin to define their desired minification behaviror. See https://github.com/cloudflare/workers-docs/issues/568

PR #430, opened 2019-08-12
Minify rust-wasm worker code

gabbifish:
This PR addresses #384. It streamlines the joining of the rust-wasm "glue" code and the actual worker js code, and minifies it.

I have a concern with the minification crate used for this (which is the only js minification crate for rust, afaik): https://crates.io/crates/minifier. This crate will not automatically add missing semicolons to distinguish between js lines of code, unlike other, more developed minifiers. This means that even one missing semicolon in the worker/worker.js code for rust-wasm projects will throw parsing errors! 

Even a lot of our example/template code lacks proper semicolon usage...
```js
})

/**
 * Fetch and log a request
 * @param {Request} request
 */
    const { greet } = wasm_bindgen;
    await wasm_bindgen(wasm)
    const greeting = greet()
    return new Response(greeting, {status: 200})
}
```

I fear that the minifier's need for perfect semicolon usage will honestly make the rust-wasm worker dev experience worse. We can either shelve this idea for now, or explore using a more advanced (though heavier to integrate) set of minification tooling like that in webpack.

I guess issue #384 is less of a low-hanging fruit than we thought!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -816,6 +816,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "macro-utils"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "matches"
 version = "0.1.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


ashleymichal:
> I fear that the minifier's need for perfect semicolon usage will honestly make the rust-wasm worker dev experience worse.

I think i agree with your gut here, i'd rather stay less than ideal technically and not make the exp worse. if i knew of a better implementation, i'd suggest it, but it doesn't seem worth churning on just yet imho. perhaps there are others with better knowledge of the rust ecosystem for this here though.

xtuc:
The package doc mentions:
> Please be aware that this is still at a very early stage of development so you shouldn't rely on it too much!

Looking at the source, I doubt the parser is spec compliant and will be updated with new features (no private fields support).

The minifier keeps the licence comments, which we don't want I guess.

I think this will break user code.

ashleygwilliams:
based on the current reviews it sounds as tho we want to close this PR? if not let me know!

EverlastingBugstopper:
I think so

xtuc:
Sorry Gabbi!

From my perspective I think it's fine to not minify them.

gabbifish:
Let's close this PR for now in favor of having users provide minification options via wasm-pack-plugin, tracked here: https://github.com/cloudflare/workers-docs/issues/568

PR #432, opened 2019-08-13
Fix ignoring exit code from spawnSync - run-wrangler.js

### LINKED ISSUE #335: Build failure returns code 0
wrangler build always returns 0 even when it fails to build

### LINKED ISSUE #433: @cloudflare/wrangler publish incorrectly exiting zero on error
#### Environment

* operating system: mac os 10.15
* output of `node -v`: v10.15.3
* output of `wrangler -V`: 1.1.1

#### Steps to reproduce

I had some api key issues while publishing a route to a specific zone. And wrangler (run from npm) was not exiting correctly (failing silently).

Reproduction:

1. add wrangler to package.json (yarn add @cloudflare/wrangler -D)
2. run a wrangler command that should fail

`CF_EMAIL=example@example.com CF_API_KEY=incorrect_api_key_here wrangler publish --release`

Exit should be non-zero with an authentication error

Exit code was zero.

defjosiah:
Fixes #433
Fixes #335

I had some api key issues while publishing a route to a specific zone.
I realized that the npm "run-wrangler" wrapper, was exiting zero when our api key was incorrect,
when I expected it to exit non-zero.

Looks like "spawnSync" returns an explicit status field, which has to be propagated up in order to 
have the correct exit status.

This diff fixes that (removes an unused import, and fixes some quotes).

Reproduction:
```
# add wrangler to package.json (yarn add @cloudflare/wrangler -D)
CF_EMAIL=example@example.com CF_API_KEY=incorrect_api_key_here wrangler publish --release
echo $? # will be zero
```

After this fix
```
CF_EMAIL=example@example.com CF_API_KEY=incorrect_api_key_here wrangler publish --release
# error Command failed with exit code 1.
```

### DIFF
--- a/npm/run-wrangler.js
+++ b/npm/run-wrangler.js
@@ -1,8 +1,8 @@
 #!/usr/bin/env node
 
-const { join, resolve } = require("path");
+const { join } = require("path");
 const { spawnSync } = require("child_process");
-const { homedir } = require('os');
+const { homedir } = require("os");
 
 const cwd = join(homedir(), ".wrangler");
 const bin = join(cwd, "out", "wrangler");
@@ -12,4 +12,4 @@ const opts = {
   cwd: process.cwd(),
   stdio: "inherit"
 };
-spawnSync(bin, args, opts);
+process.exit(spawnSync(bin, args, opts).status);


xortive:
Good work! Sounds like this fixes #335 as well. Thank you for the PR!

defjosiah:
Happy to help! We were publishing using serverless framework before, wrangler is so nice 😀

xortive:
I'm testing that npm install correctly replaces the bin script with the new one and got this error
```

PR #441, opened 2019-08-15
Lock all builds in CI

xortive:
Our CI wasn't locking builds, and a downstream dependency broke semver earlier this morning and broke CI. This PR adds the `--locked` flag to all `cargo install` and `cargo build` commands that run in CI to protect us from this happening again.


ashleygwilliams:
closing in favor of #443

PR #443, opened 2019-08-15
Don't install cargo-generate in ci

EverlastingBugstopper:
Fixes broken build due to upstream dependency


EverlastingBugstopper:
fwiw this is fixed upstream https://github.com/Goncalerta/proc-quote/issues/15

PR #445, opened 2019-08-15
Add bulk upload to Workers KV

### LINKED ISSUE #346: Add KV Command: Bulk Set Key/Value Pair on Namespace
Related to #343 - we should also support bulk upload

As a Workers KV user, I should be able to set many key/value pairs on my KV namespace from Wrangler at once.

example:
``` sh
$ wrangler kv write bulk path/to/assets
```
the path should point to either:
- a directory of assets to be uploaded with paths as keys OR
- a `.json` file which defines an array of key/value pairs such as:
``` json
[
  {
    "key": "hero",
    "value": "batman",
    "expiration": 1565912132
  },
  {
    "key": "villain",
    "value": "joker",
    "expiration": 1565912132
  }
]
```
we should support all keys specified by the api documentation in these jsons, that is:
* `expiration`
* `expiration_ttl`
* `base64`

gabbifish:
This PR adds support for bulk uploads to Workers KV; it currently supports reading from JSON files and uploading directories. The subcommand is used like the following:
```
wrangler kv write bulk <NAMESPACE_ID> <JSON_FILE_OR_DIRECTORY> 
```
JSON input is of the format 
```
[{"key": "gabbi", "value": "fisher", "expiration_ttl": <optional_second_value>, "expiration": <optional epoch_second_value>, "base64": <optional_true_or_false>}, ...]
```
and wrangler output looks like
```
$ wrangler kv write bulk 06779da6940b431db6e566b4846d64db test.json 
 Success
```

Output for a directory upload looks like
```
$ wrangler kv write bulk 06779da6940b431db6e566b4846d64db abc
 Uploading a.txt...
 Uploading b%2Fc.txt...
 Success
```
^ the `working` swirl emoji is used here for each `Uploading...` line but just doesn't show up on my Ubuntu installation for now. :)

This PR will fail to build until the necessary API changes are also merged into cloudflare-rs (https://github.com/cloudflare/cloudflare-rs/pull/28), because it is dependent on these changes.

This PR closes #346.

### DIFF
--- a/src/commands/kv/write_bulk.rs
+++ b/src/commands/kv/write_bulk.rs
@@ -1,0 +1,106 @@
+extern crate base64;
+
+use cloudflare::framework::apiclient::ApiClient;
+use walkdir::WalkDir;
+
+use std::ffi::OsString;
+use std::fs;
+use std::fs::metadata;
+use std::path::Path;
+
+use cloudflare::endpoints::workerskv::write_bulk::KeyValuePair;
+use cloudflare::endpoints::workerskv::write_bulk::WriteBulk;
+use failure::bail;
+
+use crate::terminal::message;
+
+pub fn write_bulk(namespace_id: &str, filename: &Path) -> Result<(), failure::Error> {
+    let client = super::api_client()?;
+    let account_id = super::account_id()?;
+
+    // If the provided argument for write_bulk is a json file, parse it
+    // and upload its contents. If the argument is a directory, create key-value
+    // pairs where keys are the relative pathnames of files in the directory, and
+    // values are the base64-encoded contents of those files.
+    let mut data;
+    let pairs: Result<Vec<KeyValuePair>, failure::Error> = match metadata(filename) {
+        Ok(ref file_type) if file_type.is_file() => {
+            data = fs::read_to_string(filename)?;
+            Ok(serde_json::from_str(&data)?)
+        }
+        Ok(ref file_type) if file_type.is_dir() => parse_directory(filename),
+        Ok(_file_type) => {
+            // any other file types (namely, symlinks)
+            bail!(
+                "Cannot upload a file that is a symlink: {}",
+                filename.display()
+            )
+        }
+        Err(e) => bail!(e),
+    };
+
+    let response = client.request(&WriteBulk {
+        account_identifier: &account_id,
+        namespace_identifier: namespace_id,
+        bulk_key_value_pairs: pairs?,
+    });
+
+    match response {
+        Ok(_success) => message::success("Success"),
+        Err(e) => super::print_error(e),
+    }
+
+    Ok(())
+}
+
+fn parse_directory(directory: &Path) -> Result<Vec<KeyValuePair>, failure::Error> {
+    let mut upload_vec: Vec<KeyValuePair> = Vec::new();
+    for entry in WalkDir::new(directory) {
+        let entry = entry.unwrap();
+        let path = entry.path();
+        if path.is_file() {
+            let key = generate_key(path, directory)?;
+
+            let value = std::fs::read(path)?;
+
+            // Need to base64 encode value
+            let b64_value = base64::encode(&value);
+            message::working(&format!("Uploading {}...", key.clone()));
+            upload_vec.push(KeyValuePair {
+                key: key,
+                value: b64_value,
+                expiration: None,
+                expiration_ttl: None,
+                base64: Some(true),
+            });
+        }
+    }
+    Ok(upload_vec)
+}
+
+// Courtesy of Steve Kalabnik's PoC :)
+fn generate_key(path: &Path, directory: &Path) -> Result<String, failure::Error> {
+    let path = path.strip_prefix(directory).unwrap();
+
+    // next, we have to re-build the paths: if we're on Windows, we have paths with
+    // `\` as separators. But we want to use `/` as separators. Because that's how URLs
+    // work.
+    let mut path_with_forward_slash = OsString::new();
+
+    for (i, component) in path.components().enumerate() {
+        // we don't want a leading `/`, so skip that
+        if i > 0 {
+            path_with_forward_slash.push("/");
+        }
+
+        path_with_forward_slash.push(component);
+    }
+
+    // if we have a non-utf8 path here, it will fail, but that's not realistically going to happen
+    let path = path_with_forward_slash.to_str().expect(&format!(
+        "found a non-UTF-8 path, {:?}",
+        path_with_forward_slash
+    ));
+
+    Ok(path.to_string())
+}
--- a/src/main.rs
+++ b/src/main.rs
@@ -4,6 +4,7 @@
 extern crate text_io;
 
 use std::env;
+use std::path::Path;
 use std::str::FromStr;
 
 use clap::{App, AppSettings, Arg, SubCommand};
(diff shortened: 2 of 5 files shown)


src/commands/kv/write_bulk.rs
```diff
@@ -0,0 +1,40 @@
+use cloudflare::apiclient::ApiClient;
+
+use std::fs;
+
+use cloudflare::workerskv::write_bulk::BulkWrite;
```
gabbifish: imma consolidate these imports into one line

src/main.rs
```diff
@@ -286,9 +325,28 @@ fn run() -> Result<(), failure::Error> {
...
+                    ("bulk", Some(bulk_write_matches)) => {
+                        let id = bulk_write_matches.value_of("id").unwrap();
+                        let filename = bulk_write_matches.value_of("filename").unwrap();
+                        let expiration = bulk_write_matches.value_of("expiration");
+                        let ttl = bulk_write_matches.value_of("time-to-live");
+                        let base64 = match bulk_write_matches.occurrences_of("release") {
+                            1 => true,
+                            _ => false,
+                        };
+                        commands::kv::write_bulk(id, filename, expiration, ttl, base64)?;
```
ashleymichal: expiration, ttl, and base64 are actually all set on individual objects in the JSON. We may want to provide flags for setting them for the whole set (especially for uploading a whole directory), but we'll have to decide preference order.

src/commands/kv/write_bulk.rs
```diff
@@ -0,0 +1,40 @@
...
+use cloudflare::workerskv::write_bulk::BulkWriteParams;
+use cloudflare::workerskv::write_bulk::KeyValuePair;
+
+use crate::terminal::message;
+
+pub fn write_bulk(namespace_id: &str, filename: &str, expiration: Option<&str>, ttl: Option<&str>, base64: bool) -> Result<(), failure::Error> {
```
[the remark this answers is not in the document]
gabbifish: ^ excellent catch

src/main.rs
gabbifish: I think it's worth keeping the consistency of this API usage by ensuring that folks using it know that these options apply to all things uploaded in a bulk upload action--if folks want to do this for individual key-value pairs (or files for static site hosting) they can use the individual `write` subcommand?

gabbifish: Hmm, now that I think about it, I'm okay with forcing folks to specify expiration and ttl in their json (adding them as commandline flags would imho get clunky and confusing to use for bulk). What do you think? (This is following our convo earlier today :)). 

We can still provide these as flags for individual writs!

src/main.rs
```diff
@@ -76,6 +77,39 @@ fn run() -> Result<(), failure::Error> {
...
+                            SubCommand::with_name("bulk")
+                                .about("upload multiple key-value pairs at once")
+                                .arg(
+                                    Arg::with_name("id")
+                                        .help("the id of your Workers KV namespace")
+                                        .index(1),
+                                )
+                                .arg(
+                                    Arg::with_name("path")
+                                    .help("the json file of key-value pairs to upload, in form [{\"key\":..., \"value\":...}\"...] OR the directory of files to upload.")
```
Electroid: The json and directory upload are two very distinct things, I wonder if this needs to be broken up into different argument names or even second-level command names. I'm also not sure using the bulk endpoint should be transparent to the user, what if it was just write?

```bash
wrangler kv write --pairs key=val
wrangler kv write -p key0=val0 keyN=valN
wrangler kv write -p '{"key0":"val0", "keyN":"valN"}'

wrangler kv write --directory dir/
```

src/main.rs
```diff
@@ -76,6 +77,39 @@ fn run() -> Result<(), failure::Error> {
...
+                        .arg(
+                            Arg::with_name("expiration")
+                            .short("e")
+                            .long("expiration")
+                            .takes_value(true)
+                            .value_name("SECONDS")
+                            .help("the time, measured in number of seconds since the UNIX epoch, at which the entries should expire"),
+                        )
+                        .arg(
+                            Arg::with_name("time-to-live")
```
Electroid: We should consider using one of many duration parsing [libraries](http://tailhook.github.io/humantime/humantime/struct.Duration.html) for this, unlike cache, KV typically has much longer expiration. We don't want people getting out their calculators.

By default, numbers are interpreted as seconds, but now I can do things like `24hr 30m`. Very convenient, and maybe there are other places in the codebase that could benefit from it.

src/commands/kv/write_bulk.rs
```diff
@@ -0,0 +1,111 @@
...
+            path_with_forward_slash.push("/");
+        }
+
+         path_with_forward_slash.push(component);
+    }
+
+     // if we have a non-utf8 path here, it will fail, but that's not realistically going to happen
+    let path = path_with_forward_slash
+        .to_str()
+        .expect("found a non-UTF-8 path");
```
Electroid: You'd be surprised, we've had customers run into this before, can we output the path that failed?

src/commands/kv/write_bulk.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    match response {
+        Ok(_success) => message::success(&format!("Success")),
+        Err(e) => super::print_error(e),
+    }
+
+    Ok(())
+}
+
+fn parse_directory(directory: &Path) -> Result<Vec<KeyValuePair>, failure::Error> {
```
Electroid: I foresee there will be common errors with uploading values that are over the `2MB` limit. How do we want to handle this? (actually, what does our API do right now? does the whole write fail?)

gabbifish: This functionality is being added in an upcoming commit, keep your eyes out!

src/main.rs
gabbifish: We're actually removing this from the commandline in favor of having folks specify this in their KV upload json file, but the comment still stands--
I think this is one case where providing functionality on top of the preexisting KV API can be nice (https://api.cloudflare.com/#workers-kv-namespace-write-multiple-key-value-pairs). While using an epoch time converter isn't the worst, it definitely is something that can be taken out of this upload process :)

src/main.rs
gabbifish: This is definitely a design decision still in flux! A lot of DB clis (redis for example) have separate commands for individual and bulk uploads, so @ashleymichal and I are now considering actually having two commands, like `write` and `write-bulk`. I still think that inferring the type of upload (kv pairs from a file or an entire directory) can be implicitly handled and decided by the type of file passed in to `write-bulk` (file or directory?) 

This design discussion still ongoing, so take what I'm saying right now with a grain of salt!

src/commands/kv/write_bulk.rs
gabbifish: We're going to fall back on API responses for this, at least for the time being.

src/commands/kv/write_bulk.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    // if we have a non-utf8 path here, it will fail, but that's not realistically going to happen
+    let path = path_with_forward_slash.to_str().expect(&format!(
+        "found a non-UTF-8 path, {:?}",
+        path_with_forward_slash
+    ));
+    let path_bytes = path.as_bytes();
+
+    // we use PATH_SEGMENT_ENCODE_SET since we're encoding paths, this will turn / into %2F,
+    // which is needed for the API call to put a `/` into the key.
```
ashleymichal: Steve's PR did not use the bulk upload, so he had to put the key into the request URL. Since we're using bulk upload, we don't have to percent encode our keys, because they are part of the JSON body and not the request URL.

src/main.rs
ashleymichal: we have a few ideas around where to separate these use cases, we'll be discussing next week with Steve, Ashley, etc.

src/main.rs
ashleymichal: i would advise users to generate this json, not write it by hand. that really makes this parameter make more sense as seconds.

gabbifish: ^ Agreed.

Issue #347, opened 2019-07-22
Add KV Command: Bulk Remove Key/Value Pair from Namespace


ashleymichal:
Related to #344 - we should also support delete

As a Workers KV user, I should be able to delete many key/value pairs on my KV namespace from Wrangler at once.

ashleymichal:
closed by #452

PR #452, opened 2019-08-19
Implementation of single and bulk deletion from KV

### LINKED ISSUE #344: Add KV Command: Remove Key/Value Pair from Namespace
As a Workers KV user, I should be able to remove a key/value pair from my KV namespace from Wrangler by typing

``` sh
$ wrangler kv rm myNamespace myKey1
```

gabbifish:
This PR closes #344 and #347. It implements single and bulk deletion of values from Workers KV.

This PR may potentially still be WIP because we still need to decide how exactly we want the workers subcommands to look like, especially between {namespace, individual, bulk} deletion; we can do this in this PR or on the feat-kv-commands branch.

So far, the commands are used as the following:
Individual delete:
```
$ wrangler kv delete-key 06779da6940b431db6e566b4846d64db matt
 Deleting key "matt"
 Success
```

For bulk delete (on a json blob): 
```
$ wrangler kv delete-bulk 06779da6940b431db6e566b4846d64db test-delete.json 
 Success
```

For bulk delete (on a directory):
```
$ wrangler kv delete-bulk 06779da6940b431db6e566b4846d64db abc
 Deleting a.txt...
 Deleting b%2Fc.txt...
 Success
```

This PR will fail to build until the necessary API changes are also merged into cloudflare-rs (cloudflare/cloudflare-rs#29), because it is dependent on these changes.

### DIFF
--- a/src/commands/kv/delete_bulk.rs
+++ b/src/commands/kv/delete_bulk.rs
@@ -1,0 +1,66 @@
+extern crate base64;
+
+use cloudflare::framework::apiclient::ApiClient;
+use walkdir::WalkDir;
+
+use std::fs;
+use std::fs::metadata;
+use std::path::Path;
+
+use cloudflare::endpoints::workerskv::delete_bulk::DeleteBulk;
+use failure::bail;
+
+use crate::terminal::message;
+
+pub fn delete_bulk(namespace_id: &str, filename: &Path) -> Result<(), failure::Error> {
+    let client = super::api_client()?;
+    let account_id = super::account_id()?;
+
+    // If the provided argument for delete_bulk is a json file, parse it
+    // and delete its listed keys. If the argument is a directory, delete key-value
+    // pairs where keys are the relative pathnames of files in the directory.
+    let mut data;
+    let keys: Result<Vec<String>, failure::Error> = match metadata(filename) {
+        Ok(ref file_type) if file_type.is_file() => {
+            data = fs::read_to_string(filename)?;
+            Ok(serde_json::from_str(&data)?)
+        }
+        Ok(ref file_type) if file_type.is_dir() => parse_directory(filename),
+        Ok(_file_type) => {
+            // any other file types (namely, symlinks)
+            bail!(
+                "{} should be a file or directory, but is a symlink",
+                filename.display()
+            )
+        }
+        Err(e) => bail!(e),
+    };
+
+    let response = client.request(&DeleteBulk {
+        account_identifier: &account_id,
+        namespace_identifier: namespace_id,
+        bulk_keys: keys?,
+    });
+
+    match response {
+        Ok(_success) => message::success("Success"),
+        Err(e) => super::print_error(e),
+    }
+
+    Ok(())
+}
+
+fn parse_directory(directory: &Path) -> Result<Vec<String>, failure::Error> {
+    let mut delete_vec: Vec<String> = Vec::new();
+    for entry in WalkDir::new(directory) {
+        let entry = entry.unwrap();
+        let path = entry.path();
+        if path.is_file() {
+            let key = super::generate_key(path, directory)?;
+
+            message::working(&format!("Deleting {}...", key.clone()));
+            delete_vec.push(key);
+        }
+    }
+    Ok(delete_vec)
+}
--- a/src/commands/kv/delete_key.rs
+++ b/src/commands/kv/delete_key.rs
@@ -1,0 +1,25 @@
+use cloudflare::endpoints::workerskv::delete_key::DeleteKey;
+use cloudflare::framework::apiclient::ApiClient;
+
+use crate::terminal::message;
+
+pub fn delete_key(id: &str, key: &str) -> Result<(), failure::Error> {
+    let client = super::api_client()?;
+    let account_id = super::account_id()?;
+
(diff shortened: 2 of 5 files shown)


src/commands/kv/delete_bulk.rs
```diff
@@ -0,0 +1,99 @@
...
+
+            message::working(&format!("Deleting {}...", key.clone()));
+            delete_vec.push(key);
+        }
+    }
+    Ok(delete_vec)
+}
+
+// todo(gabbi): When https://github.com/cloudflare/wrangler/pull/445 is merged, factor this
+// function out to mod.rs (it's used for both bulk write and delete)
```
ashleymichal: came here to say this 👍

src/commands/kv/delete_bulk.rs
```diff
@@ -0,0 +1,99 @@
...
+                filename.display()
+            )
+        }
+        Err(e) => bail!(e),
+    };
+
+    let response = client.request(&DeleteBulk {
+        account_identifier: &account_id,
+        namespace_identifier: namespace_id,
+        bulk_keys: keys?,
```
ashleymichal: nit: i think having `?` at variable declaration rather than in a struct constructor feels... more right? So at the end of ln39? idk, @steveklabnik or @ashleygwilliams may advise.

src/commands/kv/delete_bulk.rs
```diff
@@ -0,0 +1,99 @@
...
+    // if we have a non-utf8 path here, it will fail, but that's not realistically going to happen
+    let path = path_with_forward_slash.to_str().expect(&format!(
+        "found a non-UTF-8 path, {:?}",
+        path_with_forward_slash
+    ));
+    let path_bytes = path.as_bytes();
+
+    // we use PATH_SEGMENT_ENCODE_SET since we're encoding paths, this will turn / into %2F,
+    // which is needed for the API call to put a `/` into the key.
+    Ok(percent_encode(path_bytes, PATH_SEGMENT_ENCODE_SET).to_string())
```
ashleymichal: same as write; we don't need this part bc json.

src/commands/kv/delete_key.rs
```diff
@@ -0,0 +1,26 @@
...
+    let client = super::api_client()?;
+    let account_id = super::account_id()?;
+
+    let msg = format!("Deleting key \"{}\"", key);
+    message::working(&msg);
+
+    let response = client.request(&DeleteKey {
+        account_identifier: &account_id,
+        namespace_identifier: id,
+        key: key,
```
ashleymichal: if it doesn't happen in https://github.com/cloudflare/cloudflare-rs/pull/29, we need to percent encode here.

src/main.rs
```diff
@@ -77,6 +78,24 @@ fn run() -> Result<(), failure::Error> {
                 .subcommand(
                     SubCommand::with_name("list")
                 )
+                .subcommand(
+                    SubCommand::with_name("delete-key")
```
ashleymichal: will revisit subcommands with stakeholders, these are probably not it.

src/commands/kv/delete_bulk.rs
gabbifish: Nice catch! :)

src/commands/kv/delete_key.rs
gabbifish: I think the percent encoding logic should live in wrangler, since we'll be able to have more flexibility over it, plus cloudflare-rs (I think) is supposed to be a thin layer over the API--what do you think?

src/main.rs
gabbifish: they are mos def not it

src/commands/kv/delete_key.rs
gabbifish: resolution: percent encoding logic now lives in cloudflare-rs :)

src/commands/kv/delete_bulk.rs
```diff
@@ -0,0 +1,66 @@
...
+    let keys: Result<Vec<String>, failure::Error> = match metadata(filename) {
+        Ok(ref file_type) if file_type.is_file() => {
+            data = fs::read_to_string(filename)?;
+            Ok(serde_json::from_str(&data)?)
+        }
+        Ok(ref file_type) if file_type.is_dir() => parse_directory(filename),
+        Ok(_file_type) => {
+            // any other file types (namely, symlinks)
+            bail!(
+                "{} should be a file or directory, but is a symlink",
```
ashleymichal: nit: is symlink the only other type here? or should we interpolate `_file_type`?

gabbifish: Symlink is the only other type here! I have to use a catch-all case for remaining file_type values or else the rust compiler is unhappy and thinks I am not covering all branches. I could change the wording to get rid of `but is a symlink`, and just generally point out that the provided path is not for a file or directory. What do you think?

ashleymichal: hmm, i think don't worry about it for now. i am wondering whether following symlinks is advisable but i'm okay with not for now, that could turn into a rabbit hole with no significant benefits.

gabbifish: Coo, merging now!

PR #454, opened 2019-08-19
Add gets for single keys from Workers KV

### LINKED ISSUE #349: Add KV Command: Get Value for Key in Namespace
As a Workers KV user, I should be able to download the value set on a key in my KV namespace from Wrangler.

gabbifish:
This PR closes #349.

This GET behavior is used as such:

Success case
```
$ wrangler kv read-key 06779da6940b431db6e566b4846d64db gabbi
fisher
```
NOTE the lack of emoji in this output, in case someone wants to pipe the get output from wrangler into the stdin of another process.

Failure case
```
$ wrangler kv read-key 06779da6940b431db6e566b4846d63db gabbi
 Error 10013: namespace not found
 Run `wrangler kv list` to see your existing namespaces with IDs
```
^ There is the standard emoji logic here, it just doesn't show up in Ubuntu for now...

In a future PR, we'll use cloudflare-rs for this; its usage is unfortunately blocked due to https://github.com/cloudflare/cloudflare-rs/pull/27 (cloudflare-rs doesn't support non-json API responses, and the GET operation for Workers KV keys returns raw text on success).

### DIFF
--- a/src/commands/kv/get_key.rs
+++ b/src/commands/kv/get_key.rs
@@ -1,0 +1,38 @@
+// todo(gabbi): This file should use cloudflare-rs instead of our http::auth_client 
+// when https://github.com/cloudflare/cloudflare-rs/issues/26 is handled (this is 
+// because the GET key operation doesn't return json on success--just the raw
+// value).
+
+use cloudflare::framework::response::ApiFailure;
+
+use crate::settings::global_user::GlobalUser;
+use crate::settings::project::Project;
+use crate::http;
+
+pub fn get_key(project: &Project, user: &GlobalUser, id: &str, key: &str) -> Result<(), failure::Error> {
+    let api_endpoint = format!(
+        "https://api.cloudflare.com/client/v4/accounts/{}/storage/kv/namespaces/{}/values/{}",
+        project.account_id, id, key
+    );
+
+    let client = http::auth_client(user); 
+
+    let mut res = client
+    .get(&api_endpoint)
+    .send()?;
+
+    if res.status().is_success() {
+        let body_text = res.text()?;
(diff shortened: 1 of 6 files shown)


ashleymichal:
> NOTE the lack of emoji in this output, in case someone wants to pipe the get output from wrangler into the stdin of another process.

This is a solid point. We should look at whether more of these commands should be pure data output.

ashleymichal:
similar to `s/set/write` i think this should be `s/get/read`

src/commands/kv/get_key.rs
```diff
@@ -0,0 +1,41 @@
+// todo(gabbi): This file should use cloudflare-rs instead of our http::auth_client
+// when https://github.com/cloudflare/cloudflare-rs/issues/26 is handled (this is
+// because the GET key operation doesn't return json on success--just the raw
+// value).
+
+use cloudflare::framework::response::ApiFailure;
```
ashleymichal: oo nice!

src/commands/kv/get_key.rs
```diff
@@ -0,0 +1,41 @@
...
+
+pub fn get_key(
+    project: &Project,
+    user: &GlobalUser,
+    id: &str,
+    key: &str,
+) -> Result<(), failure::Error> {
+    let api_endpoint = format!(
+        "https://api.cloudflare.com/client/v4/accounts/{}/storage/kv/namespaces/{}/values/{}",
+        project.account_id, id, key
```
ashleymichal: consider url-encoding the key since it's part of the request url here

gabbifish: Handled! :)

PR #458, opened 2019-08-19
Add individual write subcommand for Workers KV

### LINKED ISSUE #343: Add KV Command: Write Key/Value Pair on Namespace
As a Workers KV user, I should be able to set a key/value pair on my KV namespace from Wrangler by typing

``` sh
$ wrangler kv write myNamespace myKey=”someStringValue” --exp 2d --ttl 60s
```

gabbifish:
This PR closes #343. 

The SET behavior is used the following way:

Success case (set string value):
```
$ wrangler kv set-key 06779da6940b431db6e566b4846d64db taylor swift --ttl 60
 Success
```

Success case (set value from file contents)
```
$ wrangler kv write-key 06779da6940b431db6e566b4846d64db my_file abc/a.txt --ttl 60 -f
 Success
```

Failure cases 
```
$ wrangler kv write-key 06779da6940b431db6e566b4846d63db --ttl 60 -f my_file abc/a.txt
 Error 10013: namespace not found
 Run `wrangler kv list` to see your existing namespaces with IDs
```
```
$ wrangler kv write-key 06779da6940b431db6e566b4846d64db --ttl 60 -f my_file abc/ab.txt
Error: No such file or directory (os error 2)
```
^ There is the standard emoji logic here, it just doesn't show up in Ubuntu for now...

In a future PR, we'll use cloudflare-rs for this; its usage is unfortunately blocked due to cloudflare/cloudflare-rs#27 (cloudflare-rs doesn't support non-json API responses, and the GET operation for Workers KV keys returns raw text on success).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2387,6 +2387,7 @@ dependencies = [
  "serde_json 1.0.40 (registry+https://github.com/rust-lang/crates.io-index)",
  "text_io 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "toml 0.5.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "url 2.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "uuid 0.7.4 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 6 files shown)


ashleymichal:
this command should be "write"

PR #460, opened 2019-08-21
Add Default trait to Project

ashleymichal:
This is a v. small refactor that is relevant to both work on Environments (https://github.com/cloudflare/wrangler/milestone/12, paging @EverlastingBugstopper ) and KV Additions (https://github.com/cloudflare/wrangler/milestone/10)

### DIFF
--- a/src/settings/project/mod.rs
+++ b/src/settings/project/mod.rs
@@ -16,7 +16,7 @@ use log::info;
 use config::{Config, Environment, File};
 use serde::{Deserialize, Serialize};
 
-#[derive(Clone, Debug, Deserialize, Serialize)]
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
 pub struct Project {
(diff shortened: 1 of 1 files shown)


src/settings/project/mod.rs
```diff
@@ -37,17 +37,8 @@ impl Project {
         project_type: ProjectType,
         init: bool,
     ) -> Result<Project, failure::Error> {
-        let project = Project {
-            name: name.clone(),
-            project_type: project_type.clone(),
-            private: Some(false),
```
xtuc: I guess the default of Option<bool> is None?

ashleymichal: mmm yes it is. I should likely update to use a more tailored default implementation than the one that can be derived. or make `new` define it like it does with project_type.

ashleymichal:
declining for now.

PR #461, opened 2019-08-21
[WIP][FEAT] add commands for httpie/ http-prompt

xortive:
This PR adds two new commands to wrangler for testing workers. They are conceptually similar to `wrangler preview`, but instead of opening a browser window, they allow you to test using CLI tools.

`wrangler http`, which calls [httpie](https://httpie.org/) with the necessary preview service arguments prefilled
`wrangler http-prompt` which starts a new [http-prompt](http://http-prompt.com/) session with the necessary arguments prefilled.

`wrangler preview` currently makes a test HTTP request before opening the preview in the browser, this PR opens the possibility deprecate that behavior in favor of the two new commands, so `preview` is dedicated for browser preview, and `http`/`http-prompt` are used for making HTTP requests to a worker from CLI.

These still need some code cleanup work, and there is some argument as to whether or not these commands belong in wrangler or not. They could easily be implemented as cargo-style subcommands, using `clap` and the same `process_builder.rs` file that cargo uses for transferring control to the subcommand process.

### DIFF
--- a/src/process_builder.rs
+++ b/src/process_builder.rs
@@ -1,0 +1,271 @@
+use std::collections::HashMap;
+use std::env;
+use std::ffi::{OsStr, OsString};
+use std::fmt;
+use std::path::Path;
+use std::process::{Command, Output, Stdio};
+
+use failure::{Fail, format_err};
+use jobserver::Client;
+use shell_escape::escape;
+
+/// A builder object for an external process, similar to `std::process::Command`.
+#[derive(Clone, Debug)]
+pub struct ProcessBuilder {
+    /// The program to execute.
+    program: OsString,
+    /// A list of arguments to pass to the program.
+    args: Vec<OsString>,
+    /// Any environment variables that should be set for the program.
+    env: HashMap<String, Option<OsString>>,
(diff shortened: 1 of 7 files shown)


src/process_builder.rs
```diff
@@ -0,0 +1,271 @@
...
+use std::path::Path;
+use std::process::{Command, Output, Stdio};
+
+use failure::{Fail, format_err};
+use jobserver::Client;
+use shell_escape::escape;
+
+/// A builder object for an external process, similar to `std::process::Command`.
+#[derive(Clone, Debug)]
+pub struct ProcessBuilder {
```
xortive: Borrowed this from cargo as it is a very nice cross-platform implementation of exec_replace. `std::process::Command` causes weird IO issues with http-prompt

gabbifish: preliminary comment: I think it's worth linking to https://github.com/rust-lang/cargo/blob/master/src/cargo/util/process_builder.rs just so folks can see this source for themselves as well :)

EverlastingBugstopper:
closing in favor of #845

PR #462, opened 2019-08-21
Add key listing functionality for workers kv subcommands

### LINKED ISSUE #345: Add KV Command: List Key/Value Pairs for Namespace
As a Workers KV user, I should be able to list all key/value pairs from my KV namespace from Wrangler by typing

``` sh
$ wrangler kv ls myNamespace
```

gabbifish:
This PR closes #345. A clarification: this does not list all KV pairs in a Workers KV instance, but instead, lists all keys and (if present) their expiration timestamps. 

Usage without prefix filter:
```
$ /wrangler kv list-keys 06779da6940b431db6e566b4846d64db
[{"name":"key1"},{"name":"key2"},{"name":"key3"}]
```

With prefix filter:
```
$ wrangler kv list-keys 06779da6940b431db6e566b4846d64db --prefix taylor
[{"name":"taylor"}]
```

I made sure that the logic handles pagination correctly for KV namespaces with >1000 keys :)
There are `todo` comments in my code about particular pieces of logic I'd like to write more elegantly, if possible. I'd love if you could toss ideas there if you have any.

Right now, I print out the results of every page as wrangler gets them (instead of storing them in memory and outputting them all at once). What do the reviewers think about this? I figured this was the best option because it wouldn't eat memory, but I'm curious what other folks think.

This PR will fail CI until https://github.com/cloudflare/cloudflare-rs/pull/30 is merged.


src/commands/kv/list_keys.rs
```diff
@@ -0,0 +1,82 @@
...
+        } else {
+            // Case where we still have pages to iterate through (a cursor is returned).
+            // Update cursor in request_params.params, and make another request to Workers KV API.
+            request_params.params.cursor = Some(cursor);
+
+            // todo(gabbi): Right now, I print out the results of every page
+            // as wrangler gets them (instead of storing them in memory and
+            // outputting them all at once). What do the reviewers think about this?
+            // I figured this was the best option because it wouldn't eat memory, but
+            // I'm curious what other folks think.
```
ashleymichal: fwiw i think this is the ideal base case; where i can see us wanting the "in memory" construct is in the "bucket" case, but i think we can refactor to that rather than changing this.

src/commands/kv/list_keys.rs
```diff
@@ -0,0 +1,82 @@
...
+            print_page(&result);
+            print!(","); // add comma between this set of json blobs and next set of json blobs.
+            response = client.request(&request_params);
+        }
+    }
+
+    Ok(())
+}
+
+fn get_cursor_from_result_info(result_info: Option<JsonValue>) -> String {
```
ashleymichal: maybe you return `Option<String>` here instead of `String`? That moves the `if cursor == ""` into this fn.

Also there is a `String.is_empty()` method that I believe is slightly more idiomatic than comparing to empty string.

src/commands/kv/list_keys.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    // Iterate over all pages until no pages of keys are left.
+    // This is detected when a returned cursor is an empty string.
+    // todo(gabbi): the code in this loop is the product of a looooong fight
+    // with the borrow checker. Please tell me if there's a neater way to write
+    // the logic below!
+    loop {
+        let (result, cursor) = match response {
+            Ok(ref success) => (
+                serde_json::to_string(&success.result.clone())?,
```
ashleymichal: rather than deserializing here and then slicing the result in `print_page`, perhaps let it be the `Vec` and iterate over that to print the objects?

gabbifish: I thought about doing just that :smile:! I wondered if it would be slower to iterate through the vec and serialize each entry instead of printing out a few serialized "batches", so to speak. I'd love to defer to @ashleygwilliams or @steveklabnik (someone who knows more about rust than me) to see which approach seems more efficient.

src/commands/kv/list_keys.rs
gabbifish: oooh nice call with `is_empty()!` let me experiment with the suggestion above.

gabbifish: refactored code with suggestion in this comment, lmk what you think!

src/commands/kv/list_keys.rs
ashleymichal: if i put on my dubs hat, i would be willing to bet that the answer would be somewhere along these lines:
The performance degradation from such a change would pale in comparison to the http requests we are already making in between prints. go for max readability and we can adjust from there if we find performance issues.

ashleymichal: also, if i'm understanding the code right, since we are using cloudflare rs, once we make the request, we are already bringing in the json and deserializing it to a `Vec<Key>` before this step, only to re-serialize it into a string and then slice it, which i believe makes a copy of the String. I think that all comes out to 6/half dozen compared to deserializing each Key and just printing it. This is my conjecture anyway.

src/commands/kv/list_keys.rs
```diff
@@ -0,0 +1,90 @@
...
+// the number of entries in each json blob is variable (so csv and tsv
+// representation won't make sense)
+pub fn list_keys(id: &str, prefix: Option<&str>) -> Result<(), failure::Error> {
+    let client = super::api_client()?;
+    let account_id = super::account_id()?;
+
+    let params = ListNamespaceKeysParams {
+        limit: None, // Defaults to 1000 (the maximum)
+        cursor: None,
+        prefix: prefix.map(str::to_string),
```
ashleymichal: coooool

src/commands/kv/list_keys.rs
```diff
@@ -0,0 +1,95 @@
...
+        account_identifier: &account_id,
+        namespace_identifier: id,
+        params: params,
+    };
+
+    let mut response = client.request(&request_params);
+
+    print!("["); // Open json list bracket
+
+    // used to track whether to put a glue "," before printing a json blob.
```
ashleymichal: remove old comment.

src/commands/kv/list_keys.rs
gabbifish: Benchmarked, and many individual serializes are not noticeably different from a few large serializes. We'll go with many individual serializes for readability and easier output formatting :)

src/commands/kv/list_keys.rs
```diff
@@ -0,0 +1,95 @@
...
+    // Iterate over all pages until no pages of keys are left.
+    // This is detected when a returned cursor is an empty string.
+    // todo(gabbi): the code in this loop is the product of a looooong fight
```
ashleymichal: can remove this comment :) i'm sure if folks find a reason to change it they will.

src/commands/kv/list_keys.rs
gabbifish:
```suggestion

```

src/commands/kv/list_keys.rs
gabbifish: will do!

src/commands/kv/list_keys.rs
```diff
@@ -0,0 +1,94 @@
...
+                print_page(result, true)?;
+                print!("]"); // Close json list bracket
+                break;
+            }
+            Some(_) => {
+                // Case where we still have pages to iterate through (a cursor is returned).
+                // Update cursor in request_params.params, and make another request to Workers KV API.
+                request_params.params.cursor = cursor;
+
+                // todo(gabbi): Right now, I print out the results of every page
```
ashleymichal: same; i think these are good to put in as github comments on your own PR.

PR #466, opened 2019-08-22
Gabbi/improve bulk error handling

### LINKED ISSUE #465: Batch bulk upload limits
As much as possible we should analyze payloads before sending them to prevent wrangler spinning on impossible loads for no reason. In the case of a bulk upload using a json file (the basic bulk write method), this means the following:

> The entire request size must be 100 megabytes or less.
Wrangler should check that the file size is less than 100MB, and bail with a message if it is not. Future optimizations would include streaming a file and chunking it appropriately, but this is really stupid hard with json.

> Body should be an array of up to 10,000 key-value pairs
If the first parameter is satisfied, Wrangler should batch requests to meet the 10,000 key limit.

(Bonus)
> (keys) The name may be at most 512 bytes
Assert that all keys are 512B maximum

> (values) A UTF-8 encoded string to be stored, up to 2 MB in length.
Assert that all values are 2MB maximum

https://api.cloudflare.com/#workers-kv-namespace-write-multiple-key-value-pairs

(separate issue to be filed for "bucket" behavior)

gabbifish:
This PR closes #465.

This PR also adds logic for handling and displaying HTTP status codes indicative of an error. This is especially useful when the API gateway returns errors before the request ever makes it to the KV API; for example, prior to this PR, uploads of payload > 100MB would silently error because wrangler looked for a KV api response that didn't exist. Instead, we should make sure wrangler handles cases like this and points out relevant information like `Err code 413--payload too large`.


src/commands/kv/mod.rs
```diff
@@ -61,6 +63,15 @@ fn print_error(e: ApiFailure) {
     }
 }
 
+// For handling cases where the API gateway returns errors via HTTP status codes
+// (no KV error code is given).
+fn give_status_code_context(status_code: StatusCode) {
```
ashleymichal: this should probably eventually be pulled up and out bc the same issue would come up on any request to any api endpoint; however there's no use doing that now.

PR #467, opened 2019-08-22
refactor(testability): pass project_dir from main.rs, add lib.rs

xortive:
This PR eliminates explicit and implicit use of `env::current_dir` from everywhere in wrangler, in favor of passing it to client code from main.rs. It also adds a `lib.rs` file that `main.rs` `use`s, to make wrangler more testable.

PR #468, opened 2019-08-22
Tests for live preview watcher

xortive:
Depends on #467, merge that first

this adds integration tests for just live preview's watcher. They are pretty hacky and slow due to the nature of testing such a thing, would love suggestions on making them less hacky.


ashleygwilliams:
sadly this PR is very out of date with master. sorry for the delay here and that we can't merge it :(

Issue #330, opened 2019-07-18
Overwrite Protection for Project Uploads


ashleymichal:
# Problem
One of the biggest problems we are trying to solve with KV support in Wrangler is that users unknowingly overwrite existing bindings (especially namespace bindings) when publishing their Worker with Wrangler. One part of fixing this is adding support for KV Namespace binding configuration ( https://github.com/cloudflare/wrangler/milestone/6 ) , but that still requires users to duplicate their config exactly in order to avoid screwing anything up. Ideally, Wrangler could alert a user to differences between their local project, and the project that already exists by the same name in the Workers ecosystem.

# History
Previously I have pushed back against implementing publish-time checks in Wrangler because Wrangler cannot guarantee anything about the state of the script on the server. (see #290 ) However, as of last quarter, the Script upload endpoint provides support for `if-match` and `if-none-match` headers in request, which provides the appropriate API for implementing the kinds of checks that will prevent users from unintentionally overwriting data. (This is indicated in a later comment from @jRiest on that same issue).

# Solution
The likely flow is as follows:

## Scenario 1: A New Wrangler Project
On the first attempt to publish a Worker, Wrangler should include an if-none-match  header in the upload request. The API will return success or failure based on whether a script by the same name already exists for the given user.

## Scenario 2: An existing Wrangler project
Once Wrangler has successfully uploaded a script for the first time, it should maintain an append-only list of etags locally, and use the most recently added one in upload requests as the if-match  header (unless the --force flag is invoked). The API will return success if the etag it currently has on file for the given name matches the etag sent in the header. 

### On success :
A successful upload response includes an etag key, which corresponds with the bundle hash of the newly created Worker. Wrangler should add this hash to a list locally and use it in subsequent upload requests.

### On error :
If the API finds an existing script by the same name, it will return a precondition failed error. When this happens, Wrangler can present the user with the error, and helpful follow up instructions.

## Empowering the user
The user can now:

### resolve conflicts:
- "pull" or "fetch" (strawman commands) the old script and bindings and compare them. This is the trickiest part of this and requires some careful design to avoid scope creep for an MVP; it is sort of a feature on its own, but would unlock not only overwrite protection but script import into wrangler project as well. This functionality would also probably tie in heavily with a Wrangler versioning feature implementation.
- the "pull/fetch" GET the script by name. should add the local etag (included in the headers for the GET /script/:name response) to its internal state (see below).
[...]

PR #469, opened 2019-08-25
Fixes #58 - Auto increment for default name

### LINKED ISSUE #58: default name should auto increment
default name is "worker". if you try to use the default more than once it should be worker-1, worker-2, etc

xprazak2:
Fixes #58 

Stores the name of the last default project
in global config and increments it
when user generates a project without a name.


xprazak2:
All tests are finally passing. Apologies it took me so long to make them green.

ashleymichal:
hey @xprazak2 !

Sorry this has been sitting on the stack for awhile. I reviewed the code, and actually I don't think we want to cache the directory name every time wrangler runs; rather, we want to have Wrangler respond to an "already exists" OS error by reading the contents of the current working directory, and adding a number if the default name already exists, similar to how filenames are updated in a downloads folder when you download the same asset twice.

So, if `worker` already exists in the CWD, create `worker2`; if `worker2` already exists, create `worker3`, etc.

We really try to minimize the amount of local state Wrangler has to keep track of to keep our footprint relatively small.

This does have a small disadvantage, wherein one could accidentally overwrite a worker because of naming collisions, but we would rather solve that with #330 .

xprazak2:
Thank you for your review, I have made changes as suggested - wrangler no longer keeps additional state and scans current dir for the next available name.

src/commands/generate/mod.rs
```diff
@@ -33,14 +35,80 @@ pub fn run_generate(name: &str, template: &str) -> Result<(), failure::Error> {
...
+        num = num.map(|val| val + 1);
+        new_name = construct_name(&bare_name, num);
+        if num.is_none() {
+            num = Some(1);
+        }
+    }
+    Ok(new_name)
+}
+
+fn read_current_dir() -> Result<Vec<OsString>, failure::Error> {
```
ashleymichal: i would also suggest passing the name in and only returning colliding directories.

ashleymichal:
hey @xprazak2 , just bumping this, sorry i forgot to mention you in my review last week. what do you think about these comments? if you have the bandwidth to implement that would be lovely, if not please let me know and i can pick this up and finish it up, you'll of course be credited in the changelog. thanks again for this PR, it is very  much appreciated!

xprazak2:
I simplified the logic for generating a new name. The project name is no longer scanned for digits at the end but `-\d+` is appended when we detect the dir with the same name already exists.

xprazak2:
@ashleymichal, any chance you could do a re-review please?

PR #471, opened 2019-08-26
Add API token support

### LINKED ISSUE #354: Add Support for API Tokens
API Keys and API Tokens both work when authenticating with Cloudflare API endpoints: https://api.cloudflare.com/#getting-started-requests.

This is a feature request to add support for using API Tokens, instead of having to use the Global API Key & Email.

gabbifish:
This closes #354.

API tokens can now be added to `wrangler config` with the following workflow:
```
$ wrangler config
Looking to use a Global API Key and email instead? Run "wrangler config --api-key". (Not Recommended)
Enter your API token:
none_of_your_beeswax_this_is_my_token
Success  

$ wrangler config --api-key
We don't recommend using your Global API Key! Please consider using an API Token instead. https://support.cloudflare.com/hc/en-us/articles/200167836-Managing-API-Tokens-and-Keys
Enter your email:
gabbi@cloudflare.com
Enter your API key:
none_of_your_beeswax
Success
```

Currently, a user is expected to enter an email/API key pair or API token.


src/settings/global_user.rs
```diff
@@ -10,7 +10,8 @@ use config::{Config, Environment, File};
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct GlobalUser {
     pub email: String,
-    pub api_key: String,
+    pub api_key: Option<String>,
+    pub api_token: Option<String>,
```
xtuc: Maybe it would simplify the logic a bit:
```rust
pub enum Authentification { ApiToken(String), ApiKey(String) };
pub struct GlobalUser {
  pub email: String,
  pub authentification: Authentification,
}
```

Then in `src/http.rs` you could:

```rust
    match &user.authentification {
         ApiKey(key) => headers.insert("X-Auth-Key", HeaderValue::from_str(&key).unwrap()),
         ApiToken(token) => headers.insert(
             "Authorization",
             HeaderValue::from_str(&format!("Bearer {}", &token)).unwrap(),
         ),
     };
```

src/main.rs
```diff
@@ -161,11 +161,24 @@ fn run() -> Result<(), failure::Error> {
...
+
+        println!("If using API token, enter API token (otherwise press Enter): ");
+        let mut api_token_str: String = read!("{}\n");
+        api_token_str.truncate(api_token_str.trim_end().len());
+        let mut api_token = None;
+        if !api_token_str.is_empty() {
+            api_token = Some(api_token_str);
+        }
+
+        commands::global_config(email, api_key, api_token)?;
```
xtuc: What if the user specifies both an API key and an API token? They will be send both in headers (unless you implement my suggestion  https://github.com/cloudflare/wrangler/pull/471/files#r318516334).
Maye we should restrict that?

garrettgalow: you should prefer api token. that is what we do in terraform

src/commands/config.rs
```diff
@@ -17,18 +17,37 @@ pub fn set_file_mode(file: &PathBuf) {
...
+        }
+    } else {
+        println!("Enter email: ");
+        let mut email_str: String = read!("{}\n");
+        email_str.truncate(email_str.trim_end().len());
+        if !email_str.is_empty() {
+            user.email = Some(email_str);
+        }
+
+        println!("Enter API key: ");
```
EverlastingBugstopper: nit: can we make this `Enter global API key: ` to differentiate them a little bit?

EverlastingBugstopper: possible idea: maybe print a warning saying you can use API tokens instead of global API keys? not sure if we need it or want it since it's in beta but perhaps would be useful?

src/commands/whoami.rs
```diff
@@ -2,11 +2,17 @@ use crate::settings::global_user::GlobalUser;
 use crate::terminal::{emoji, message};
 
 pub fn whoami(user: &GlobalUser) {
-    let msg = format!(
-        "{} You are logged with the email '{}'.",
-        emoji::WAVING,
-        user.email
-    );
+    // Only returns email if present (using email + API key for auth).
```
EverlastingBugstopper: what happens with `whoami` if people use api tokens? should probably ping API with token and pull down email

src/commands/whoami.rs
```diff
@@ -2,11 +2,17 @@ use crate::settings::global_user::GlobalUser;
...
+            let msg = format!(
+                "{} You are logged with the email '{}'.",
+                emoji::WAVING,
+                email
+            );
 
-    message::info(&msg);
+            message::info(&msg);
+        }
+        None => (),
```
EverlastingBugstopper: if we dont want to do api call for this release we should print out the api token instead of nothing for this command

src/main.rs
```diff
@@ -136,7 +136,14 @@ fn run() -> Result<(), failure::Error> {
                 .about(&*format!(
                     "{} Setup wrangler with your Cloudflare account",
                     emoji::SLEUTH
-                )),
+                ))
+                .arg(
+                    Arg::with_name("token")
+                        .help("use an API token for authentication. API tokens are still in beta")
```
EverlastingBugstopper: nit: use an API token instead of your email and global API key for authentication. maybe add a link to where they can set up api tokens in dashboard in help message

src/settings/global_user.rs
gabbifish: Two points here:
1. email is only used with API key, so this enum doesn't need to pair an email with an api_token
1. we're going to remove API key support in an upcoming release anyways so I don't think we need to introduce an enum of two values when the second value is going to get removed in the near future.

src/commands/whoami.rs
gabbifish: The api token is very sensitive information and doesn't say anything about the user...

src/commands/config.rs
gabbifish: sure!

src/commands/whoami.rs
gabbifish: i can do that! will take a few mins to set up all the api handling.

src/main.rs
gabbifish: that link depends on what the user's account is... so I'm not the biggest fan of providing it. How about we just tell users that they can issue a token via the dashboard?

src/commands/whoami.rs
```diff
@@ -2,11 +2,17 @@ use crate::settings::global_user::GlobalUser;
...
 
 pub fn whoami(user: &GlobalUser) {
-    let msg = format!(
-        "{} You are logged with the email '{}'.",
-        emoji::WAVING,
-        user.email
-    );
+    // Only returns email if present (using email + API key for auth).
+    match &user.email {
+        Some(email) => {
```
EverlastingBugstopper: this should probably be a pair of if lets that handle four cases:

email + api key
api token

email + api key + api token. 

unless we explicitly unset email + api keys when people run `wrangler config --token`.

gabbifish: it does explicitly unset.

src/commands/whoami.rs
gabbifish: poop. the endpoint for getting this info about a user does not support API tokens yet. I have started another branch https://github.com/cloudflare/wrangler/tree/gabbi/add-api-token-whoami that we can merge in once this endpoint supports API tokens.

src/settings/global_user.rs
```diff
@@ -22,9 +23,16 @@ impl GlobalUser {
 
 impl From<GlobalUser> for Credentials {
     fn from(user: GlobalUser) -> Credentials {
+        if let Some(token) = user.api_token {
+            return Credentials::UserAuthToken { token: token };
+        }
+
+        // fallback to email and global API key.
+        // If either of the fields below are empty, just substitute in an empty string
```
gabbifish:
```suggestion
        // If either of the fields below are None, just substitute in an empty string
```

tests/config.rs
```diff
@@ -23,7 +23,6 @@ fn generate_config_with(eol: &str) {
 
     write!(stdin, "email@example.com{}", eol).unwrap();
     write!(stdin, "apikeythisissecretandlong{}", eol).unwrap();
-    write!(stdin, "{}", eol).unwrap();
```
EverlastingBugstopper: just interested in why this was removed

gabbifish: this was a newline added when the old UX had folks fill out three fields when running `wrangler config` (the last field was API token, which was just an empty string when email + global API key combo was used)

README.md
```diff
@@ -83,14 +83,21 @@ $ wrangler publish
...
-  Configure your global Cloudflare user. This is an interactive command that will prompt you for your email and API key:
+  Configure your global Cloudflare user. This is an interactive command that will prompt you for your API token:
 
   ```bash
   wrangler config
+  Enter API token:
+  superlongapitoken
+  ```
+
+  You can also provide your email and global API key (this is not recommended for security reasons):
```
ashleymichal: nit: should we link here to the same resource we link to in the terminal warning?

README.md
```diff
@@ -83,14 +83,21 @@ $ wrangler publish
...
+  wrangler config --api-key
   Enter email:
   testuser@example.com
-  Enter api key:
-  ...
+  Enter global API key:
+  superlongapikey
   ```
 
   You can also [use environment variables](https://developers.cloudflare.com/workers/tooling/wrangler/configuration/) to configure these values.
```
ashleymichal: for later, and perhaps i should make an issue for it: we should be clear about precedence.

e.g. (strawman): if you have both a config file and env vars, env vars take precedence; if you have all three env vars defined, api token takes precedence.

src/commands/config/mod.rs
```diff
@@ -17,10 +17,40 @@ pub fn set_file_mode(file: &PathBuf) {
...
+        if !email_str.is_empty() {
+            user.email = Some(email_str);
+        }
+
+        println!("Enter global API key: ");
+        let mut api_key_str: String = read!("{}\n");
+        api_key_str.truncate(api_key_str.trim_end().len());
+        if !api_key_str.is_empty() {
+            user.api_key = Some(api_key_str);
+        }
```
EverlastingBugstopper: shouldn't this else be moved to the level of `if !api_key {} else {`, not if the API token string is empty? If the API token string is empty, I'd expect an error to be thrown (might as well check for length as well), but I wouldn't expect it to just continue to entering api token

src/commands/config/mod.rs
```diff
@@ -17,10 +17,40 @@ pub fn set_file_mode(file: &PathBuf) {
         .expect("could not set permissions on file");
 }
 
-pub fn global_config(email: String, api_key: String) -> Result<(), failure::Error> {
-    let s = GlobalUser { email, api_key };
+pub fn global_config(api_key: bool) -> Result<(), failure::Error> {
+    let mut user = GlobalUser {
```
ashleymichal: i assume this is where you were alluding to an enum refactor? i agree with that; there should be two constructors for this type. is there an issue?

src/commands/preview/upload.rs
```diff
@@ -78,7 +78,7 @@ pub fn upload(
                 "You haven't run `wrangler config`. Running preview without authentication",
             );
             message::help(
-                "Run `wrangler config` or set $CF_API_KEY and $CF_EMAIL to configure your user.",
+                "Run `wrangler config` or set $CF_EMAIL and $CF_API_KEY/$CF_API_TOKEN to configure your user.",
```
EverlastingBugstopper: I find this a bit confusing - can we just promote `$CF_API_TOKEN` here?

src/commands/whoami/mod.rs
```diff
@@ -1,12 +1,20 @@
 use crate::settings::global_user::GlobalUser;
 use crate::terminal::{emoji, message};
 
-pub fn whoami(user: &GlobalUser) {
+pub fn whoami(user: &GlobalUser) -> Result<(), failure::Error> {
+    // If using email + API key for auth, simply prints out email from config file.
+    let email: String = match &user.email {
+        Some(email) => email.to_string(),
+        None => failure::bail!(
+            "At the moment, Wrangler cannot get user information for users using API tokens"
```
EverlastingBugstopper: 😭

src/commands/preview/upload.rs
gabbifish: yes!

src/commands/whoami/mod.rs
gabbifish: ikr :( API team will hopefully implement this soon.

src/commands/config/mod.rs
```diff
@@ -17,10 +17,40 @@ pub fn set_file_mode(file: &PathBuf) {
...
+        email: None,
+        api_key: None,
+        api_token: None,
+    };
 
-    let toml = toml::to_string(&s)?;
+    if !api_key {
+        // Default: use API token.
+        message::info("Looking to use a Global API Key and email instead? Run \"wrangler config --api-key\". (Not Recommended)");
```
ashleymichal: nit: i don't think the presence of this message makes sense. i think worst case it is confusing, best case it is noise.

src/http.rs
```diff
@@ -38,12 +38,35 @@ pub fn client(feature: Option<&str>) -> Client {
...
 
     builder()
-        .default_headers(headers)
+        .default_headers(auth_headers.to_owned())
         .redirect(RedirectPolicy::none())
         .build()
         .expect("could not create authenticated http client")
 }
+
+fn add_auth_headers<'a>(headers: &'a mut HeaderMap, user: &GlobalUser) -> &'a HeaderMap {
```
EverlastingBugstopper: i understand that `<'a>` is a lifetime - could you explain a bit why they're needed here and what they're doing?

src/commands/config/mod.rs
gabbifish: ah, so our baseline was that an empty string would not be serialized and would trigger the default API authentication failure message. This is preserving the preexisting behavior, but we can revisit this and think about adding additional error messages. We can do this after the RC is cut.

src/main.rs
```diff
@@ -478,7 +478,7 @@ fn run() -> Result<(), failure::Error> {
         log::info!("Getting User settings");
         let user = settings::global_user::GlobalUser::new()?;
 
-        commands::whoami(&user);
+        commands::whoami(&user)?;
```
EverlastingBugstopper: beautiful

src/commands/config/mod.rs
```diff
@@ -17,10 +17,40 @@ pub fn set_file_mode(file: &PathBuf) {
...
+        println!("Enter API token: ");
+        let mut api_token_str: String = read!("{}\n");
```
ashleymichal: it feels...wrong to put all this interactivity into a constructor, when originally we fed arguments to this function. i think it makes more sense to create a second function for token and call it from main, and to get the arguments from main as well.

src/commands/kv/mod.rs
```diff
@@ -72,6 +72,7 @@ fn api_client(user: &GlobalUser) -> Result<HttpApiClient, failure::Error> {
             // This is useful for bulk upload operations.
             http_timeout: Duration::from_secs(5 * 60),
         },
+        Environment::Production,
```
ashleymichal: i don't understand this change.

src/http.rs
gabbifish: ah! we want to edit the preexisting headers struct instead of making a clone of it (in general this is just better/cleaner memory-wise). This did remind me though, we don't need to even return headers here; I've removed this return.

src/commands/preview/upload.rs
gabbifish:
```suggestion
                "Run `wrangler config` or set either of $CF_API_TOKEN or ($CF_EMAIL, $CF_API_KEY) to configure your user.",
```

src/commands/preview/upload.rs
```diff
@@ -78,7 +78,7 @@ pub fn upload(
                 "You haven't run `wrangler config`. Running preview without authentication",
             );
             message::help(
-                "Run `wrangler config` or set $CF_API_KEY and $CF_EMAIL to configure your user.",
+                "Run `wrangler config` or set either of $CF_API_TOKEN or ($CF_EMAIL, $CF_API_KEY) to configure your user.",
```
ashleymichal:
```suggestion
                "Run `wrangler config` or set either $CF_API_TOKEN or ($CF_EMAIL, $CF_API_KEY) to configure your user.",
```

src/commands/whoami/mod.rs
```diff
@@ -1,12 +1,20 @@
...
+        ),
+    };
+
     let msg = format!(
-        "{} You are logged in with the email '{}'.",
+        "{} You are logged with the email '{}'.",
```
ashleymichal: i think the `in` was good?
```suggestion
        "{} You are logged in with the email '{}'.",
```

src/main.rs
```diff
@@ -391,15 +397,9 @@ fn run() -> Result<(), failure::Error> {
 
     let config_path = Path::new("./wrangler.toml");
 
-    if let Some(_matches) = matches.subcommand_matches("config") {
-        println!("Enter email: ");
-        let mut email: String = read!("{}\n");
```
ashleymichal: yah i think all the interactivity should stay up here.

src/commands/kv/mod.rs
gabbifish: this is part of the new cloudflare-rs version :)

src/commands/config/mod.rs
gabbifish: this was on the PRD--i feel like we can shunt this to after the RC is cut?

src/commands/config/mod.rs
gabbifish: yup! i can file an issue for the refactor.

src/main.rs
gabbifish: yup this is an easy fix! will do.

README.md
gabbifish: can do, easy fix :)

src/commands/config/mod.rs
ashleymichal: yah is a nit

src/commands/kv/mod.rs
ashleymichal: but what does it do?

src/commands/config/mod.rs
gabbifish: yeah, that'll be a small fix!

src/commands/kv/mod.rs
gabbifish: it uses api.cloudflare.com instead of staging.api.cloudflare.com--the cloudflare-rs team introduced a new staging environment you can use to test your code on staging API endpoints with this client library.

ashleymichal: cool! i'll ask more questions about it over there.

README.md
gabbifish: for sure! I think we'll discuss this behavior in next week's wrangler maintainer meeting and make sure we 100% want it before detailing it in the readme.

ashleymichal: yah again is strawman suggestion, but the key point is to document the decision. i can make an issue.

ashleymichal: linked: https://github.com/cloudflare/wrangler/issues/834

PR #485, opened 2019-08-28
Remove logic for uploading directories from bulk

### LINKED ISSUE #459: Separate "bucket" use case from JSON uploads
Currently we have two ways to handle KV bulk uploads: 

1. You can use the API raw and build your own JSON file with custom keys and values (we'll call this a standard `write`).

2. We also have a model where we walk a directory and generate keys on your behalf, effectively simulating a "bucket" that you access by path as if it was just a remote copy of your local assets directory.

Because the latter is really a specialized use case of the former, it seems more appropriate to make it its own class of commands; as a strawman, `wrangler bucket`, where you can either `wrangler bucket push` or `wrangler bucket pull` to sync the directory up or down as appropriate.

ashleymichal:
Closes #459 

We will be re-implementing this work for the KV Additions milestone, but under a separate subcommand. This PR will assert that `wrangler kv:bulk put` and `wrangler kv:bulk delete` specify a properly formatted JSON file, rather than matching on json file OR directory.

### DIFF
--- a/src/commands/kv/delete_bulk.rs
+++ b/src/commands/kv/delete_bulk.rs
@@ -1,6 +1,7 @@
 extern crate base64;
 
 use cloudflare::framework::apiclient::ApiClient;
+use walkdir::WalkDir;
 
 use std::fs;
 use std::fs::metadata;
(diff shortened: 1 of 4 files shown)


gabbifish:
I took a look at this and thought about how to future-proof this for our bucket implementation. I'm thinking that we could even re-write the bulk signatures from
```rust
pub fn delete_bulk(namespace_id: &str, filename: &Path) -> Result<(), failure::Error> {
```
to 
```rust
pub fn delete_bulk(namespace_id: &str, keys: Result<Vec<String>>) -> Result<(), failure::Error> {
```
We can then create wrapper functions around `delete_bulk` (like `delete_json` and `delete_bucket`) that are responsible for validating the filetype passed in (e.g. `delete_bulk` requires a path to a json file, `delete_bucket` requires a path to a directory). These wrapper functions are responsible for the parsing and formatting of the `keys` (for bulk delete) and `key_value_pairs` (for bulk write) that are passed into `delete_bulk` and `write_bulk`, respectively.

src/commands/kv/delete_bulk.rs
```diff
@@ -14,32 +13,23 @@ use crate::terminal::message;
...
         }
-        Ok(ref file_type) if file_type.is_dir() => parse_directory(filename),
-        Ok(_file_type) => {
-            // any other file types (namely, symlinks)
-            bail!(
-                "{} should be a file or directory, but is a symlink",
-                filename.display()
-            )
-        }
+        Ok(_) => bail!("{} should be a JSON file, but is not", filename.display()),
```
ashleymichal: we may want to smooth this and make it tell a user not only that it should be json but what the schema is; however this is mergable.

src/commands/kv/write_bulk.rs
```diff
@@ -15,33 +14,24 @@ use crate::terminal::message;
...
         }
-        Ok(ref file_type) if file_type.is_dir() => parse_directory(filename),
-        Ok(_file_type) => {
-            // any other file types (namely, symlinks)
-            bail!(
-                "Cannot upload a file that is a symlink: {}",
-                filename.display()
-            )
-        }
+        Ok(_) => bail!("{} should be a JSON file, but is not", filename.display()),
```
ashleymichal: same

src/commands/kv/delete_bulk.rs
gabbifish: Nice catch :)

src/commands/kv/write_bulk.rs
gabbifish: Nice catch :)

src/commands/kv/delete_bulk.rs
gabbifish: This might be better to put in the docs, I think? cuz the format for this can get quite long...
```
[{"key": "<KEY>", "value": "<VALUE>", "expiration": "<OPTIONAL_SECONDS>", "expiration_ttl": "<OPTIONAL_SECONDS>", "base64": "<OPTIONAL_BOOL>"}]
```

Issue #477, opened 2019-08-27
use target: webworker


xtuc:
We should force the `target: webworker` configuration in webpack, it will avoid webpack's runtime to use `window` or `document`.

lolcoolkat:
@xtuc is there any way we can get a minor version push to npm to include this update? :cry:

xtuc:
You can set `target: "webworker"` in your webpack configuration, if you have one as a workaround.

EverlastingBugstopper:
You can also clone this repository and run `cargo install --force --debug --path .` to get the latest version. This does not come with many QA guarantees though, we try to only merge things into master that we are confident with, but we do more rigorous testing on releases.

PR #490, opened 2019-08-30
feat: use target webworker

xtuc:
Fixes https://github.com/cloudflare/wrangler/issues/477. We want to
configure webpack to emit only webworker compatible code.

### DIFF
--- a/tests/build.rs
+++ b/tests/build.rs
@@ -171,6 +171,49 @@ fn it_builds_with_webpack_wast() {
     cleanup(fixture);
 }
 
+#[test]
+fn it_fails_with_webpack_target_web() {
+    let fixture = "webpack_target_web";
+    create_temporary_copy(fixture);
+
+    webpack_config(
+        fixture,
+        r#"{
+          entry: "./index.js",
+          target: "node",
+        }"#,
+    );
+    settings! {fixture, r#"
+        type = "webpack"
+    "#};
+
+    build_fails_with(
+        fixture,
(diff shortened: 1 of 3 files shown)


wranglerjs/index.js
```diff
@@ -28,6 +28,15 @@ if (args["no-webpack-config"] === "1") {
   config = require(join(process.cwd(), args["webpack-config"]));
 }
 
+if (config.target !== undefined && config.target !== "webworker") {
+  throw error(
+    "Building a Cloudflare Worker with target " + JSON.stringify(config.target) +
```
EverlastingBugstopper: Can we use string literals here? Also putting a stringified JSON in the middle of an error message seems... interesting. Do you have an example of what this would output?

xtuc: It's adding quotes arround string, normalize priting functions, NaN, undefined, etc.

lolcoolkat:
LGTM <3

tests/build.rs
```diff
@@ -129,6 +129,49 @@ fn it_builds_with_webpack_wast() {
...
+
+    build_fails_with(
+        fixture,
+        "Building a Cloudflare Worker with target \"node\" is not supported",
+    );
+    cleanup(fixture);
+}
+
+#[test]
+fn it_fails_with_webpack_target_webworker() {
```
ashleymichal: misleading test function name

xtuc: right, it shouldn't fail

PR #501, opened 2019-08-30
Gabbi/update subcommand invocations

### LINKED ISSUE #495: Update all Subcommands per design
Also add descriptions

https://wiki.cfops.it/display/EW/Add+Wrangler+commands+for+KV+operations (V3)

gabbifish:
This PR closes #495. 

A note: This PR still treats namespace-id arg as required, because the default env logic (#220) is not complete yet.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -49,139 +49,198 @@ fn run() -> Result<(), failure::Error> {
         .setting(AppSettings::ArgRequiredElseHelp)
         .setting(AppSettings::DeriveDisplayOrder)
         .subcommand(
-            SubCommand::with_name("kv")
+            SubCommand::with_name("kv:namespace")
                 .about(&*format!(
-                    "{} Interact with your Workers KV Store",
+                    "{} Interact with your Workers KV Namespaces",
                     emoji::KV
                 ))
                 .subcommand(
                     SubCommand::with_name("create")
+                        .about("Create a new namespace")
                         .arg(
                             Arg::with_name("title")
+                            .help("The name for your new namespace")
                             .required(true)
                         )
                 )
                 .subcommand(
                     SubCommand::with_name("delete")
+                        .about("Delete namespace")
                         .arg(
                             Arg::with_name("namespace-id")
+                            .help("The ID of the namespace this action applies to")
                             .required(true)
                         )
                 )
                 .subcommand(
                     SubCommand::with_name("rename")
+                        .about("Rename a namespace")
                         .arg(
                             Arg::with_name("namespace-id")
+                            .help("The ID of the namespace this action applies to")
                             .required(true)
                         )
                         .arg(
                             Arg::with_name("title")
+                            .help("New title for the namespace")
                             .required(true)
                         )
                 )
                 .subcommand(
                     SubCommand::with_name("list")
+                        .about("List all namespaces on your Cloudflare account")
                 )
+        )
+            .subcommand(SubCommand::with_name("kv:key")
+                .about(&*format!(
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -49,139 +49,198 @@ fn run() -> Result<(), failure::Error> {
...
+                            .help("Number of seconds since the UNIX epoch, indicating when the key-value pair should expire")
                             .short("e")
                             .long("expiration")
                             .takes_value(true)
                             .value_name("SECONDS")
-                            .help("the time, measured in number of seconds since the UNIX epoch, at which the entries should expire"),
                         )
                         .arg(
                             Arg::with_name("expiration-ttl")
+                            .help("Number of seconds for which the entries should be visible before they expire. At least 60")
```
ashleymichal: should we include the fact that this takes precedence over `expiration`?

src/main.rs
```diff
@@ -49,139 +49,198 @@ fn run() -> Result<(), failure::Error> {
...
+                            Arg::with_name("key")
+                            .help("Key whose value to get")
                             .required(true)
-                            .index(2),
                         )
                 )
                 .subcommand(
-                    SubCommand::with_name("delete-key")
+                    SubCommand::with_name("delete")
+                        .about("Delete a key (and its value) from a namespace")
```
ashleymichal: i think the parens are unnecessary
```suggestion
                        .about("Delete a key and its value from a namespace")
```

src/main.rs
```diff
@@ -49,139 +49,198 @@ fn run() -> Result<(), failure::Error> {
...
                         .arg(
                             Arg::with_name("namespace-id")
+                            .help("The ID of the namespace this action applies to")
                             .required(true)
+                            // .short("n")
```
ashleymichal: i generally hate committing commented code to master... 😬

src/main.rs
```diff
@@ -49,139 +49,198 @@ fn run() -> Result<(), failure::Error> {
...
                         .arg(
                             Arg::with_name("key")
+                            .help("Key whose value to delete")
                             .required(true)
                         )
                 )
                 .subcommand(
-                    SubCommand::with_name("delete-bulk")
+                    SubCommand::with_name("list")
+                        .about("List all keys in a namespace. Produces JSON output")
```
ashleymichal: should we also print namespace ids as json to start and introduce table printing for both later as a flag? like `--table`?

src/main.rs
```diff
@@ -49,139 +49,198 @@ fn run() -> Result<(), failure::Error> {
...
                             Arg::with_name("path")
+                            .help("the JSON file of key-value pairs to upload, in form [{\"key\":..., \"value\":...}\"...]")
                             .required(true)
+                            .index(2),
                         )
                 )
                 .subcommand(
-                    SubCommand::with_name("list-keys")
+                    SubCommand::with_name("delete")
+                        .about("Delete multiple keys (and their values) from a namespace")
```
ashleymichal: same
```suggestion
                        .about("Delete multiple keys and their values from a namespace")
```

src/main.rs
gabbifish: Yes!

src/main.rs
gabbifish: ugh me too. we can remove from the feature branch if we (for whatever reason) don't land this before the 1.3.0 release date.

src/main.rs
gabbifish: We could! I still am hesitant about printing tables, given the a) dynamic number of items in each entry and b) how big this output is going to get. I think folks, at most, would want a csv/tsv print format on top of json printing.

PR #405, opened 2019-08-06
Feat kv commands

ashleymichal:
Closes #339. Comprises complete support for KV commands in Wrangler, should close https://github.com/cloudflare/wrangler/milestone/7 also.

keep this branch up to date with master using ui merges rather than rebasing.

PR #507, opened 2019-09-03
Gabbi/url encode

### LINKED ISSUE #504: Make sure to URL encode all keys
This applies to singleton put and get operations. Without this, puts and gets with keys that have spaces will fail.

gabbifish:
This PR closes #504. Since this change also impacts how KV command methods import the `url_encode_key()` function from the mod.rs for the kv package, it also implements the change suggested here: https://github.com/cloudflare/wrangler/pull/405#discussion_r318173571

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -119,20 +119,20 @@ fn run() -> Result<(), failure::Error> {
                             .required(true)
                         )
                         .arg(
-                            Arg::with_name("expiration")
-                            .help("Number of seconds since the UNIX epoch, indicating when the key-value pair should expire")
-                            .short("e")
-                            .long("expiration")
-                            .takes_value(true)
-                            .value_name("SECONDS")
-                        )
-                        .arg(
                             Arg::with_name("expiration-ttl")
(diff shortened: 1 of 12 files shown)


src/main.rs
```diff
@@ -120,7 +120,7 @@ fn run() -> Result<(), failure::Error> {
                         )
                         .arg(
                             Arg::with_name("expiration-ttl")
-                            .help("Number of seconds for which the entries should be visible before they expire. At least 60. Takes precedence over 'expiration' option")
+                            .help("Number of seconds for which the entries should be visible before they expire. At least 60")
```
ashleymichal: did you mean to remove this?

gabbifish: Oop! That is an accident. Fixing.

gabbifish:
```suggestion
                            .help("Number of seconds for which the entries should be visible before they expire. At least 60. Takes precedence over 'expiration' option")
```

EverlastingBugstopper: `. At least 60.` --> `. Must be at least 60 seconds.`

Issue #474, opened 2019-08-27
Allow deploying to zoneless with environments


EverlastingBugstopper:
### How do we specify an environment for a worker that will be deployed to workers.dev?

Here are a couple of ideas for when the behavior of `wrangler publish --environment dev` is intended to push a worker to workers.dev:

**1) zone_id and route are excluded**
```toml
[env.dev]
name = "wrangler-dev"
type = "webpack"
account_id = "e04d3d27a6829775c8ef937145d7f3ec"
```

**2) route is workers.dev (or includes it)**
```toml
[env.dev]
name = "wrangler-dev"
type = "webpack"
account_id = "e04d3d27a6829775c8ef937145d7f3ec"
route = "workers.dev"
```

**3) we specify that this environment is for workers.dev with a to-be-named boolean value**
```toml
[env.dev]
name = "wrangler-dev"
type = "webpack"
account_id = "e04d3d27a6829775c8ef937145d7f3ec"
somethingworkersdotdev = true
```

or something else entirely!

My thoughts are that 1 is a potential footgun because you could just forget some fields in your wrangler.toml and not expect your worker to be published to workers.dev
2 seems hacky
3 feels bad but maybe the best. I'm not sure what we should call the boolean in the toml.

Issue #442, opened 2019-08-15
Specifying the `default` environment


ashleymichal:
Suggestion: add a `default` boolean key that indicates which environment to use when no environment is passed to `preview` or `publish`.

e.g.
``` toml
[env.prod]
# ...
[env.staging]
default = true
# ...
```
would cause `$ wrangler publish` to use the `staging` environment. To publish to prod, the user would have to specifically pass the env, as in `$ wrangler publish --env=prod` This can help the ergonomics of a development cycle by preventing users from accidentally publishing to "prod" when they don't intend to. 

Alternative:
set a default for `preview` and `publish` rather than making default a bool.

``` toml
[env.prod]
# ...
[env.staging]
default = "publish"
# ...
[env.dev]
default = "preview"
# ...
```
would cause `$ wrangler publish` to use the `staging` environment, and `$ wrangler preview` to use the `dev` environment. This may or may not really help a thing, but it is rather common to have these three environments, and these are the two ways one might "push" things; in the case of workers that use KV, for example, you need to have your assets stored in KV for your worker to access them even from the preview service, so defining a 'dev' namespace for each developer would be a must-have for local previewing to work ergonomically.

ashleygwilliams:
[default]
"preview" = prod
"preview" = staging
publish" = staging

EverlastingBugstopper:
@ashleygwilliams I'm super confused by that comment, could you elaborate?

ashleymichal:
@EverlastingBugstopper updated ashley's comment with context from the meeting where we came up with it.

EverlastingBugstopper:
Makes sense! I like it 😄

EverlastingBugstopper:
Instead of this design, we are moving forward with a new design that defines the default behavior at the top level of `wrangler.toml.`

PR #508, opened 2019-09-03
New Environments Design

### LINKED ISSUE #481: Create environment toml with wrangler generate

EverlastingBugstopper:
Parse .toml into `Manifest` with `Environments` - these are dummy structs, and are only used to parse the toml. 

Rename `Project` to `Target` - this is what is actually used.
Handling zoneless/zoned deploys is now in `Manifest::get_target`

Deprecate `private`
Generate includes `workers_dot_dev`

Fixes #481, #474, #442

### DIFF
--- a/src/settings/project/mod.rs
+++ b/src/settings/project/mod.rs
@@ -127,18 +127,15 @@ impl Manifest {
             },
             // top level (legacy)
             None => {
-                match release {
-                    true => {
-                        message::warn(deprecate_warning);
-                        false // --release means not workers.dev
-                    }
-                    false => {
-                        match self.workers_dot_dev {
-                            Some(wdd) => wdd,
-                            None => {
-                                message::warn(deprecate_warning);
-                                true // no --release means workers.dev
-                            }
+                if release {
+                    message::warn(deprecate_warning);
+                    false // --release means not workers.dev
+                } else {
+                    match self.workers_dot_dev {
+                        Some(wdd) => wdd,
+                        None => {
+                            message::warn(deprecate_warning);
+                            true // no --release means workers.dev
                         }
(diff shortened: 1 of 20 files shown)


src/settings/project/mod.rs
```diff
@@ -8,50 +8,238 @@ use crate::terminal::emoji;
...
 
 use log::info;
 
-use config::{Config, Environment, File, Value};
+use config::{Config, File};
 use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
-pub struct Project {
+pub struct Target {
```
xtuc: What's the difference between Target and Environment?

EverlastingBugstopper: `Environment` is a dummy struct that the toml is parsed into, and `Target` is constructed from merging the base `Manifest` with the parsed `Environment`

ashleymichal:
QA feedback thus far:

* requiring top level name when individual names are all given might be annoying; but perhaps not as annoying as it feels off the bat.
* appears to break environment variable config for account id, which I was relying on.
* should improve this error messaging:
``` sh
$ wrangler-dev publish —env=jekyll
Error: Your wrangler.toml is missing the fields [“zone_id”, “route”] which are required to publish to a route!
```
This happens when there is no zone_id, route_ or workersdotdev set
* I would expect to be able to use `workersdotdev = true` at the top level to deploy anything to workers.dev by default. this may cause some implementation hairiness in envs where you have a zoned deploy, but i think it'll work?

EverlastingBugstopper:
`workers_dot_dev` can now be specified at the top and inherited through your environments

PR #511, opened 2019-09-03
Gabbi/add interactive delete

### LINKED ISSUE #410: KV delete operations should confirm interactively
Destructive actions such as "delete key" or "delete namespace" should ask the user "are you sure?" before completing. Include a `force` or `yes` flag to facilitate automation.

gabbifish:
This PR closes #410. 

It adds interactive delete checking to 
1. `kv:namespace delete`
2. `kv:key delete`
3. `kv:bulk delete`

Example usage:
```
$ wrangler kv:key delete 06779da6940b431db6e566b4846d64db "gabbi fisher"
Are you sure you want to delete key "gabbi fisher"? [y/n]
yes
 Deleting key "gabbi fisher"
 Success
```
If someone wanted to automate deletion without going through the interactive workflow:
```
$ echo "yes" | wrangler kv:key delete 06779da6940b431db6e566b4846d64db "gabbi fisher"
Are you sure you want to delete key "gabbi fisher"? [y/n]
 Deleting key "gabbi fisher"
 Success
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -71,6 +71,13 @@ fn run() -> Result<(), failure::Error> {
                             .help("The ID of the namespace this action applies to")
                             .required(true)
                         )
+                        .arg(
+                            Arg::with_name("force")
+                            .help("Force delete key-value pair")
+                            .short("f")
+                            .long("force")
+                            .takes_value(false)
+                        )
                 )
                 .subcommand(
                     SubCommand::with_name("rename")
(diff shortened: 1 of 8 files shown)


src/main.rs
```diff
@@ -71,6 +71,13 @@ fn run() -> Result<(), failure::Error> {
                             .help("The ID of the namespace this action applies to")
                             .required(true)
                         )
+                        .arg(
+                            Arg::with_name("force")
```
gabbifish: Question: there is also a `--filename / -f` flag for `kv:key put`. Do we want to make this filename flag different as to avoid confusion with the force flag used by all delete subcommands?

ashleymichal: should we change it to `--path | -p`?

ashleymichal:
> It also adds --force flags to all of these subcommands so that the interactive delete flow can be skipped

I didn't understand the functionality at the time so failed to update the issue, but the request from product was the ability to pipe `yes` to the command.

gabbifish:
@ashleymichal Got rid of --force flag in favor of piping yes in :)

PR #514, opened 2019-09-03
Add bucket::upload, bucket::delete

### LINKED ISSUE #512: Add subcommand `kv:bucket upload`
This command should take as an argument the path to a directory of static assets to be uploaded to Workers KV, keyed on path relative to the root of same directory.

### LINKED ISSUE #513: Add subcommand `kv:bucket delete`
This command should take as an argument the path to a directory of static assets to be deleted from Workers KV, keyed on path relative to the root of same directory.

ashleymichal:
Closes #512 , Closes #513

### DIFF
--- a/src/commands/kv/bucket.rs
+++ b/src/commands/kv/bucket.rs
@@ -19,7 +19,7 @@ pub fn upload(namespace_id: &str, filename: &Path) -> Result<(), failure::Error>
     let pairs: Result<Vec<KeyValuePair>, failure::Error> = match metadata(filename) {
         Ok(ref file_type) if file_type.is_dir() => parse_directory(filename),
         Ok(_file_type) => {
-            // any other file types (namely, symlinks)
+            // any other file types (files, symlinks)
             bail!("wrangler kv:bucket upload takes a directory",)
         }
         Err(e) => bail!(e),
@@ -37,7 +37,7 @@ pub fn delete(namespace_id: &str, filename: &Path) -> Result<(), failure::Error>
                 .map(|kv| kv.key.clone())
                 .collect::<Vec<_>>())
         }
-        Ok(_file_type) => {
+        Ok(_) => {
             // any other file types (namely, symlinks)
             bail!(
                 "{} should be a file or directory, but is a symlink",
--- a/src/commands/kv/bucket.rs
+++ b/src/commands/kv/bucket.rs
@@ -11,8 +11,6 @@ use std::ffi::OsString;
 use std::fs::metadata;
 use std::path::Path;
 
-use failure::bail;
-
 use crate::terminal::message;
 
 pub fn upload(namespace_id: &str, filename: &Path) -> Result<(), failure::Error> {
@@ -20,9 +18,9 @@ pub fn upload(namespace_id: &str, filename: &Path) -> Result<(), failure::Error>
         Ok(ref file_type) if file_type.is_dir() => parse_directory(filename),
         Ok(_file_type) => {
             // any other file types (files, symlinks)
-            bail!("wrangler kv:bucket upload takes a directory",)
+            failure::bail!("wrangler kv:bucket upload takes a directory")
         }
-        Err(e) => bail!(e),
+        Err(e) => failure::bail!(e),
     };
 
     write_bulk(namespace_id, pairs?)
@@ -39,12 +37,12 @@ pub fn delete(namespace_id: &str, filename: &Path) -> Result<(), failure::Error>
         }
         Ok(_) => {
             // any other file types (namely, symlinks)
-            bail!(
+            failure::bail!(
                 "{} should be a file or directory, but is a symlink",
                 filename.display()
             )
         }
-        Err(e) => bail!(e),
+        Err(e) => failure::bail!(e),
     };
 
     delete_bulk(namespace_id, keys?)
--- a/src/commands/kv/bucket.rs
+++ b/src/commands/kv/bucket.rs
@@ -15,7 +15,7 @@ use crate::terminal::message;
 
 pub fn upload(namespace_id: &str, filename: &Path) -> Result<(), failure::Error> {
     let pairs: Result<Vec<KeyValuePair>, failure::Error> = match &metadata(filename) {
-        Ok(file_type) if file_type.is_dir() => parse_directory(filename),
+        Ok(file_type) if file_type.is_dir() => directory_keys_values(filename),
         Ok(_file_type) => {
             // any other file types (files, symlinks)
             failure::bail!("wrangler kv:bucket upload takes a directory")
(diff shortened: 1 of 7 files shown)


src/commands/kv/bucket.rs
```diff
@@ -0,0 +1,103 @@
...
+
+use failure::bail;
+
+use crate::terminal::message;
+
+pub fn upload(namespace_id: &str, filename: &Path) -> Result<(), failure::Error> {
+    let pairs: Result<Vec<KeyValuePair>, failure::Error> = match metadata(filename) {
+        Ok(ref file_type) if file_type.is_dir() => parse_directory(filename),
+        Ok(_file_type) => {
+            // any other file types (namely, symlinks)
```
gabbifish: This comment is old! We can replace with
```suggestion
            // any other file types (files, symlinks)
```

src/commands/kv/bucket.rs
```diff
@@ -0,0 +1,103 @@
...
+    };
+
+    write_bulk(namespace_id, pairs?)
+}
+
+pub fn delete(namespace_id: &str, filename: &Path) -> Result<(), failure::Error> {
+    let keys: Result<Vec<String>, failure::Error> = match metadata(filename) {
+        Ok(ref file_type) if file_type.is_dir() => {
+            let key_value_pairs = parse_directory(filename)?;
+            Ok(key_value_pairs
```
gabbifish: I don't think we need any of this clone logic; we should just be fine with `parse_directory(filename)` (that worked when I implemented this previously):
```rust
pub fn delete_directory(namespace_id: &str, filename: &Path) -> Result<(), failure::Error> {
    let keys: Result<Vec<String>, failure::Error> = match metadata(filename) {
        Ok(ref file_type) if file_type.is_dir() => parse_directory(filename),
        Ok(_) => bail!("{} should be a directory, but is not", filename.display()),
        Err(e) => bail!(e),
    };

    delete_bulk(namespace_id, keys?)
}

```

src/commands/kv/bucket.rs
```diff
@@ -0,0 +1,103 @@
...
+        Err(e) => bail!(e),
+    };
+
+    delete_bulk(namespace_id, keys?)
+}
+
+fn parse_directory(directory: &Path) -> Result<Vec<KeyValuePair>, failure::Error> {
+    let mut upload_vec: Vec<KeyValuePair> = Vec::new();
+    for entry in WalkDir::new(directory) {
+        let entry = entry.unwrap();
```
gabbifish: Per Steve's comments, `let entry = entry.unwrap().path();`

src/commands/kv/bucket.rs
```diff
@@ -0,0 +1,103 @@
...
+            bail!("wrangler kv:bucket upload takes a directory",)
+        }
+        Err(e) => bail!(e),
+    };
+
+    write_bulk(namespace_id, pairs?)
+}
+
+pub fn delete(namespace_id: &str, filename: &Path) -> Result<(), failure::Error> {
+    let keys: Result<Vec<String>, failure::Error> = match metadata(filename) {
```
gabbifish: Per Steve's comments, 
```
match &metadata(filename) {
    Ok(file_type) if ...
```

src/commands/kv/bucket.rs
```diff
@@ -0,0 +1,103 @@
...
+        Ok(ref file_type) if file_type.is_dir() => {
+            let key_value_pairs = parse_directory(filename)?;
+            Ok(key_value_pairs
+                .iter()
+                .map(|kv| kv.key.clone())
+                .collect::<Vec<_>>())
+        }
+        Ok(_file_type) => {
```
gabbifish: Per Steve-O:
```suggestion
        Ok(_) => {
```

src/commands/kv/bucket.rs
```diff
@@ -0,0 +1,103 @@
...
+            // any other file types (namely, symlinks)
+            bail!(
+                "{} should be a file or directory, but is a symlink",
+                filename.display()
+            )
+        }
+        Err(e) => bail!(e),
```
gabbifish: We can use `failure::bail!(e)` to improve consistency within Wrangler; it was my bad for making the `use failure::bail` pattern a thing in the first place :( I have since moved to calling `failure::bail!(e)` directly.

ashleymichal:
also need to update documentation.

src/commands/kv/bucket.rs
ashleymichal: the old logic had individual parse_directory functions that returned the appropriate type (`Vec<String>` and `Vec<KeyValuePair>`. This one uses a single function.

src/commands/kv/bucket.rs
ashleymichal: tried this; the compiler complained about a dropped borrow and suggested re-adding the let...

src/commands/kv/bucket.rs
ashleymichal: tried this, the compiler hated me.

src/commands/kv/bucket.rs
ashleymichal: done.

src/commands/kv/bucket.rs
ashleymichal: joe lee helped me to figure out why the compiler hated me when i fixed this, tho not sure why this worked previous... 🤷🏻‍♀️

src/commands/kv/bucket.rs
```diff
@@ -0,0 +1,101 @@
...
+}
+
+pub fn delete(namespace_id: &str, filename: &Path) -> Result<(), failure::Error> {
+    let keys: Result<Vec<String>, failure::Error> = match &metadata(filename) {
+        Ok(file_type) if file_type.is_dir() => {
+            let key_value_pairs = parse_directory(filename)?;
+            Ok(key_value_pairs
+                .iter()
+                .map(|kv| kv.key.clone())
+                .collect::<Vec<_>>())
```
gabbifish: do we still think we can get rid of this cloning logic? I think it would be a good idea :)

ashleymichal: yes, apparently if it isn't in a PR comment I don't remember it.

src/commands/kv/bucket.rs
```diff
@@ -0,0 +1,109 @@
...
+use std::ffi::OsString;
+use std::fs::metadata;
+use std::path::Path;
+
+use crate::terminal::message;
+
+pub fn upload(namespace_id: &str, filename: &Path) -> Result<(), failure::Error> {
+    let pairs: Result<Vec<KeyValuePair>, failure::Error> = match &metadata(filename) {
+        Ok(file_type) if file_type.is_dir() => directory_keys_values(filename),
+        Ok(_file_type) => {
```
gabbifish:
```suggestion
        Ok(_) => {
```
:package:

src/commands/kv/bucket.rs
```diff
@@ -0,0 +1,109 @@
...
+    write_bulk(namespace_id, pairs?)
+}
+
+pub fn delete(namespace_id: &str, filename: &Path) -> Result<(), failure::Error> {
+    let keys: Result<Vec<String>, failure::Error> = match &metadata(filename) {
+        Ok(file_type) if file_type.is_dir() => directory_keys_only(filename),
+        Ok(_) => {
+            // any other file types (namely, symlinks)
+            failure::bail!(
+                "{} should be a file or directory, but is a symlink",
```
gabbifish:
```suggestion
                "wrangler kv:bucket delete takes a directory",
```

src/commands/kv/bucket.rs
gabbifish: *shakes fist* BORROW CHECKER!!!

PR #515, opened 2019-09-03
Minor formatting fix in README.md

kentonv:
Github treats the sentence "This key is optional..." as part of the bullet point from the previous line, which is not what was intended. Inserting a blank line in between fixes the problem.


EverlastingBugstopper:
Thanks @kentonv!

PR #525, opened 2019-09-05
Disallow duplicate environment names

### LINKED ISSUE #480: Disallow duplicate names in environments
We don't want people accidentally overwriting production workers when they think they're deploying to staging

EverlastingBugstopper:
Fixes #480

### DIFF
--- a/src/settings/project/mod.rs
+++ b/src/settings/project/mod.rs
@@ -381,7 +381,7 @@ fn check_for_duplicate_names(manifest: &Manifest) -> Result<(), failure::Error> 
     let duplicate_message = match duplicate_names.len() {
         1 => Some(format!("this name is duplicated: {}", duplicate_names[0])),
         n if n >= 2 => Some(format!("these names are duplicated: {:?}", duplicate_names)),
-        _ => None
+        _ => None,
     };
     match duplicate_message {
         Some(msg) => failure::bail!(format!(
@@ -389,7 +389,7 @@ fn check_for_duplicate_names(manifest: &Manifest) -> Result<(), failure::Error> 
             emoji::WARN,
             msg
         )),
-        None => Ok(())
+        None => Ok(()),
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/settings/project/mod.rs
```diff
@@ -355,5 +358,40 @@ id = "0f2ac74b498b48028cb68387c421e279"
...
+    names.push(manifest.name.to_string());
+    match &manifest.env {
+        Some(environments) => {
+            for (_, environment) in environments.iter() {
+                match &environment.name {
+                    Some(name) => {
+                        if names.contains(name) && !duplicate_names.contains(name) {
+                            duplicate_names.push(name.to_string());
+                        }
+                        names.push(name.to_string());
```
ashleymichal: should this be in an else? it likely won't matter in the long run, but this vec will effectively end up including the duplicates.

src/settings/project/mod.rs
```diff
@@ -355,5 +358,40 @@ id = "0f2ac74b498b48028cb68387c421e279"
...
+fn check_for_duplicate_names(manifest: &Manifest) -> Result<(), failure::Error> {
+    let mut names: Vec<String> = Vec::new();
+    let mut duplicate_names: Vec<String> = Vec::new();
+    names.push(manifest.name.to_string());
+    match &manifest.env {
+        Some(environments) => {
+            for (_, environment) in environments.iter() {
+                match &environment.name {
+                    Some(name) => {
+                        if names.contains(name) && !duplicate_names.contains(name) {
```
gabbifish: General feedback: this type of repetition detection logic is fantastic to do with Sets: https://doc.rust-lang.org/std/collections/struct.HashSet.html. I would recommend completely replacing `names` and `duplicate_names` with Hashsets instead (it's also more performant do to this, though scale isn't really a problem here).

src/settings/project/mod.rs
EverlastingBugstopper: yes

PR #527, opened 2019-09-05
refactor: organize kv by command type

ashleymichal:
Because I'm picky like that.

### DIFF
--- a/src/commands/kv/bulk/mod.rs
+++ b/src/commands/kv/bulk/mod.rs
@@ -1,0 +1,7 @@
+mod delete;
+mod put;
+
+pub use delete::delete;
+pub use put::put;
+
+const MAX_PAIRS: usize = 10000;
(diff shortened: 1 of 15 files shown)


src/commands/kv/bulk/mod.rs
```diff
@@ -1,37 +1,50 @@
...
 use crate::commands::kv;
+use crate::message;
 use crate::settings::global_user::GlobalUser;
 use crate::settings::project::Project;
-use crate::terminal::message;
 
 const MAX_PAIRS: usize = 10000;
 
-pub fn write_json(
+fn delete_bulk(
```
gabbifish: I think it would be good to move this to delete.rs and make it public, given that bucket will use it in the future :)

src/commands/kv/bulk/mod.rs
```diff
@@ -1,37 +1,50 @@
...
+
+    match response {
+        Ok(_success) => message::success("Success"),
+        Err(e) => kv::print_error(e),
+    }
+
+    Ok(())
 }
 
 fn write_bulk(
```
gabbifish: I think it would be good to move this to write.rs and make it public, given that buckets will use it in the future :)

Issue #522, opened 2019-09-04
Live update preview doesn't work for rust projects


gabbifish:
#### Environment

* operating system:
* output of `rustc -V`: rustc 1.36.0 (a53f9df32 2019-07-03)
* output of `node -v`: v8.10.0
* output of `wrangler -V`: wrangler 1.2.0

#### Steps to reproduce

```
wrangler generate rust-test --type=rust
cd rust-test
wrangler preview --watch
```

#### What did you expect to see?
Live reload should successfully stay alive

#### What did you see instead?
```
$ wrangler preview --watch
⬇️ Installing wasm-pack...
 Compiling your project to WebAssembly...
[INFO]: Checking for the Wasm target...
[INFO]: Compiling to Wasm...
    Finished release [optimized] target(s) in 0.01s
[INFO]: Installing wasm-bindgen...
[INFO]: Optional fields missing from Cargo.toml: 'description', 'repository', and 'license'. These are not necessary, but recommended
[INFO]: :-) Done in 0.09s
[INFO]: :-) Your wasm pkg is ready to publish at ./pkg.
⬇️ Installing wasm-pack...
Error: Your JavaScript project is missing a `package.json` file; is `./` the wrong directory?
```

PR #528, opened 2019-09-06
Fix javascript live preview for linux

### LINKED ISSUE #517: `wrangler preview --watch` only detects the first change (on Linux)
#### Environment

* operating system: Linux (Debian Buster; kernel 4.19.37)
* output of `rustc -V`: N/A
* output of `node -v`: N/A
* output of `wrangler -V`: 1.2.0

#### Steps to reproduce

1. Run `wrangler preview --watch`.
2. See preview.
3. Make a change to the code.
4. See preview update. Cool!
5. Make another change to the code.

Preview updates again.

Preview did not update.

The full output of Wrangler at this point is:

```
 watching "worker.js"
 Detected changes...
 Detected change during cooldown...
 Cooldown over, propogating changes...
 Updated preview with changes
```

No matter how many times I make changes to `worker.js`, no more output is shown and the preview is not updated.

It appears that Wrangler is failing to notice the file changing after the first time. Since file change notification APIs vary wildly across platforms, it's very possible this bug is Linux-specific.

gabbifish:
Closes #517. 

This was a fun one to investigate. The bug, at its core, was the fact that different editors have different approaches to saving files--some, like vim, actually delete and re-create an entire file when saving its contents. Different file notification APIs across different operating systems track changes to files differently; in the case of the Linux file watcher in the notify crate (inotify) this watcher will stop watching a file once it has been deleted; vim deletes files regularly when saving new contents (https://unix.stackexchange.com/questions/188873/using-inotifywait-along-with-vim). 

I wasn't familiar with the live preview code, so I spent time reading through it. Ultimately I couldn't find any bugs that could cause the behavior I observed. Something we noticed was that webpack workers watch directories and worked just fine with vim, implying that watching a directory for javascript workers would also work (inotify would continue to watch the index.js file even if vim deleted and re-created it).

This PR may also need to be replicated for #522, given that the Rust live preview service also watches an individual javascript file.

This PR also reduces noisy output by the live preview service. Notifying over directories produced a large amount of repeated `"Detected change during cooldown..."` terminal output.

### DIFF
--- a/src/commands/build/watch/mod.rs
+++ b/src/commands/build/watch/mod.rs
@@ -14,6 +14,7 @@ use std::thread;
 use std::time::Duration;
 
 pub const COOLDOWN_PERIOD: Duration = Duration::from_millis(2000);
+const JAVASCRIPT_PATH: &str = "./";
 
 /// watch a project for changes and re-build it when necessary,
 /// outputting a build event to tx.
@@ -24,17 +25,12 @@ pub fn watch_and_build(
     let project_type = &project.project_type;
     match project_type {
         ProjectType::JavaScript => {
-            let path = "./";
-            let package = Package::new(path)?;
-            // Confirm entrypoint is provided
-            package.main()?;
-
             thread::spawn(move || {
                 let (watcher_tx, watcher_rx) = mpsc::channel();
                 let mut watcher = notify::watcher(watcher_tx, Duration::from_secs(1)).unwrap();
 
-                watcher.watch(&path, RecursiveMode::Recursive).unwrap();
-                message::info(&format!("watching {:?}", &path));
+                watcher.watch(JAVASCRIPT_PATH, RecursiveMode::Recursive).unwrap();
+                message::info(&format!("watching {:?}", &JAVASCRIPT_PATH));
 
                 loop {
                     match wait_for_changes(&watcher_rx, COOLDOWN_PERIOD) {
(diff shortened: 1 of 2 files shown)


src/commands/build/watch/watcher.rs
```diff
@@ -18,15 +18,12 @@ pub fn wait_for_changes(
         match get_changed_path_from_event(event) {
             Ok(Some(path)) => {
                 message::working("Detected changes...");
-                //wait for cooldown
-                while let Ok(_) = rx.recv_timeout(cooldown) {
-                    message::working("Detected change during cooldown...");
```
xtuc: Maybe the logs were useful and we can turn them into debug logs?

src/commands/build/watch/mod.rs
```diff
@@ -24,14 +24,17 @@ pub fn watch_and_build(
     let project_type = &project.project_type;
     match project_type {
         ProjectType::JavaScript => {
-            let package = Package::new("./")?;
-            let entry = package.main()?;
+            let path = "./";
+            let package = Package::new(path)?;
+            // Confirm entrypoint is provided
```
xtuc: Why do we need to confirm that an entrypoint is provided? I would assume it will fail during the build, but the watch shouldn't care?

gabbifish: OOh, this is a good point. I will remove this logic.

src/commands/build/watch/watcher.rs
gabbifish: I chatted with @ashleymichal and she said that these logs initially made sense for smaller projects (they make users feel like wrangler watch is actively working), but just became too verbose when editing entire directories (e.g. large webpack workers). I think it's worth keeping these messages out for now, unless we come up with more detailed/meaningful logs to replace them.

PR #531, opened 2019-09-06
Alewis/refactor key list

### LINKED ISSUE #447: Add `bucket` config to wrangler.toml
adding `bucket` to a kv namespace entry to allow Wrangler to orchestrate sync of a directory full of assets. Bucket should be a relative path from the location of the wrangler.toml to a directory including assets to be uploaded to KV.

EDITED: s/project.assets/kv-namespace.bucket/

ashleymichal:
Closes #447 

This extracts the logic of fetching all of the keys for a given KV namespace into an iterator, which will allow us to either print to standard out, or collect the keys as needed for other operations, without repeating all the fetch cursor logic. This is handy for implementing `sync`, which will need to collect a list of uploaded keys from the API to compare to local keys in order to decide what to delete.

### DIFF
--- a/src/commands/kv/key/key_list.rs
+++ b/src/commands/kv/key/key_list.rs
@@ -1,0 +1,120 @@
+use cloudflare::endpoints::workerskv::list_namespace_keys::ListNamespaceKeys;
+use cloudflare::endpoints::workerskv::list_namespace_keys::ListNamespaceKeysParams;
+use cloudflare::endpoints::workerskv::Key;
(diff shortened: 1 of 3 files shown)


ashleymichal:
need to handle errors outside of the iterator.

gabbifish:
I opened a PR (#536) that should make this logic a little simpler--let me know what you think!

ashleymichal:
> I opened a PR (#536) that should make this logic a little simpler--let me know what you think!

Perfect! Merged. Ready for review

PR #534, opened 2019-09-06
Refactor environment parsing

EverlastingBugstopper:
This PR changes a lot of my bad Rust code into slightly better Rust code. The main goal here is readability.

### DIFF
--- a/src/settings/project/mod.rs
+++ b/src/settings/project/mod.rs
@@ -85,6 +85,117 @@ impl Manifest {
         Ok(manifest)
     }
 
+    fn get_environment(
+        &self,
+        environment_name: Option<&str>,
+    ) -> Result<Option<&Environment>, failure::Error> {
+        if let Some(environment_name) = environment_name {
+            if let Some(environment_table) = &self.env {
+                if let Some(environment) = environment_table.get(environment_name) {
+                    Ok(Some(environment))
+                } else {
+                    failure::bail!(format!(
+                        "{} Could not find environment with name {}",
+                        emoji::WARN,
+                        environment_name
+                    ))
+                }
+            } else {
+                failure::bail!(format!(
+                    "{} There are no environments specified in your wrangler.toml",
+                    emoji::WARN
+                ))
+            }
+        } else {
+            Ok(None)
+        }
+    }
+
+    // TODO: when --release is deprecated, this will be much easier
+    fn zoneless_or_dot_dev(
+        &self,
+        environment: Option<&Environment>,
+        release: bool,
+    ) -> Result<(Option<String>, bool), failure::Error> {
+        let use_dot_dev_warning =
+            "Please specify the workers_dot_dev boolean in the top level of your wrangler.toml";
+        let wdd_failure = format!(
+            "{} Your environment should only include `workers_dot_dev` or `route`",
+            emoji::WARN
+        );
+
+        // TODO: deprecate --release, remove warnings and parsing
+        // switch wrangler publish behavior to act the same at top level
+        // and environments
+        // brace yourself, this is hairy
+        let workers_dot_dev: bool = match environment {
+            // top level configuration
+            None => {
+                if release {
+                    // --release means zoned, not workers.dev
+                    match self.workers_dot_dev {
+                        Some(_) => failure::bail!(use_dot_dev_warning),
+                        None => {
+                            message::warn(use_dot_dev_warning);
+                            false // workers_dot_dev defaults to false when it's top level and --release is passed
+                        }
+                    }
+                } else {
+                    if let Some(wdd) = self.workers_dot_dev {
+                        if wdd {
+                            if let Some(route) = &self.route {
+                                if !route.is_empty() {
+                                    failure::bail!(wdd_failure)
+                                }
+                            }
+                        }
+                        wdd
+                    } else {
+                        message::warn(use_dot_dev_warning);
+                        true
+                    }
+                }
+            }
+
+            // environment configuration
+            Some(environment) => {
+                if let Some(wdd) = environment.workers_dot_dev {
+                    if wdd && environment.route.is_some() {
+                        failure::bail!(wdd_failure)
+                    }
+                    wdd
+                } else {
+                    if let Some(wdd) = self.workers_dot_dev {
+                        if wdd && environment.route.is_some() {
+                            false // allow route to override workers_dot_dev = true if wdd is inherited
+                        } else {
+                            wdd // inherit from top level
+                        }
+                    } else {
+                        false // if absent -> false
+                    }
+                }
+            }
+        };
+
+        let route = if let Some(environment) = environment {
+            if let Some(route) = &environment.route {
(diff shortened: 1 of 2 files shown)


src/settings/project/mod.rs
```diff
@@ -84,54 +84,50 @@ impl Manifest {
...
+    fn negotiate_with_terrorists(
+        &self,
+        environment: Option<&Environment>,
+        release: bool,
+    ) -> Result<(Option<String>, bool), failure::Error> {
         let deprecate_warning =
             "Please specify the workers_dot_dev boolean in the top level of your wrangler.toml";
         let wdd_failure = format!(
-            "{} Your environment should only include `workers_dot_dev` or `route`",
+            "{} Your environment should only include `workers_dot_dev` or `route`. If you are trying to publish to workers.dev, add `workers_dot_dev = true`, if you are trying to publish to your own domain, add a route.",
```
EverlastingBugstopper: @victoriabernard92 does this look better?

victoriabernard92:
```suggestion
            "{} Your environment should only include `workers_dot_dev` or `route`. To publish to workers.dev, remove the line with route. To publish to your own domain, set `workers_dot_dev` to false ",
```

EverlastingBugstopper: awesome thanks!

src/settings/project/mod.rs
```diff
@@ -143,144 +139,116 @@ impl Manifest {
                             false // workers_dot_dev defaults to false when it's top level and --release is passed
```
gabbifish: Does it actually make sense to bail! on the Some(_) case above, given that the user is actually using `workers_dot_dev`? I would expect for us to just follow what `workers_dot_dev` specifies in this case. This surprised me a little--let me know if I'm thinking incorrectly!

Also, in the None case where we return false, should we modify the warning message to also clarify that we are falling back to deploying to a route?

src/settings/project/mod.rs
```diff
@@ -84,54 +84,50 @@ impl Manifest {
...
+            "{} Your environment should only include `workers_dot_dev` or `route`. If you are trying to publish to workers.dev, remove `route` from your wrangler.toml, if you are trying to publish to your own domain, remove `workers_dot_dev`.",
             emoji::WARN
         );
 
         // TODO: deprecate --release, remove warnings and parsing
         // switch wrangler publish behavior to act the same at top level
         // and environments
         // brace yourself, this is hairy
-        let workers_dot_dev = match environment {
+        let workers_dot_dev: bool = match environment {
```
gabbifish: Part of me wonders if there is a way to further simplify the logic around the truth table for `workers_dot_dev`, `release`, and `route`. What we have here is totally serviceable--I think, however, this logic could become even more understandable for folks if we reduced the nested matches and focused on easy-to-follow compound conditionals.

For example, we could use logic that could do the following pseudocode:
1. Extract a specified environment only once using one match statement.
2. If environment does not exist (is `None`, so top level configs used):
    a. If `release == true` and `workers_dot_dev.is_some()`, throw error message stating conflict between using `--release` and `workers_dot_dev`.
    b. If `release == true` and `workers_dot_dev.is_none()`, warn of deprecation and return false (proceed to deploy to route)
    c. If `release == false` and `self.workers_dot_dev.is_some()` and `!self.route.is_empty()`, bail with `wdd_failure`
    d. If `release == false` and `self.workers_dot_dev.is_some()` and `self.route.is_empty()`, return `workers_dot_dev` value
    e. If `release == false` and `self.workers_dot_dev.is_none()` return true.

src/settings/project/mod.rs
```diff
@@ -143,144 +139,116 @@ impl Manifest {
...
-        let name = match environment {
-            Some(environment) => match &environment.name {
-                Some(name) => {
-                    let name = name.clone();
-                    if name == self.name {
-                        failure::bail!(format!(
-                            "{} Each `name` in your wrangler.toml must be unique",
-                            emoji::WARN
-                        ))
+        let route = if let Some(environment) = environment {
```
gabbifish: Same advice above can apply here--I think it's possible we can reduce nested matches :D

src/settings/project/mod.rs
```diff
@@ -143,144 +139,116 @@ impl Manifest {
...
+        if let Some(environment) = environment {
+            target.name = if let Some(name) = &environment.name {
+                name.clone()
+            } else {
+                match environment_name {
+                    Some(environment_name) => format!("{}-{}", self.name, environment_name),
+                    None => failure::bail!("You must specify `name` in your wrangler.toml"),
+                }
+            };
+            if let Some(account_id) = &environment.account_id {
```
gabbifish: This and following logic looks great!

src/settings/project/mod.rs
```diff
@@ -143,144 +139,116 @@ impl Manifest {
...
+            if environment.routes.is_some() {
+                target.routes = environment.routes.clone();
+            }
+            if environment.webpack_config.is_some() {
+                target.webpack_config = environment.webpack_config.clone();
+            }
+            if environment.zone_id.is_some() {
+                target.zone_id = environment.zone_id.clone();
+            }
+            target.kv_namespaces = environment.kv_namespaces.clone();
```
gabbifish: Do we want to wrap `environment.kv_namespaces` in the same conditionals as above? I would think kv inheritance would work the same as other toml fields...

src/settings/project/mod.rs
EverlastingBugstopper: We don't want `--release` and `workers_dot_dev` to be compatible, if they are using `workers_dot_dev`, they have opted in to the new type of `wrangler.toml` and should not be using the `--release` flag

src/settings/project/mod.rs
EverlastingBugstopper: @ashleymichal and I talked about this and said that it doesn't really make sense to inherit kv namespaces because it's an anti-pattern to use multiple namespaces across different environments. I don't want my staging worker to use my production kv namespaces. In this case - if you wanted to use the same namespace for both, you would just have to copy + paste to explicitly use the same namespaces for both environments. Open to discussion about this.

src/settings/project/mod.rs
EverlastingBugstopper: I'll work on this today :)

EverlastingBugstopper: Actually, can we leave this so we can merge refactor today? @gabbifish

gabbifish: Makes sense to me! I'll save the link to this convo and we can revisit it after this is merged.

src/settings/project/mod.rs
gabbifish: This makes sense. It could be worth adding this as a comment to clarify why this logic diverges from the rest of the logic preceding it. Other than that, LGTM!

src/settings/project/mod.rs
gabbifish: Gotcha :)

PR #537, opened 2019-09-09
Gabbi/more detailed subcommand docs

### LINKED ISSUE #510: Readme for KV

Add to readme

#### Basic example

* A copy pasteable .json file example for bulk uploads instead of just the definition
* Example of how to set up a file to be the value of a key (that is possible with the commands right 🤔?)
* what `--prefix` does

gabbifish:
Closes #510. Adds more detailed documentation for KV subcommands.

### DIFF
--- a/docs/content/kv_commands.md
+++ b/docs/content/kv_commands.md
@@ -71,7 +71,7 @@ $ wrangler kv:namespace list
 
 Writes a single key/value pair to the given namespace. Optional params include 
 1. `--ttl`: Number of seconds for which the entries should be visible before they expire. At least 60. Takes precedence over 'expiration' option.
-2. `--expiration`: Number of seconds since the UNIX epoch, indicating when the key-value pair should expire.
+1. `--expiration`: Number of seconds since the UNIX epoch, indicating when the key-value pair should expire.
 3. `--path`: Read value from the file at a given path. *This is good for security-sensitive operations, like uploading keys to KV; uploading from a file prevents a key value from being saved in areas like your terminal history.*
 
 
--- a/docs/content/kv_commands.md
+++ b/docs/content/kv_commands.md
@@ -94,7 +94,7 @@ Reads a single value by key from the given namespace.
 
 ```sh
 $ wrangler kv:key get f7b02e7fc70443149ac906dd81ec1791 "key"
-value
+=> value
 ```
 
 ### `delete`
--- a/docs/content/kv_commands.md
+++ b/docs/content/kv_commands.md
@@ -67,12 +67,16 @@ $ wrangler kv:namespace list
 
 ### `put`
 
-Writes a single key/value pair to the given namespace.
+Writes a single key/value pair to the given namespace. Optional params include 
+1. `--ttl`: Number of seconds for which the entries should be visible before they expire. At least 60. Takes precedence over 'expiration' option.
+2. `--expiration`: Number of seconds since the UNIX epoch, indicating when the key-value pair should expire.
+
 
 #### Usage
 
 ```sh
 $ wrangler kv:key put f7b02e7fc70443149ac906dd81ec1791 "key" "value" --ttl=10000
+✨  Success
 ```
 
 ### `get`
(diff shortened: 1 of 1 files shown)


docs/content/kv_commands.md
```diff
@@ -67,12 +69,21 @@ $ wrangler kv:namespace list
 
 ### `put`
 
-Writes a single key/value pair to the given namespace.
+Writes a single key/value pair to the given namespace. Optional params include 
+1. `--ttl`: Number of seconds for which the entries should be visible before they expire. At least 60. Takes precedence over 'expiration' option.
+2. `--expiration`: Number of seconds since the UNIX epoch, indicating when the key-value pair should expire.
```
xtuc:
```suggestion
1. `--expiration`: Number of seconds since the UNIX epoch, indicating when the key-value pair should expire.
```

docs/content/kv_commands.md
```diff
@@ -67,12 +69,21 @@ $ wrangler kv:namespace list
...
+3. `--path`: Read value from the file at a given path. *This is good for security-sensitive operations, like uploading keys to KV; uploading from a file prevents a key value from being saved in areas like your terminal history.*
```
xtuc:
```suggestion
1. `--path`: Read value from the file at a given path. *This is good for security-sensitive operations, like uploading keys to KV; uploading from a file prevents a key value from being saved in areas like your terminal history.*
```

xtuc: Markdown will renum them correctly, it's easier for adding/removing one later. Feel free to refuse if it seems overkilled to you.

docs/content/kv_commands.md
```diff
@@ -83,6 +94,7 @@ Reads a single value by key from the given namespace.
 
 ```sh
 $ wrangler kv:key get f7b02e7fc70443149ac906dd81ec1791 "key"
+value
```
xtuc: This looks like I need to enter `value` in my terminal on the next line.

docs/content/kv_commands.md
```diff
@@ -101,26 +113,41 @@ yes
...
+$ wrangler kv:key list f7b02e7fc70443149ac906dd81ec1791 --prefix="public" | python -m json.tool
+[
+    {
+        "name": "public_key"
+    }, 
+    {
+        "name": "public_key_with_expiration",
+        "expiration": 1568014518
+    } 
+]
```
xtuc: So you removed the table for JSON output? I think if you prettier print it from wrangler it should be readable enough :+1:

docs/content/kv_commands.md
```diff
@@ -101,26 +113,41 @@ yes
...
 ```
 
 ## `kv:bulk`
 
-### JSON body
+### `put`
 
-Bulk operations take as an argument a pre-built JSON file, which should be a list of objects with the following schema:
+Writes a file full of key/value pairs to the given namespace. Takes as an argument a JSON file with a list of key-value pairs to upload (see JSON spec above). An example of json input:
```
xtuc:
```suggestion
Writes a file full of key/value pairs to the given namespace. Takes as an argument a JSON file with a list of key-value pairs to upload (see JSON spec above). An example of JSON input:
```

docs/content/kv_commands.md
gabbifish: I agree. I can make replace this with a more explicit string that signifies "this is a value returned by the API and output to the terminal". Like 
```suggestion
value-returned-from-api-and-output-to-terminal
```

docs/content/kv_commands.md
gabbifish: Yeah, the table would get super unwieldy for very large namespaces :D JSON is easier for folks to handle.

docs/content/kv_commands.md
ashleymichal: or give a classic `=>` return signifier

gabbifish: Ooh yes!

gabbifish:
```suggestion
=> value
```

PR #541, opened 2019-09-09
Print out namespaces list as json by default

gabbifish:
Make json output default for listing namespaces as well (and update documentation)!

### DIFF
--- a/docs/content/kv_commands.md
+++ b/docs/content/kv_commands.md
@@ -86,16 +86,20 @@ yes
 Outputs a list of all KV namespaces associated with your account id.
 
 #### Usage
+The example below uses Python's JSON pretty-printing command line tool to pretty-print output.
 
 ```sh
-$ wrangler kv:namespace list
-🌀  Retrieving namespaces
-✨  Success:
-+---------------+----------------------------------+
-| TITLE         | ID                               |
-+---------------+----------------------------------+
-| New Namespace | f7b02e7fc70443149ac906dd81ec1791 |
-+---------------+----------------------------------+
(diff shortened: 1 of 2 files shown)


EverlastingBugstopper:
Do we want to default to pretty print with the table and then print json if we pass a `--json` flag? I feel like that's the usual pattern for tools that output json. Not a hill I'll die on though

gabbifish:
@EverlastingBugstopper I think we can get away without offering pretty-printing right now--there are plenty of other commandline tools one can just pipe the wrangler output to in order to pretty-print it :) (I include an example of this in the docs)

docs/content/kv_commands.md
```diff
@@ -53,16 +53,20 @@ $ wrangler kv:namespace rename f7b02e7fc70443149ac906dd81ec1791 "updated kv name
...
 ```sh
-$ wrangler kv:namespace list
-🌀  Retrieving namespaces
-✨  Success:
-+------------------+----------------------------------+
-| TITLE            | ID                               |
-+------------------+----------------------------------+
-| new kv namespace | f7b02e7fc70443149ac906dd81ec1791 |
-+------------------+----------------------------------+
+$ wrangler kv:namespace list | python -m json.tool
```
xtuc: Why can't we pretty print it from Rust?

gabbifish: Enough people have brought this up that I'll add pretty printing and a flag for it :)

gabbifish: nvm, gonna recommend folks use jq. It's hard for us to pretty-print the keys json since it is streamed.

PR #557, opened 2019-09-11
Gabbi/perscriptive namespace titles

### LINKED ISSUE #551: Wrangler should be prescriptive about KV Namespace Titles
## The Problem
KV Namespaces are generally identified in the UI by their "title" property, a human-readable string which must be unique within an account. However, in the context of a Wrangler project, this property is generally only referenced when creating and listing namespaces. In all other cases, it makes more sense to refer to a namespace by its binding as specified in the project's wrangler.toml.

## The Solution
To ease confusion over the relationship between a title and a binding, Wrangler should provide an interface, feedback, and naming conventions that help to bridge these concepts.

### Example Interaction
``` sh
& wrangler kv create myblogassets --env=prod
Success! Namespace created: {
  title: "projectname-myblogassets", id: "2ce8545fc04147fab45aaceb9fb9393f"
}
Add the following to your wrangler.toml under `[env.prod]`:
kv-namespaces = [
  { binding: "myblogassets", id: "2ce8545fc04147fab45aaceb9fb9393f" }
]
```
Where the convention for namespace "title" is `<name>-<binding>`. This should preserve a reasonable level of uniqueness (worker names must be unique, but a user might reasonably use the same binding names across worker projects), while still providing human-readable namespace titles in the UI, and connecting a namespace title with its binding.

We can refine the help message based on things like whether they have existing kv namespaces ("add this to your kv-namespaces array"). Perhaps future iterations will just make a namespace per environment on your behalf as a convenience. We'll see.

We should update the KV command documentation accordingly to refer to this arg as `binding` rather than `title`.

EDIT: a previous conversation concluded that `binding + env` would be sufficient for uniqueness, but that only applies within a project. Updated to suggest we use `project name + binding` instead.

gabbifish:
Closes #551

### DIFF
--- a/src/commands/kv/namespace/create.rs
+++ b/src/commands/kv/namespace/create.rs
@@ -7,9 +7,15 @@ use crate::settings::global_user::GlobalUser;
 use crate::settings::target::Target;
 use crate::terminal::message;
 
-pub fn create(target: &Target, user: GlobalUser, title: &str) -> Result<(), failure::Error> {
+pub fn create(
+    target: &Target,
+    env: Option<&str>,
+    user: GlobalUser,
+    binding: &str,
+) -> Result<(), failure::Error> {
     let client = kv::api_client(user)?;
 
+    let title = format!("{}-{}", target.name, binding);
     let msg = format!("Creating namespace with title \"{}\"", title);
     message::working(&msg);
 
@@ -21,7 +27,18 @@ pub fn create(target: &Target, user: GlobalUser, title: &str) -> Result<(), fail
     });
 
     match response {
-        Ok(success) => message::success(&format!("Success: {:#?}", success.result)),
+        Ok(_) => {
+            message::success("Success");
+            match env {
+                Some(env) => println!("Add the following to your TOML under [env.{}]:", env),
+                None => println!("Add the following to your TOML:"),
+            };
+            println!(
+                "kv-namespaces = [ \n\
+                \t {{ binding: \"myblogassets\", id: \"2ce8545fc04147fab45aaceb9fb9393f\" }} \n\
+                ]"
+            );
+        }
         Err(e) => kv::print_error(e),
     }
 
--- a/src/commands/kv/namespace/create.rs
+++ b/src/commands/kv/namespace/create.rs
@@ -38,8 +38,9 @@ pub fn create(
             };
             println!(
                 "kv-namespaces = [ \n\
-                 \t {{ binding: \"myblogassets\", id: \"2ce8545fc04147fab45aaceb9fb9393f\" }} \n\
-                 ]"
+                 \t {{ binding: \"{}\", id: \"{}\" }} \n\
+                 ]",
+                binding, success.result.id
             );
         }
         Err(e) => kv::print_error(e),
--- a/src/commands/kv/namespace/create.rs
+++ b/src/commands/kv/namespace/create.rs
@@ -27,16 +27,19 @@ pub fn create(
     });
 
     match response {
-        Ok(_) => {
-            message::success("Success");
+        Ok(success) => {
+            message::success(&format!("Success: {:#?}", success.result));
             match env {
-                Some(env) => println!("Add the following to your TOML under [env.{}]:", env),
-                None => println!("Add the following to your TOML:"),
+                Some(env) => message::success(&format!(
+                    "Add the following to your TOML under [env.{}]:",
+                    env
+                )),
+                None => message::success(&format!("Add the following to your TOML:")),
             };
             println!(
                 "kv-namespaces = [ \n\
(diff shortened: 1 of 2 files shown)


src/commands/kv/namespace/create.rs
```diff
@@ -7,9 +7,15 @@ use crate::settings::global_user::GlobalUser;
...
-pub fn create(target: &Target, user: GlobalUser, title: &str) -> Result<(), failure::Error> {
+pub fn create(
+    target: &Target,
+    env: Option<&str>,
+    user: GlobalUser,
+    binding: &str,
+) -> Result<(), failure::Error> {
     let client = kv::api_client(user)?;
 
+    let title = format!("{}-{}", target.name, binding);
```
ashleymichal: this will result in a no-env title that ends with `-` like `STATIC_ASSETS-`

src/commands/kv/namespace/create.rs
```diff
@@ -21,7 +27,22 @@ pub fn create(target: &Target, user: GlobalUser, title: &str) -> Result<(), fail
...
+                Some(env) => message::success(&format!(
+                    "Add the following to your TOML under [env.{}]:",
+                    env
+                )),
+                None => message::success(&format!("Add the following to your TOML:")),
+            };
+            println!(
+                "kv-namespaces = [ \n\
+                 \t {{ binding: \"{}\", id: \"{}\" }} \n\
+                 ]",
```
ashleymichal: use `r##""##` strings here. see https://github.com/cloudflare/wrangler/blob/master/src/settings/target/mod.rs#L338

src/commands/kv/namespace/create.rs
```diff
@@ -21,7 +27,22 @@ pub fn create(target: &Target, user: GlobalUser, title: &str) -> Result<(), fail
     });
 
     match response {
-        Ok(success) => message::success(&format!("Success: {:#?}", success.result)),
+        Ok(success) => {
+            message::success(&format!("Success: {:#?}", success.result));
+            match env {
+                Some(env) => message::success(&format!(
+                    "Add the following to your TOML under [env.{}]:",
```
EverlastingBugstopper: can we say `wrangler.toml` here instead of TOML to be consistent with other messages across the codebase?

src/commands/kv/namespace/create.rs
```diff
@@ -21,7 +27,22 @@ pub fn create(target: &Target, user: GlobalUser, title: &str) -> Result<(), fail
...
+                    env
+                )),
+                None => message::success(&format!("Add the following to your TOML:")),
```
EverlastingBugstopper: same as above - `wrangler.toml`

src/commands/kv/namespace/create.rs
```diff
@@ -21,7 +27,22 @@ pub fn create(target: &Target, user: GlobalUser, title: &str) -> Result<(), fail
...
+            };
+            println!(
```
EverlastingBugstopper: should add a check here to see if `kv-namespaces` already exists, if not, keep current logic, else print "please add this to your kv-namespaces array"

src/main.rs
```diff
@@ -525,9 +534,10 @@ fn run() -> Result<(), failure::Error> {
 
         match kv_matches.subcommand() {
             ("create", Some(create_matches)) => {
-                let target = manifest.get_target(create_matches.value_of("env"), false)?;
-                let title = create_matches.value_of("title").unwrap();
-                commands::kv::namespace::create(&target, user, title)?;
+                let env = create_matches.value_of("env");
+                let target = manifest.get_target(env, false)?;
+                let binding = create_matches.value_of("binding").unwrap();
```
EverlastingBugstopper: `unwrap` here - can we do a match on `binding` and print a nice error message if it's excluded? or put in `unreachable` if `clap` prevents this from ever being `None`?

gabbifish: We're alright to use unwrap because this value is already guaranteed to exist by Clap :)

src/commands/kv/namespace/create.rs
gabbifish: This logic actually should be fine; doesn't #551 have you use the binding and target name instead of env? Both values going into the formatting string are guaranteed to exist.

src/commands/kv/namespace/create.rs
gabbifish: So for the r##<text>## point, something fascinating i just learned is that
using that notation requires you to put the text into a separate variable
you can't use a separate variable for a formatting string in rust
All calls to println! and format! require string literals as the first argument!
So I'm not actually sure if we could use r##<text>##, as much as I wish it would work

src/commands/kv/namespace/create.rs
ashleymichal: yes. 🤦‍♀

PR #578, opened 2019-09-12
Use notify-forked

EverlastingBugstopper:
The `notify` crate updated its `fsevent-sys` version which broke our builds. @ashleygwilliams forked `notify` and pinned its dependency to a version that works.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1058,7 +1058,7 @@ dependencies = [
 ]
 
 [[package]]
-name = "notify"
+name = "notify-forked"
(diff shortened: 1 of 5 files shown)


xtuc:
What's the advantage over having the code into a folder `./deps` (or similiar) in this repo?

EverlastingBugstopper:
@xtuc this is temporary - will switch back when this issue is solved https://github.com/passcod/notify/issues/211

Going to merge this because it's breaking everything

PR #598, opened 2019-09-16
feat #450: support negate flags for publish

### LINKED ISSUE #450: Support negation flags for publish, preview
A refinement of #449 :
Sometimes you just want to publish your assets, but your Worker should stay the same.
Sometimes you just want to update your worker, but your assets should stay the same.

Wrangler should support `--no-assets` and `--no-worker` flags for these use cases.

ashleymichal:
closes #450

### DIFF
--- a/src/commands/publish/mod.rs
+++ b/src/commands/publish/mod.rs
@@ -23,16 +23,16 @@ use crate::terminal::{emoji, message};
 pub fn publish(
     user: &GlobalUser,
     target: &Target,
-    no_worker: bool,
-    no_bucket: bool,
+    push_worker: bool,
+    push_bucket: bool,
 ) -> Result<(), failure::Error> {
     info!("workers_dev = {}", target.workers_dev);
 
     validate_target(target)?;
-    if !no_bucket {
+    if push_bucket {
         upload_buckets(target, user)?;
     }
-    if !no_worker {
+    if push_worker {
         commands::build(&target)?;
         publish_script(&user, &target)?;
     }
--- a/src/commands/publish/mod.rs
+++ b/src/commands/publish/mod.rs
@@ -20,13 +20,23 @@ use crate::settings::global_user::GlobalUser;
 use crate::settings::target::Target;
 use crate::terminal::{emoji, message};
 
-pub fn publish(user: &GlobalUser, target: &Target) -> Result<(), failure::Error> {
+pub fn publish(
+    user: &GlobalUser,
+    target: &Target,
+    no_worker: bool,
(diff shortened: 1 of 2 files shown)


src/commands/publish/mod.rs
```diff
@@ -20,13 +20,23 @@ use crate::settings::global_user::GlobalUser;
 use crate::settings::target::Target;
 use crate::terminal::{emoji, message};
 
-pub fn publish(user: &GlobalUser, target: &Target) -> Result<(), failure::Error> {
+pub fn publish(
+    user: &GlobalUser,
+    target: &Target,
+    no_worker: bool,
```
EverlastingBugstopper: can we change these booleans from double negatives? `!no_worker` is confusing. perhaps `push_worker` and `push_bucket` or something like that?

src/main.rs
```diff
@@ -376,6 +376,18 @@ fn run() -> Result<(), failure::Error> {
                         .short("e")
                         .long("env")
                         .takes_value(true)
+                )
+                .arg(
+                    Arg::with_name("no_worker")
```
EverlastingBugstopper: these could be the same

gabbifish: nit: these names tend to be kebab case: `no-worker`

src/main.rs
```diff
@@ -376,6 +376,18 @@ fn run() -> Result<(), failure::Error> {
...
+                        .help("run publish without uploading your worker. used if you want to publish a bucket on its own")
+                        .long("no-worker")
+                        .takes_value(false)
+                )
+                .arg(
+                    Arg::with_name("no_bucket")
```
gabbifish: same nit on kebab case as before

src/main.rs
```diff
@@ -481,16 +493,18 @@ fn run() -> Result<(), failure::Error> {
             failure::bail!("You can only pass --env or --release, not both")
         }
         let manifest = settings::target::Manifest::new(config_path)?;
+        let target;
```
gabbifish: lowkey amazed the borrow checker allowed this!!! heck yeah :dancer:

src/commands/publish/mod.rs
ashleymichal: for the publish command sure, but these are negation flags, so their purpose is to suppress standard behavior, so the flag itself is a negative. we can flip it to pass to publish, though.

src/main.rs
ashleymichal: i think i misunderstood things, and was thinking this was the variable name I'd get. Obvs that is ridiculous and this is just for display.

PR #601, opened 2019-09-17
Implement large bucket uploads split across multiple bulk KV API requests

### LINKED ISSUE #499: Split to make multiple requests when total payload exceeds 100MB
The Cloudflare API puts a hard limit of 100MB on total payload size. For bucket uploads, Wrangler should intelligently split up payloads to fit this requirement. 

We should also check the size of each key and value to prevent partial uploads.

gabbifish:
This PR closes #499.

It segments a large directory (bucket) upload to Workers KV into multiple KV calls if the payload (keys and values) provided is one of the below:
1. Over 50MB in size, OR
1. Over 10,000 KV pairs

Otherwise, the logic simply uploads the key-value pairs without segmentation.

This logic has been tested with a directory of 50 2MB files, all uploaded to a Workers KV namespace.

Waiting on #613 to be merged into master before continuing this.

### DIFF
--- a/src/commands/kv/bucket/upload.rs
+++ b/src/commands/kv/bucket/upload.rs
@@ -32,7 +32,7 @@ pub fn upload(
     validate_file_uploads(pairs.clone())?;
 
     // Create a vector of uploads; that is, a vector of vectors of key-value pairs, each of which are
-    // maximum 10K key-value pairs in size OR maximum 100MB in size.
+    // maximum 10K key-value pairs in size OR maximum ~50MB in size.
     let mut key_count = 0;
     let mut key_pair_bytes = 0;
     let mut key_value_batch: Vec<KeyValuePair> = Vec::new();
@@ -43,9 +43,9 @@ pub fn upload(
             call_put_bulk_api(target, user.clone(), namespace_id, &mut key_value_batch)?;
         } else {
             let pair = pairs.pop().unwrap();
-            if key_count + pair.key.len() > PAIRS_MAX_COUNT // Max KV pairs for request met
+            if key_count + pair.key.len() > PAIRS_MAX_COUNT
+            // Keep upload size small to keep KV bulk API happy
             || key_pair_bytes + pair.key.len() + pair.value.len() > UPLOAD_MAX_SIZE * 1/2
-            // key+value sums nearly at UPLOAD_MAX_SIZE
             {
                 call_put_bulk_api(target, user.clone(), namespace_id, &mut key_value_batch)?;
 
--- a/src/commands/kv/bucket/upload.rs
+++ b/src/commands/kv/bucket/upload.rs
@@ -9,24 +9,96 @@ use crate::settings::global_user::GlobalUser;
 use crate::settings::target::Target;
 use crate::terminal::message;
 
+const KEY_MAX_SIZE: usize = 512;
+const VALUE_MAX_SIZE: usize = 2 * 1024 * 1024;
+const PAIRS_MAX_COUNT: usize = 10000;
+const UPLOAD_MAX_SIZE: usize = 100 * 1024 * 1024;
+
 pub fn upload(
     target: &Target,
     user: GlobalUser,
     namespace_id: &str,
     path: &Path,
 ) -> Result<(), failure::Error> {
-    let pairs: Result<Vec<KeyValuePair>, failure::Error> = match &metadata(path) {
+    let mut pairs: Vec<KeyValuePair> = match &metadata(path) {
         Ok(file_type) if file_type.is_dir() => directory_keys_values(path),
         Ok(_file_type) => {
             // any other file types (files, symlinks)
             failure::bail!("wrangler kv:bucket upload takes a directory")
         }
         Err(e) => failure::bail!("{}", e),
-    };
+    }?;
 
-    match put_bulk(target, user, namespace_id, pairs?) {
-        Ok(_) => message::success("Success"),
-        Err(e) => print!("{}", e),
+    validate_file_uploads(pairs.clone())?;
+
+    // Create a vector of uploads; that is, a vector of vectors of key-value pairs, each of which are
+    // maximum 10K key-value pairs in size OR maximum 100MB in size.
+    let mut key_count = 0;
+    let mut key_pair_bytes = 0;
+    let mut key_value_batch: Vec<KeyValuePair> = Vec::new();
+
+    while !(pairs.is_empty() && key_value_batch.is_empty()) {
+        if pairs.is_empty() {
+            // Last batch to upload
+            call_put_bulk_api(target, user.clone(), namespace_id, &mut key_value_batch)?;
+        } else {
+            let pair = pairs.pop().unwrap();
+            if key_count + pair.key.len() > PAIRS_MAX_COUNT // Max KV pairs for request met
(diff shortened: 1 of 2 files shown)


src/commands/kv/bucket/upload.rs
```diff
@@ -23,5 +29,83 @@ pub fn upload(
...
+    let mut key_count = 0;
+    let mut key_pair_bytes = 0;
+    let mut key_value_batch: Vec<KeyValuePair> = Vec::new();
+
+    while !(pairs.is_empty() && key_value_batch.is_empty()) {
+        if pairs.is_empty() {
+            // Last batch to upload
+            call_put_bulk_api(target, user.clone(), namespace_id, &mut key_value_batch)?;
+        } else {
+            let pair = pairs.pop().unwrap();
```
gabbifish: Can't use `?` here because the error type from `pop()` doesn't work with this failure crate trait :(

ashleymichal: womp womp

src/commands/kv/bucket/upload.rs
```diff
@@ -9,24 +9,98 @@ use crate::settings::global_user::GlobalUser;
...
+            if key_count + pair.key.len() > PAIRS_MAX_COUNT
+            // Keep upload size small to keep KV bulk API happy
+            || key_pair_bytes + pair.key.len() + pair.value.len() > UPLOAD_MAX_SIZE * 1/2
+            {
+                call_put_bulk_api(target, user.clone(), namespace_id, &mut key_value_batch)?;
```
ashleymichal: what do we do when this errors?

src/commands/kv/bucket/upload.rs
```diff
@@ -9,24 +9,98 @@ use crate::settings::global_user::GlobalUser;
...
+}
+
+// Ensure that all key-value pairs being uploaded have valid sizes (this ensures that
+// no partial uploads happen). I don't like this function because it duplicates the
+// size checking the API already does--but doing a preemptive check like this (before
+// calling the API) will prevent partial bucket uploads from happening.
+pub fn validate_file_uploads(pairs: Vec<KeyValuePair>) -> Result<(), failure::Error> {
+    for pair in pairs {
+        if pair.key.len() > KEY_MAX_SIZE {
+            failure::bail!(
```
ashleymichal: should we bail immediately? or collect failures for the whole collection and show all?

src/commands/kv/bulk/put.rs
```diff
@@ -41,6 +41,7 @@ pub fn put(
     Ok(())
 }
 
+//todo(gabbi): Let's make sure to split very large payloads into multiple requests.
```
ashleymichal: this is this pr, no? 🙃

gabbifish: Hah! yes!

src/commands/kv/bucket/upload.rs
gabbifish: I think it's worth bailing immediately--this message should be clear enough to give users the memo of what they need to ensure.

src/commands/kv/bucket/upload.rs
gabbifish: the error is bubbled up and printed out using our default error handling logic

gabbifish: (i've seen this in action many times, especially when i was figuring out how to navigate timeouts!)

src/commands/kv/bucket/upload.rs
```diff
@@ -9,24 +9,98 @@ use crate::settings::global_user::GlobalUser;
...
+            {
+                call_put_bulk_api(target, user.clone(), namespace_id, &mut key_value_batch)?;
+
+                // If upload successful, reset counters
+                key_count = 0;
+                key_pair_bytes = 0;
+            }
+
+            // Add the popped key-value pair to the running batch of key-value pair uploads
+            key_count = key_count + pair.key.len();
```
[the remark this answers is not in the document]
gabbifish: this is a derp. This should definitiely be incrementing!!

src/commands/kv/bucket/upload.rs
```diff
@@ -9,24 +9,98 @@ use crate::settings::global_user::GlobalUser;
...
+
+    while !(pairs.is_empty() && key_value_batch.is_empty()) {
+        if pairs.is_empty() {
+            // Last batch to upload
+            call_put_bulk_api(target, user.clone(), namespace_id, &mut key_value_batch)?;
+        } else {
+            let pair = pairs.pop().unwrap();
+            if key_count + pair.key.len() > PAIRS_MAX_COUNT
+            // Keep upload size small to keep KV bulk API happy
+            || key_pair_bytes + pair.key.len() + pair.value.len() > UPLOAD_MAX_SIZE * 1/2
```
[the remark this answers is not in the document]
gabbifish: cargo fmt requires it :(

src/commands/kv/bucket/upload.rs
```diff
@@ -9,24 +9,98 @@ use crate::settings::global_user::GlobalUser;
...
+    let mut key_pair_bytes = 0;
+    let mut key_value_batch: Vec<KeyValuePair> = Vec::new();
+
+    while !(pairs.is_empty() && key_value_batch.is_empty()) {
+        if pairs.is_empty() {
+            // Last batch to upload
+            call_put_bulk_api(target, user.clone(), namespace_id, &mut key_value_batch)?;
+        } else {
+            let pair = pairs.pop().unwrap();
+            if key_count + pair.key.len() > PAIRS_MAX_COUNT
```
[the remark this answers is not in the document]
gabbifish: yeah, it should be plus one!

src/commands/kv/bucket/upload.rs
```diff
@@ -9,24 +9,98 @@ use crate::settings::global_user::GlobalUser;
 use crate::settings::target::Target;
 use crate::terminal::message;
 
+const KEY_MAX_SIZE: usize = 512;
+const VALUE_MAX_SIZE: usize = 2 * 1024 * 1024;
+const PAIRS_MAX_COUNT: usize = 10000;
```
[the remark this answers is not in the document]
gabbifish: gonna make it 5000 :)

src/commands/kv/bucket/upload.rs
ashleymichal: k. we can wait for the issue to be filed, but i can anticipate an impatient user wishing for an enumeration of their sins rather than finding them one at a time.

src/commands/kv/bucket/upload.rs
```diff
@@ -9,24 +9,98 @@ use crate::settings::global_user::GlobalUser;
...
+const UPLOAD_MAX_SIZE: usize = 100 * 1024 * 1024;
```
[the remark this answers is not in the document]
gabbifish: hmm, it's a question of whether we want to document the API's true max limits here... i like the idea of just making this halving part of the const, though.

src/commands/kv/bucket/upload.rs
ashleymichal: i guess more what i mean is, how do we rectify a partial upload? if we are hashing contents, i think we're mostly fine, but we might want to consider doing the `delete` part of `sync` _after_ the `upload` part of `sync` to prevent leaving users in a weird half state, as that will at least leave their old assets up.

gabbifish: Ah, this makes sense!

PR #613, opened 2019-09-17
Refactor kv::print_error to kv::format_error and hoist error printing out of private kv functions

### LINKED ISSUE #608: Remove success and error printouts from bucket KV API callers + refactor kv::print_error
Currently, all API functions perform a variant of the following:
```rust
    match response {
        Ok(_) => message::success("Success"),
        Err(e) => kv::print_error(e),
    }
```
This works for most API calls, but this poses a problem when functions like `kv:bucket upload` and `kv:bucket delete` are called; namely, we shouldn't print out the result of each KV API call when there are multiple, but instead just print out a final result after all KV API results are complete.

Theoretically, we could simply return a `Result<(), failure::Error>` from API calling functions and have the caller call `print_error()` on the result. The problem is that this result's error value is of type `failure::Error`; `print_error()` expects an `cloudflare::ApiFailure`, not a `failure::Error`.

I propose making `print_error()` a function that formats errors but doesn't actually print them out. Callers of print_error can wrap the output of `print_error()` in a `failure::bail!` to successfully still return `Result<(), failure::Error>` results and print them out.

gabbifish:
This PR closes #608. It will ensure that building bucket functionality on top of API calling functions is cleaner; that is, API calling functions that are private functions should not print the result of an API call, but its caller should.

### DIFF
--- a/src/commands/kv/mod.rs
+++ b/src/commands/kv/mod.rs
@@ -71,7 +71,6 @@ fn format_error(e: ApiFailure) -> String {
             give_status_code_context(status);
             let mut complete_err = "".to_string();
             for error in api_errors.errors {
-                // message::warn(&format!("Error {}: {}", error.code, error.message));
                 let error_msg =
                     format!("{} Error {}: {}\n", emoji::WARN, error.code, error.message);
 
--- a/src/commands/kv/mod.rs
+++ b/src/commands/kv/mod.rs
@@ -6,6 +6,7 @@ use percent_encoding::{percent_encode, PATH_SEGMENT_ENCODE_SET};
 
 use crate::settings::global_user::GlobalUser;
 use crate::settings::target::Target;
+use crate::terminal::emoji;
 use crate::terminal::message;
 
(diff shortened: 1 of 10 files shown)


src/commands/kv/mod.rs
```diff
@@ -64,20 +65,27 @@ fn api_client(user: GlobalUser) -> Result<HttpApiClient, failure::Error> {
...
 
-fn print_error(e: ApiFailure) {
+fn format_error(e: ApiFailure) -> String {
     match e {
         ApiFailure::Error(status, api_errors) => {
             give_status_code_context(status);
+            let mut complete_err = "".to_string();
             for error in api_errors.errors {
-                message::warn(&format!("Error {}: {}", error.code, error.message));
+                // message::warn(&format!("Error {}: {}", error.code, error.message));
```
ashleymichal: don't leave commented code.

gabbifish:
```suggestion
```

PR #614, opened 2019-09-18
Implement crude version of `kv:bucket sync` that does not use hashes to determine changed files yet

### LINKED ISSUE #484: Add command for syncing a "bucket"
per https://github.com/cloudflare/wrangler/issues/459, we want to separate the use case where a user specifies a directory for upload with paths as keys into its own subclass of commands.

There should be one command (or at least the logic for publish to use) for syncing the contents of a directory up to the specified namespace; this includes clearing keys that no longer exist in the local directory.

gabbifish:
This PR implements a crude version of #484. (Fixes #484).

This implementation does two things:
1. It deletes all keys/filenames in the remote KV namespace whose keys/filenames no longer exist in the given local directory.
2. It uploads all files in the given local directory.

Ideally, we only update files in the remote directory if we know they have changed (e.g. do a hash comparison). We will wait for manifest work to be complete before we move to that step.

Usage cases:
When stale files (exist in remote but not in local) are deleted:
```console
$ wrangler kv:bucket sync -b=BUTTS abc/
 Deleting stale files...
 Preparing to upload updated files...
 Parsing a.txt...
 Parsing b/c.txt...
 Success
```

When no stale files exist:
```console
$ wrangler kv:bucket sync -b=BUTTS abc/
 Preparing to upload updated files...
 Parsing a.txt...
 Parsing b/c.txt...
 Success
```

### DIFF
--- a/src/commands/kv/bucket/sync.rs
+++ b/src/commands/kv/bucket/sync.rs
@@ -19,12 +19,11 @@ pub fn sync(
     path: &Path,
 ) -> Result<(), failure::Error> {
     // First get local keys
-    let local_keys_vec: Result<Vec<String>, failure::Error> = match &metadata(path) {
+    let local_keys_vec: Vec<String> = match &metadata(path) {
         Ok(file_type) if file_type.is_dir() => directory_keys_only(path),
         Ok(_) => failure::bail!("{} should be a directory", path.display()),
         Err(e) => failure::bail!("{}", e),
-    };
-    let local_keys_vec = local_keys_vec?;
+    }?;
     let local_keys: HashSet<_> = HashSet::from_iter(local_keys_vec.iter());
 
     // Then get remote keys
--- a/src/commands/kv/bucket/sync.rs
+++ b/src/commands/kv/bucket/sync.rs
@@ -1,0 +1,62 @@
+use std::collections::HashSet;
+use std::fs::metadata;
+use std::iter::FromIterator;
+use std::path::Path;
+
+use crate::commands::kv;
+use crate::commands::kv::bucket::directory_keys_only;
+use crate::commands::kv::bucket::upload::upload;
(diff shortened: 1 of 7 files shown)


src/commands/kv/bucket/sync.rs
```diff
@@ -0,0 +1,62 @@
...
+use crate::terminal::message;
+
+pub fn sync(
+    target: &Target,
+    user: GlobalUser,
+    namespace_id: &str,
+    path: &Path,
+) -> Result<(), failure::Error> {
+    // First get local keys
+    let local_keys_vec: Result<Vec<String>, failure::Error> = match &metadata(path) {
```
ashleymichal: You can actually just `?` this entire expression and get the Vec directly. (suggestion spread across three comments bc this interface does not support multiline suggestions yet)
```suggestion
    let local_keys_vec: Vec<String> = match &metadata(path) {
```

src/commands/kv/bucket/sync.rs
```diff
@@ -0,0 +1,62 @@
...
+        Ok(file_type) if file_type.is_dir() => directory_keys_only(path),
+        Ok(_) => failure::bail!("{} should be a directory", path.display()),
+        Err(e) => failure::bail!("{}", e),
+    };
```
ashleymichal:
```suggestion
    }?;
```

src/commands/kv/bucket/sync.rs
```diff
@@ -0,0 +1,62 @@
...
+    let local_keys_vec = local_keys_vec?;
```
ashleymichal:
```suggestion

```

src/commands/kv/bucket/sync.rs
gabbifish: Heck yeah!

PR #615, opened 2019-09-18
Alewis/sites interface

ashleymichal:
This PR establishes a new optional configuration in the Wrangler.toml, `site`, which automatically provisions a "bucket" type KV namespace, and publishes to it on `wrangler publish`. Not yet implemented: generating/building the worker code to serve sites.

### DIFF
--- a/src/commands/kv/namespace/site.rs
+++ b/src/commands/kv/namespace/site.rs
@@ -1,0 +1,60 @@
+use cloudflare::endpoints::workerskv::create_namespace::CreateNamespace;
+use cloudflare::endpoints::workerskv::create_namespace::CreateNamespaceParams;
+use cloudflare::endpoints::workerskv::list_namespaces::ListNamespaces;
+use cloudflare::endpoints::workerskv::WorkersKvNamespace;
+use cloudflare::framework::apiclient::ApiClient;
+use cloudflare::framework::response::ApiFailure;
+
+use crate::commands::kv;
+use crate::settings::global_user::GlobalUser;
+use crate::settings::target::Target;
+use crate::terminal::message;
+
+pub fn site(target: &Target, user: &GlobalUser) -> Result<WorkersKvNamespace, failure::Error> {
+    let client = kv::api_client(user.to_owned())?;
+
+    let title = format!("__{}-{}", target.name, "workers_sites_assets");
+    let msg = format!("Creating namespace for Workers Site \"{}\"", title);
+    message::working(&msg);
+
+    let response = client.request(&CreateNamespace {
+        account_identifier: &target.account_id,
+        params: CreateNamespaceParams {
+            title: title.to_owned(),
+        },
+    });
+
+    match response {
+        Ok(success) => {
+            return Ok(success.result);
+        }
+        Err(e) => match e {
+            ApiFailure::Error(_status, api_errors) => {
+                if api_errors
+                    .errors
+                    .iter()
+                    .find(|&e| e.code == 10014)
+                    .is_some()
+                {
+                    log::info!("Namespace {} already exists.", title);
+                    let response = client.request(&ListNamespaces {
+                        account_identifier: &target.account_id,
+                    });
+
+                    match response {
+                        Ok(success) => Ok(success
+                            .result
+                            .iter()
+                            .find(|ns| ns.title == title)
+                            .unwrap()
+                            .to_owned()),
+                        Err(e) => failure::bail!("{:?}", e),
+                    }
+                } else {
+                    failure::bail!("{:?}", api_errors.errors)
+                }
+            }
+            ApiFailure::Invalid(reqwest_err) => failure::bail!("Error: {}", reqwest_err),
+        },
+    }
+}
--- a/src/commands/publish/mod.rs
+++ b/src/commands/publish/mod.rs
@@ -4,6 +4,7 @@ pub mod preview;
 mod route;
 mod upload_form;
 
+use crate::settings::target::kv_namespace::KvNamespace;
 pub use package::Package;
 use route::Route;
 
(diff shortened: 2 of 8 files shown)


src/commands/publish/upload_form/mod.rs
```diff
@@ -1,14 +1,13 @@
 mod project_assets;
 mod wasm_module;
 
-use log::info;
```
ashleymichal: cleanup consistent with changes made elsewhere in the code.

src/commands/publish/preview/upload.rs
```diff
@@ -109,7 +109,7 @@ fn authenticated_upload(client: &Client, target: &Target) -> Result<Preview, fai
     );
     log::info!("address: {}", create_address);
 
-    let script_upload_form = publish::build_script_upload_form(&target)?;
+    let script_upload_form = publish::build_script_upload_form(target)?;
```
ashleymichal: this is already a reference, so removing the reference-to-a-reference

src/commands/publish/mod.rs
```diff
@@ -17,18 +18,30 @@ use crate::commands::kv;
...
+
+    if let Some(site_config) = &target.site {
+        let site_namespace = kv::namespace::site(target, user)?;
+
+        target.add_kv_namespace(KvNamespace {
+            binding: "__STATIC_CONTENT".to_string(),
+            id: site_namespace.id,
+            bucket: Some(site_config.bucket.to_owned()),
+        });
+    }
```
ashleymichal: considered stuffing all of this logic in the kv_namespaces method on target, but because there is an api call, we need the global user, and it feels wrong to add that dependency to target. since we don't re-use clients in this function, there is also a world in which we initialize the user and the client inside the kv::namespace::site function, and I think that might be OK, but I think I want to wait until I implement this for preview.

src/commands/publish/mod.rs
```diff
@@ -17,18 +18,30 @@ use crate::commands::kv;
...
 use crate::http;
 use crate::settings::global_user::GlobalUser;
+
 use crate::settings::target::Target;
 use crate::terminal::{emoji, message};
 
 pub fn publish(
     user: &GlobalUser,
-    target: &Target,
+    target: &mut Target,
```
ashleymichal: we need to ask the target to update its in-memory kv_namespaces in the case that sites is specified. we do this here because we want to delay this logic to preview (implementing later) and publish, rather than requiring it for every command.

src/commands/kv/namespace/site.rs
```diff
@@ -0,0 +1,60 @@
...
+use cloudflare::endpoints::workerskv::WorkersKvNamespace;
+use cloudflare::framework::apiclient::ApiClient;
+use cloudflare::framework::response::ApiFailure;
+
+use crate::commands::kv;
+use crate::settings::global_user::GlobalUser;
+use crate::settings::target::Target;
+use crate::terminal::message;
+
+pub fn site(target: &Target, user: &GlobalUser) -> Result<WorkersKvNamespace, failure::Error> {
```
ashleymichal: the purpose of this function is to create a specially-named namespace for storing the static assets for a particular worker. tl;dr - try to create a namespace with a title specified by convention. if it already exists, get its id by listing namespaces and finding the one with the matching title. return the namespace as a result.

src/main.rs
```diff
@@ -494,7 +494,7 @@ fn run() -> Result<(), failure::Error> {
             failure::bail!("You can only pass --env or --release, not both")
         }
         let manifest = settings::target::Manifest::new(config_path)?;
-        let target;
+        let mut target;
```
ashleymichal: target must be mutable in order to update its kv-namespaces for publish. could alternatively write a separate constructor that did this inside target, but for now this works Just Fine.

src/commands/kv/namespace/site.rs
```diff
@@ -0,0 +1,60 @@
...
+    let client = kv::api_client(user.to_owned())?;
+
+    let title = format!("__{}-{}", target.name, "workers_sites_assets");
```
gabbifish: Why do we use `__`?

src/commands/kv/namespace/site.rs
```diff
@@ -0,0 +1,60 @@
...
+    let msg = format!("Creating namespace for Workers Site \"{}\"", title);
+    message::working(&msg);
+
+    let response = client.request(&CreateNamespace {
```
gabbifish: Couldn't we use `kv::namespace::create::create()?` I think this would reduce a lot of redundancy.

src/commands/kv/namespace/site.rs
```diff
@@ -0,0 +1,60 @@
...
+        Err(e) => match e {
+            ApiFailure::Error(_status, api_errors) => {
+                if api_errors
+                    .errors
+                    .iter()
+                    .find(|&e| e.code == 10014)
+                    .is_some()
+                {
+                    log::info!("Namespace {} already exists.", title);
+                    let response = client.request(&ListNamespaces {
```
gabbifish: I think it's worth refactoring list() and create() so that the API call happens in a separate (private) function, then two public functions call to it (the default `kv:namespace` one, and then this function for sites!)

src/commands/publish/mod.rs
gabbifish: do we write this change back out to wrangler.toml?

src/commands/publish/mod.rs
gabbifish: ^ this (putting user and client logic into kv::namespace::site) sounds appropriate!

src/commands/publish/mod.rs
ashleymichal: no.

src/commands/kv/namespace/site.rs
ashleymichal: to sort of call it out as an auto-generated namespace, and sort of prevent collisions. these titles will still be visible in the UI.

src/commands/kv/namespace/site.rs
ashleymichal: i'd be stoked to make refactors later.

src/commands/kv/namespace/site.rs
ashleymichal: i'd be stoked to make refactors later.

src/commands/kv/namespace/site.rs
ashleymichal: i think it is too; i'm happy to do this later this week or early next week once we've cut a release candidate.

src/commands/publish/mod.rs
gabbifish: cool, this makes sense!

src/commands/kv/namespace/site.rs
gabbifish: this would be an awesome comment to put in code in the future!

ashleymichal: putting it there now

PR #626, opened 2019-09-19
Add verbose logging option for `kv:bucket sync/upload`

### LINKED ISSUE #625: Add `--verbose` flag for KV bucket commands
Listing out every file that is being uploaded is useful in some cases, but is typically quite verbose. We should make printouts of bucket action information something that is toggled on by a `--verbose` flag.

gabbifish:
This PR closes #625.

### DIFF
--- a/src/commands/kv/bucket/upload.rs
+++ b/src/commands/kv/bucket/upload.rs
@@ -22,9 +22,25 @@ pub fn upload(
     user: GlobalUser,
     namespace_id: &str,
     path: &Path,
+    verbose: bool,
+) -> Result<(), failure::Error> {
+    match upload_files(target, user, namespace_id, path, verbose) {
+        Ok(_) => message::success("Success"),
+        Err(e) => print!("{}", e),
+    }
+
+    Ok(())
+}
+
+pub fn upload_files(
+    target: &Target,
+    user: GlobalUser,
+    namespace_id: &str,
+    path: &Path,
+    verbose: bool,
 ) -> Result<(), failure::Error> {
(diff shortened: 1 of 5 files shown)


src/commands/kv/bucket/upload.rs
```diff
@@ -22,9 +22,25 @@ pub fn upload(
     user: GlobalUser,
     namespace_id: &str,
     path: &Path,
+    verbose: bool,
+) -> Result<(), failure::Error> {
+    match upload_files(target, user, namespace_id, path, verbose) {
```
gabbifish: I have `upload()` wrap `upload_files()` to ensure that callers of `upload_files()` (like `sync()`) don't have Success messages printed out for them (this would lead to duplicative Success messages). `upload()` handles this printing instead.

src/main.rs
```diff
@@ -594,19 +605,10 @@ fn run() -> Result<(), failure::Error> {
...
                 let key = put_key_matches.value_of("key").unwrap();
-                let is_file = match put_key_matches.occurrences_of("path") {
-                    1 => true,
-                    _ => false,
-                };
 
                 // If is_file is true, overwrite value to be the contents of the given
                 // filename in the 'value' arg.
-                let value_arg = put_key_matches.value_of("value").unwrap();
-                let value = if is_file {
```
gabbifish: This logic duplicates that already in `commands::kv::key::put`; I have removed it here and tested that individual file upload still works as expected.

ashleymichal: i think this is the result of a merge problem, because i think i removed is_file from the put command to pull it up to main, but another commit put it back.

PR #628, opened 2019-09-19
Alewis/generate site

ashleymichal:
Add a flag `--site` with special logic for which template to generate.

### DIFF
--- a/src/settings/target/mod.rs
+++ b/src/settings/target/mod.rs
@@ -123,7 +123,7 @@ impl Manifest {
     pub fn generate(
         name: String,
         target_type: TargetType,
-        init: bool,
+        config_path: PathBuf,
     ) -> Result<Manifest, failure::Error> {
         let manifest = Manifest {
             account_id: String::new(),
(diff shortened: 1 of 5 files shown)


src/settings/target/mod.rs
```diff
@@ -120,6 +120,113 @@ impl Manifest {
...
+
+        let toml = toml::to_string(&manifest)?;
+        let config_file = config_path.join("wrangler.toml");
+
+        log::info!("Writing a wrangler.toml file at {}", config_file.display());
+        fs::write(&config_file, &toml)?;
+        Ok(manifest)
+    }
+
+    pub fn get_target(
```
ashleymichal: it looks like i did things to these methods. I did not, there is a refactor commit that simply moves the public function declarations to the top of the Manifest impl

src/settings/target/mod.rs
```diff
@@ -120,6 +120,113 @@ impl Manifest {
         Ok(manifest)
     }
 
+    pub fn generate(
```
ashleymichal: the change here is a refactor to use `Manifest::default()` and move this `pub fn` declaration to the top.

ashleymichal:
closing in favor of a larger PR sorrynotsorry

PR #629, opened 2019-09-19
Asset Manifest

EverlastingBugstopper:
This PR adds support for uploading text blobs and also an asset manifest that maps paths to content w/their sha-256 hash

### DIFF
--- a/src/commands/publish/upload_form/mod.rs
+++ b/src/commands/publish/upload_form/mod.rs
@@ -85,7 +85,7 @@ pub fn build_script_upload_form(target: &Target) -> Result<Form, failure::Error>
     }
 }
 
-fn get_asset_manifest(directory: &String) -> Result<String, failure::Error> {
+fn get_asset_manifest(directory: &str) -> Result<String, failure::Error> {
     let directory = Path::new(&directory);
     let (_, manifest) = directory_keys_values(directory)?;
     let manifest = serde_json::to_string(&manifest)?;
--- a/src/commands/publish/upload_form/mod.rs
+++ b/src/commands/publish/upload_form/mod.rs
@@ -1,4 +1,5 @@
 mod project_assets;
+mod text_blob;
 mod wasm_module;
 
 use reqwest::multipart::{Form, Part};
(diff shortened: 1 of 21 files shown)


src/commands/publish/upload_form/mod.rs
```diff
@@ -63,16 +66,32 @@ pub fn build_script_upload_form(target: &Target) -> Result<Form, failure::Error>
...
                 let wasm_module = WasmModule::new(path, binding)?;
-                wasm_modules.push(wasm_module)
+                wasm_modules.push(wasm_module);
             }
 
-            let assets = ProjectAssets::new(script_path, wasm_modules, kv_namespaces)?;
+            let mut text_blobs = Vec::new();
+
+            if let Some(site) = &target.site {
+                let binding = "__STATIC_CONTENT_MANFIEST".to_string();
```
gabbifish: I prefer making this a const, if possible :)

src/commands/publish/upload_form/mod.rs
```diff
@@ -45,7 +48,7 @@ pub fn build_script_upload_form(target: &Target) -> Result<Form, failure::Error>
 
             let script_path = package.main()?;
 
-            let assets = ProjectAssets::new(script_path, Vec::new(), kv_namespaces)?;
+            let assets = ProjectAssets::new(script_path, Vec::new(), kv_namespaces, Vec::new())?;
```
gabbifish: IT looks like if this vector is empty, it is never used--think we can make this an `Option<>` instead of an empty vector?

src/commands/publish/upload_form/mod.rs
EverlastingBugstopper: calls in constants are limited to constant functions, tuple structs and tuple variants

EverlastingBugstopper: made with <3 from the compiler

src/commands/publish/upload_form/mod.rs
EverlastingBugstopper: cool

PR #632, opened 2019-09-19
Add support for generating and initializing site projects

ashleymichal:
# `wrangler generate --site`

``` sh
# creates a new directory structure from an external repo (currently not the real thing)
$ wrangler generate --site
$ cd worker
$ tree
.
├── README.md
├── public
│   └── index.html
├── workers-site
│   ├── index.js
│   ├── package-lock.json
│   └── package.json
└── wrangler.toml
# update the wrangler.toml here to have bucket = "./public"
$ wrangler publish
```

# `wrangler init --site`

inside an existing project, run 

```sh 
$ wrangler init --site
tree (in addition to your existing project)
.
├── workers-site
│   ├── index.js
│   ├── package-lock.json
│   └── package.json
└── wrangler.toml
# update the wrangler.toml with your build asset folder
$ wrangler publish
```

### DIFF
--- a/src/commands/build/wranglerjs/mod.rs
+++ b/src/commands/build/wranglerjs/mod.rs
@@ -201,6 +201,8 @@ pub fn scaffold_site_worker(target: &Target) -> Result<(), failure::Error> {
                 &target.target_type,
             )?;
 
+            // This step is to prevent having a git repo within a git repo after
+            // generating the scaffold into an existing project.
             fs::remove_dir_all(&build_dir.join(".git"))?;
         }
     }
--- a/src/commands/build/wranglerjs/mod.rs
+++ b/src/commands/build/wranglerjs/mod.rs
@@ -130,7 +130,9 @@ fn setup_build(target: &Target) -> Result<(Command, PathBuf, Bundle), failure::E
 
     let build_dir = target.build_dir()?;
 
-    scaffold_site_worker(&target)?;
+    if target.site.is_some() {
+        scaffold_site_worker(&target)?;
+    }
 
     run_npm_install(&build_dir).expect("could not run `npm install`");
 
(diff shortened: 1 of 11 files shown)


gabbifish:
note to self: I think it would be super important to highlight that folks bringing their own site generators should `wrangler init --site` and to print out a message saying something along the lines of "Hello! please change your bucket value in wrangler.toml to the output directory of your static site generator."

src/commands/generate.rs
```diff
@@ -9,18 +9,36 @@ pub fn generate(
...
+    if site {
+        run_generate(name, template, &target_type)?;
+        let config_path = PathBuf::from("./").join(&name);
+        Manifest::generate(name.to_string(), target_type, config_path, site)?;
+        Ok(())
+    } else {
+        run_generate(name, template, &target_type)?;
+        let config_path = PathBuf::from("./").join(&name);
+        Manifest::generate(name.to_string(), target_type, config_path, site)?;
+        Ok(())
```
gabbifish: maybe i'm silly... but why don't we scaffold a site here? I would have thought the scaffolding would happen in `generate` and not `init` (since I thought `init` just introduces a `wrangler.toml`)

src/main.rs
```diff
@@ -319,6 +320,13 @@ fn run() -> Result<(), failure::Error> {
                         .long("type")
                         .takes_value(true)
                         .help("the type of project you want generated"),
+                )
+                .arg(
+                    Arg::with_name("site")
+                        .short("s")
+                        .long("site")
+                        .takes_value(false)
+                        .help("initializes a Workers Sites project overrides `type` and `template`"),
```
gabbifish: small nit: 
```suggestion
                        .help("initializes a Workers Sites project. Overrides `type` and `template`"),
```

src/main.rs
```diff
@@ -338,6 +346,13 @@ fn run() -> Result<(), failure::Error> {
...
```
gabbifish: same suggestion as above

src/settings/target/mod.rs
```diff
@@ -22,6 +22,15 @@ pub struct Site {
     pub entry_point: Option<String>,
 }
 
+impl Default for Site {
+    fn default() -> Site {
+        Site {
+            bucket: String::new(),
+            entry_point: Some(String::from("workers-site")),
```
gabbifish: nit: this would be a lovely const!

src/settings/target/mod.rs
```diff
@@ -120,6 +129,106 @@ impl Manifest {
         Ok(manifest)
     }
 
+    pub fn generate(
+        name: String,
+        target_type: TargetType,
+        config_path: PathBuf,
+        site: bool,
+    ) -> Result<Manifest, failure::Error> {
+        let mut manifest = Manifest::default();
```
gabbifish: Why did we get rid of the nice struct init logic from before?

src/commands/generate.rs
```diff
@@ -9,18 +9,36 @@ pub fn generate(
...
+    site: bool,
+) -> Result<(), failure::Error> {
+    let target_type = target_type.unwrap_or_else(|| get_target_type(template));
+    if site {
+        run_generate(name, template, &target_type)?;
+        let config_path = PathBuf::from("./").join(&name);
+        Manifest::generate(name.to_string(), target_type, config_path, site)?;
+        Ok(())
+    } else {
+        run_generate(name, template, &target_type)?;
```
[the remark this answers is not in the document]
ashleymichal: clippy found this.

src/commands/generate.rs
ashleymichal: it's a bit wacky BUT, scaffold only puts the contents of the workers-site directory (and the directory itself). generate does a full project directory from a github repo, and doesn't remove the git project.

src/settings/target/mod.rs
ashleymichal: idk it sounded like a good idea to use default but i'm not using it well here.

src/commands/generate.rs
gabbifish: gotcha, this makes more sense now. let me make sure wrangler generate also calls scaffolding function.

ashleymichal: i think there's a better way to do it, but at the moment this is kinda what i got

src/settings/target/mod.rs
gabbifish: this can live for now in the interest of getting RC out!!

src/commands/build/wranglerjs/mod.rs
```diff
@@ -182,6 +188,27 @@ fn setup_build(target: &Target) -> Result<(Command, PathBuf, Bundle), failure::E
     Ok((command, temp_file, bundle))
 }
 
+pub fn scaffold_site_worker(target: &Target) -> Result<(), failure::Error> {
```
gabbifish: omg this is still killing me. do we just go ahead with this?

PR #633, opened 2019-09-19
Better file support for kv:put

phayes:
This PR adds better file support for `kv:put`

1. Allows non-utf8 files to be used as values
2. Streams the file directly from disk to the endpoint, instead of buffering them in memory first.

### DIFF
--- a/src/commands/kv/key/put.rs
+++ b/src/commands/kv/key/put.rs
@@ -41,11 +41,19 @@ pub fn put(
     };
     let url = Url::parse_with_params(&api_endpoint, query_params);
 
+    let client = http::auth_client(&user);
+
+    let url_into_str = url?.into_string();
+
     // If is_file is true, overwrite value to be the contents of the given
     // filename in the 'value' arg.
-    let body_text = if is_file {
+    let mut res;
+    if is_file {
         match &metadata(value) {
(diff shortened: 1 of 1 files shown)


gabbifish:
@phayes Thank you for this PR! This is fantastic.

src/commands/kv/key/put.rs
```diff
@@ -41,26 +41,28 @@ pub fn put(
...
     let url = Url::parse_with_params(&api_endpoint, query_params);
 
+    let client = http::auth_client(&user);
+
+    let url_into_str = url?.into_string();
+
     // If is_file is true, overwrite value to be the contents of the given
     // filename in the 'value' arg.
-    let body_text = if is_file {
+    let mut res = if is_file {
```
gabbifish: I don't think this needs to be mutable?

phayes: Weirdly, it does need to be mutable. 

On line 73 we have `let parsed = res.json();`, which needs the response to be mutable.   The signature of this method is:

```
pub fn json<T: DeserializeOwned>(&mut self) -> ::Result<T>
```

I think this is because when we return the response, the body hasn't been read yet.  And if we have an error then we need to read the body, which mutates the response as we're reading it in.

gabbifish: This makes sense (and TIL)!

PR #651, opened 2019-09-20
Update upload.rs

### LINKED ISSUE #650: include actual size in "value too big" error
KV has a maximum value size of 2MB that we check for on uploads. When a user's value has exceeded this size, we should print the actual size.

stevenfranks:
Fixes #650 

Added actual KV size to upload error message

### DIFF
--- a/src/commands/kv/bucket/upload.rs
+++ b/src/commands/kv/bucket/upload.rs
@@ -110,15 +110,17 @@ pub fn validate_file_uploads(pairs: Vec<KeyValuePair>) -> Result<(), failure::Er
     for pair in pairs {
         if pair.key.len() > KEY_MAX_SIZE {
             failure::bail!(
-                "Path `{}` exceeds the maximum key size limit of {} bytes",
+                "Path `{}` of {} bytes exceeds the maximum key size limit of {} bytes",
                 pair.key,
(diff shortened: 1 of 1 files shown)


src/commands/kv/bucket/upload.rs
```diff
@@ -110,15 +110,17 @@ pub fn validate_file_uploads(pairs: Vec<KeyValuePair>) -> Result<(), failure::Er
...
+                pair.key.len(),
                 KEY_MAX_SIZE
             );
         }
         if pair.value.len() > VALUE_MAX_SIZE {
             failure::bail!(
-                "File `{}` exceeds the maximum value size limit of {} bytes",
+                "File `{}` of {} bytes exceeds the maximum value size limit of {} bytes",
                 pair.key,
+                pair.key.len(),
```
ashleymichal:
```suggestion
                pair.value.len(),
```

ashleymichal:
thanks for this PR @stevenfranks !! so prompt! just one small change requested.

stevenfranks:
> thanks for this PR @stevenfranks !! so prompt! just one small change requested.

Not a problem, more than happy to help! Thank you for the suggestion, this has now been commited.

PR #654, opened 2019-09-20
Alewis/remove buckets

### LINKED ISSUE #634: Defer `kv:bucket` subcommands for later release
The functionality we've created around using kv as a bucket is tailored very specifically to the `site` use case. we can generalize this later, but for now we should remove the commands as they are not stable enough for release.

ashleymichal:
closes #634 . also removes `--no-worker` and `--no-bucket` flags from publish.


src/main.rs
```diff
@@ -705,57 +691,6 @@ fn run() -> Result<(), failure::Error> {
             ("", None) => message::warn("kv:bulk expects a subcommand"),
             _ => unreachable!(),
         }
-    } else if let Some(kv_matches) = matches.subcommand_matches("kv:bucket") {
```
gabbifish: Need to remove all bucket logic in the Clap::App declaration above as well

ashleymichal: did i not? lame.

gabbifish: Just did for ya!

ashleymichal: oh man you confused to 💩 out of me. thanks!

PR #664, opened 2019-09-23
Sync that only uploads files if their contents have changed from the corresponding file in remote

### LINKED ISSUE #656: Ensure bucket sync only uploads files that have changed
Now that asset manifests have landed, we are in a place to implement the last component of Wrangler file syncing to Workers KV: only uploading files if they have changed. We do this by generating the hash of a file and seeing if the file exists in the remote with that hash; if so, do not upload the file again (the remote version is up-to-date).

gabbifish:
This PR closes #656. It contains a unit test to ensure that the filtering logic for changed files works as expected. 

I have also tested this new upload logic with my own site https://blog-test-again.ziggy.workers.dev/.


src/commands/kv/bucket/sync.rs
```diff
@@ -19,11 +19,33 @@ pub fn sync(
     path: &Path,
     verbose: bool,
 ) -> Result<(), failure::Error> {
-    // First, upload all existing files in given directory
+    // First, upload all changed files in given local directory (aka file in Workers KV
+    // is now stale).
+
+    // Gremote keys, which contain the hash of the file (value) as the suffix.
```
gabbifish:
```suggestion
    // Get remote keys, which contain the hash of the file (value) as the suffix.
```

src/commands/kv/bucket/sync.rs
```diff
@@ -19,11 +19,35 @@ pub fn sync(
...
     if verbose {
         message::info("Preparing to upload updated files...");
     }
-    upload_files(target, user.clone(), namespace_id, path, verbose)?;
+    upload_files(
+        target,
+        user.clone(),
+        namespace_id,
+        path,
+        Some(remote_keys.clone()),
```
ashleymichal: can this not take a reference?

gabbifish: It can! doing a few edits to the function signatures so that they take more memory-efficient references rather than copy :)

signalnerve:
Hey @gabbifish, does this update something existing with Wrangler or is it just a pre-1.4.0 update? Wondering if this should make it into the CHANGELOG for 1.4.0.

gabbifish:
This is a pre-1.4.0 update—feel free to exclude it from changelog!

PR #665, opened 2019-09-23
Ensure KV subcommands check for presence of required fields in wrangler.toml

### LINKED ISSUE #607: Provide better messaging when account_id is missing
When the api gateway receives an invalid account tag, it returns a rather cryptic error:

``` terminal
Error 7003: Could not route to /accounts/storage/kv/namespaces/63ab5438e79d4678bb16262d4b7f5f79/keys,
```
we should catch code 7003 and return a helpful message suggesting the user check their account id.

gabbifish:
This PR closes #607.

UPDATE: this PR is no longer WIP; we do not need to validate account_id anymore (scope of ticket has changed as per maintainer meeting 09.23.19).

previously:
It is still considered a WIP given that this only checks for the absence of required fields like account_id for KV subcommands; it does not implement things like account_id validity checking. Do we still want to implement account_id validity checking? I am still keen to get this logic in before 1.4.0 because I don't want KV users to get confused about the error messages referenced in #607; so, I'm okay without account_id validity checking for now.

### DIFF
--- a/src/commands/kv/mod.rs
+++ b/src/commands/kv/mod.rs
@@ -148,21 +148,6 @@ pub fn validate_target(target: &Target) -> Result<(), failure::Error> {
         missing_fields.push("account_id")
     };
 
-    match &target.kv_namespaces {
-        Some(kv_namespaces) => {
-            for kv in kv_namespaces {
-                if kv.binding.is_empty() {
-                    missing_fields.push("kv-namespace binding")
-                }
-
-                if kv.id.is_empty() {
-                    missing_fields.push("kv-namespace id")
-                }
-            }
-        }
-        None => {}
-    }
-
     if !missing_fields.is_empty() {
         failure::bail!(
             "Your wrangler.toml is missing the following field(s): {:?}",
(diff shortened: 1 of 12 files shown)


src/commands/kv/mod.rs
```diff
@@ -138,6 +141,41 @@ fn help(error_code: u16) -> &'static str {
...
 }
 
+pub fn validate_target(target: &Target) -> Result<(), failure::Error> {
+    let mut missing_fields = Vec::new();
+
+    if target.account_id.is_empty() {
+        missing_fields.push("account_id")
+    };
+    if target.name.is_empty() {
+        missing_fields.push("name")
```
ashleymichal: idk if this makes sense here.

gabbifish: this makes sense. Happy to remove this!

gabbifish: oh wait, isn't `name` required when you create a KV namespace?

src/commands/kv/mod.rs
```diff
@@ -121,6 +121,9 @@ fn give_status_code_context(status_code: StatusCode) {
 fn help(error_code: u16) -> &'static str {
     // https://api.cloudflare.com/#workers-kv-namespace-errors
     match error_code {
+        7003 | 7000 => {
+            "Your wrangler.toml is likely missing the field \"account_id\", which is required to write to Workers KV."
```
ashleymichal: i think the other situation this applies to is if email or api key is invalid.

PR #682, opened 2019-09-23
Download releases from our proxy rather than from GitHub

zackbloom:
Performance 📈
403s 📉

### DIFF
--- a/npm/install-wrangler.js
+++ b/npm/install-wrangler.js
@@ -10,11 +10,7 @@ const { homedir } = require('os');
 
 const cwd = join(homedir(), ".wrangler");
 
-function getReleaseByTag(tag) {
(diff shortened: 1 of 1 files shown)


zackbloom:
@ashleygwilliams, et. al.: What would you like me to do with the source of the Worker which powers https://cloudflare.works/download-wrangler? Should it be placed here, or in a new repo?

ashleygwilliams:
@zackbloom i think we'd want it in here! trying to think where the right place would be,  im imagining a restructuring where there is an installer dir, with worker and npm inside it? curious what other folks think

EverlastingBugstopper:
I think this makes sense. We should use wrangler to publish the proxy worker 😛

PR #684, opened 2019-09-23
Add error messages for improperly configured buckets

### LINKED ISSUE #644: improve error / help message when bucket not configured
right now if a user fails to configure a site.bucket, they just get a cryptic OS does not exist error. wrangler should suggest to them that they configure their bucket.

EverlastingBugstopper:
This PR checks if `bucket` is a valid `path` and gives the user an actionable error message if it is not :)

```console
$ wrangler publish                     
🌀  Creating namespace for Workers Site "__my-new-site-workers_sites_assets"
Error: ⚠️  You need to specify a bucket directory in your wrangler.toml                                                                                                                                                                                       
$ wrangler publish
🌀  Creating namespace for Workers Site "__my-new-site-workers_sites_assets"
Error: ⚠️  directory for bucket does not exist at "./no"                                                                                                                                                                                   
$ wrangler publish
🌀  Creating namespace for Workers Site "__my-new-site-workers_sites_assets"
Error: ⚠️  bucket "./public/index.html" is not a directory
```

Fixes #644

### DIFF
--- a/src/commands/publish/mod.rs
+++ b/src/commands/publish/mod.rs
@@ -98,7 +98,7 @@ fn upload_buckets(target: &Target, user: &GlobalUser) -> Result<(), failure::Err
             let path = Path::new(&bucket);
             if !path.exists() {
                 failure::bail!(
-                    "{} directory for bucket does not exist at \"{}\"",
+                    "{} bucket directory \"{}\" does not exist",
                     emoji::WARN,
                     path.to_string_lossy()
                 )
--- a/src/commands/publish/mod.rs
+++ b/src/commands/publish/mod.rs
@@ -99,12 +99,19 @@ fn upload_buckets(target: &Target, user: &GlobalUser) -> Result<(), failure::Err
             if !path.exists() {
                 if let Some(path_str) = path.to_str() {
                     failure::bail!(
-                        "{} could not find a directory for bucket \"{}\"",
+                        "{} directory for bucket does not exist at \"{}\"",
                         emoji::WARN,
                         path_str
                     )
+                } else {
+                    failure::bail!("{} bucket does not exist")
                 }
-                failure::bail!("{} bucket path is not valid UTF-8")
+            } else if !path.is_dir() {
+                if let Some(path_str) = path.to_str() {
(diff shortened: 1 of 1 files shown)


src/commands/publish/mod.rs
```diff
@@ -89,7 +89,23 @@ fn publish_script(user: &GlobalUser, target: &Target) -> Result<(), failure::Err
...
+                    "{} You need to specify a bucket directory in your wrangler.toml",
+                    emoji::WARN
+                )
+            }
             let path = Path::new(&bucket);
+            if !path.exists() {
+                if let Some(path_str) = path.to_str() {
+                    failure::bail!(
+                        "{} could not find a directory for bucket \"{}\"",
+                        emoji::WARN,
```
victoriabernard92: what if it's a file but not a directory?

EverlastingBugstopper: oh good point

ashleymichal: bucket must be a directory.

src/commands/publish/mod.rs
```diff
@@ -89,7 +89,26 @@ fn publish_script(user: &GlobalUser, target: &Target) -> Result<(), failure::Err
...
+                )
+            }
             let path = Path::new(&bucket);
+            if !path.exists() {
+                failure::bail!(
+                    "{} bucket directory \"{}\" does not exist",
+                    emoji::WARN,
+                    path.to_string_lossy()
+                )
+            } else if !path.is_dir() {
```
ashleymichal: some places we use metadata, some places we use path. curious if metadata would be better since i suspect it comes from the OS rather than being inferred by the string contents.

ashleygwilliams: we should do a review and audit of this- filed an issue: https://github.com/cloudflare/wrangler/issues/696

ashleymichal: made a todo for later.

signalnerve:
Added a changelog tag to this but I think we can actually skip - it's an improvement to a feature that isn't totally out yet. OK with you @EverlastingBugstopper?

Issue #653, opened 2019-09-20
preview --watch does not work for site


ashleymichal:
We need to watch the specified `bucket` directory for changes and sync with the preview namespace on changes

edit from @EverlastingBugstopper

EverlastingBugstopper:
watch `bucket` separately from worker, upload kv assets on bucket file change, no changes to worker watch logic

EverlastingBugstopper:
also - probably blocked by https://github.com/cloudflare/wrangler/issues/351

PR #693, opened 2019-09-24
Add preview functionality for static assets

### LINKED ISSUE #599: make and destroy "preview" namespaces
on preview, wrangler should push bucket assets to the appropriate namespace. if the user has only one top level environment specified, wrangler should provision a separate kv-namespace (append `--preview` to the standard title name) to allow a user to preview their site without overwriting their uploaded assets.

ALTERNATIVELY

We could assume preview is _only_ for previewing the worker (after all, most static site generators have their own local dev server cli for developing the content itself), and use the same binding, but _not_ upload bucket on preview. This is likely the easiest implementation for MVP.

preview should also support `no-worker` and `no-bucket` flags like publish (see #598 )

gabbifish:
This PR closes #599. 

Its main contribution is that, when a preview is authenticated, it uses a special `*_preview` Workers KV namespace to allow someone to upload their static assets for use by the preview service (this avoids clobbering production and preview workers instances).

As part of this PR, this code does a small refactor of the preview and publication code. Namely, for preview and publish, it bundles the step of building the script's (`commands::build`) next to the step of generating the script's upload form. This function has been renamed `build_script_and_upload_form` to reflect this change accurately. This change ensures that script actions are bundled together. (When asset upload for preview was first implemented, it awkwardly had to happen between code building and code publication. This refactor should better delineate steps.)

This PR also ensures that the response for a static assets worker is not returned to the commandline, because it will be large and unwieldy. Instead, you get 

```console
$ wrangler preview
 Creating namespace for Workers Site "__blog-butt-workers_sites_assets_preview"
 Success
⬇️ Installing wasm-pack...
 Built successfully, built project size is 4 KiB.
 GET https://00000000000000000000000000000000.cloudflareworkers.com
 Your Worker is a Workers Sites preview, please preview on browser window.

```

I have tested that:
1. This static assets preview works; a static site's links can be followed on the preview page.
1. Preview and live reload functionality has remained unchanged for all workers types (JS, webpack, rust)
1. Publish still works as expected for all workers types.

I have noticed the following:
1. The preview service is substantially slower than production. Perhaps we should provide a disclaimer.
1. Live reload for the preview service doesn't work yet. This can be implemented by putting the assets upload logic in `watch_for_changes()`. Tracked in #653.

### DIFF
--- a/src/commands/publish/preview/mod.rs
+++ b/src/commands/publish/preview/mod.rs
@@ -7,7 +7,7 @@ mod http_method;
 pub use http_method::HTTPMethod;
 
 mod upload;
-use upload::upload_and_get_id;
+use upload::build_and_upload;
 
 use crate::commands;
 
@@ -34,8 +34,9 @@ pub fn preview(
     body: Option<String>,
     livereload: bool,
 ) -> Result<(), failure::Error> {
-    commands::build(&target)?;
-    let script_id = upload_and_get_id(&mut target, user.as_ref())?;
+    let sites_preview: bool = target.site.is_some();
+
+    let script_id = build_and_upload(&mut target, user.as_ref(), sites_preview)?;
 
     let session = Uuid::new_v4().to_simple();
     let preview_host = "example.com";
@@ -76,7 +77,12 @@ pub fn preview(
             HTTPMethod::Get => get(cookie, &client)?,
             HTTPMethod::Post => post(cookie, &client, body)?,
         };
-        let msg = format!("Your worker responded with: {}", worker_res);
+        let msg = match sites_preview {
+            false => format!("Your Worker responded with: {}", worker_res),
+            true => {
+                "Your Worker is a Workers Site, please preview it in browser window.".to_string()
+            }
+        };
         message::preview(&msg);
     }
 
@@ -130,11 +136,13 @@ fn watch_for_changes(
     session_id: String,
     broadcaster: Sender,
 ) -> Result<(), failure::Error> {
+    let sites_preview: bool = target.site.is_some();
+
     let (tx, rx) = channel();
     commands::watch_and_build(&target, Some(tx))?;
 
     while let Ok(_e) = rx.recv() {
-        if let Ok(new_id) = upload_and_get_id(&mut target, user) {
+        if let Ok(new_id) = build_and_upload(&mut target, user, sites_preview) {
             let msg = FiddleMessage {
                 session_id: session_id.clone(),
                 data: FiddleMessageData::LiveReload { new_id },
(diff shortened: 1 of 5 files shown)


src/commands/publish/preview/mod.rs
```diff
@@ -76,7 +77,12 @@ pub fn preview(
             HTTPMethod::Get => get(cookie, &client)?,
             HTTPMethod::Post => post(cookie, &client, body)?,
         };
-        let msg = format!("Your worker responded with: {}", worker_res);
+        let msg = match sites_preview {
+            false => format!("Your Worker responded with: {}", worker_res),
+            true => {
+                "Your Worker is a Workers Site, please preview it in browser window.".to_string()
```
EverlastingBugstopper: nit: I feel like this messaging is a bit awkward but I'm not sure what I'd replace it with. Maybe "This worker can only be previewed in the browser window"

src/commands/publish/preview/upload.rs
```diff
@@ -45,13 +47,24 @@ pub fn upload_and_get_id(
             if missing_fields.is_empty() {
                 let client = http::auth_client(&user);
 
+                // todo add sites to bucket here
```
EverlastingBugstopper: what exactly do you mean here? is there an issue filed for this?

gabbifish: This is for live preview for sites :) #653

src/commands/publish/preview/upload.rs
```diff
@@ -45,13 +47,24 @@ pub fn upload_and_get_id(
...
                 authenticated_upload(&client, &target)?
             } else {
                 message::warn(&format!(
                     "Your wrangler.toml is missing the following fields: {:?}",
                     missing_fields
                 ));
                 message::warn("Falling back to unauthenticated preview.");
+                if sites_preview {
+                    message::warn(
+                        "Note that unauthenticated preview will not preview Workers Sites.",
```
EverlastingBugstopper: nit: can we dry this message up? `message::warn(site_preview_warning)`

src/commands/publish/preview/upload.rs
```diff
@@ -65,6 +78,10 @@ pub fn upload_and_get_id(
                 "Run `wrangler config` or set $CF_API_KEY and $CF_EMAIL to configure your user.",
             );
 
+            if sites_preview {
+                message::warn("Note that unauthenticated preview will not preview Workers Sites.");
```
EverlastingBugstopper: dry

src/commands/publish/preview/mod.rs
gabbifish: I still think it's worth explicitly stating you're using a Workers Site :)

PR #699, opened 2019-09-24
Fix Rust Live Preview

### LINKED ISSUE #618: (rustwasm template) wrangler preview --watch enters "detected changes" loop
### 🐛 Bug Report

Wrangler 1.3.1 with rust template has major malfunction in the `wrangler preview --watch`. After command is launched all seems as expected. Though after first edit on a watched file the refresh cycle keeps spinning non-stop. Thanks a lot for looking into this and advising how to solve without waiting next release of wrangler.

#### Environment

* operating system: OSX 10.13.16
* output of `rustc -V`: rustc 1.37.0 (eae3437df 2019-08-13)
* output of `node -v`: v10.16.2
* output of `wrangler -V`: 1.3.1

Tried on both node and cargo installations with wrangler.

#### Steps to reproduce
1. `wrangler preview --watch`
2. Make an edit in, i.e. src/lib.rs file.

Single auto-realtime-update cycle.

7NM5TfhgYO
I.e. the refresh cycle keeps cycling nonstop without me touching the code at all .

gabbifish:
Fixes #618. The problem with Rust live preview was that it would keep on detecting changes as the rust-wasm toolchain built the wasm worker, meaning that changes to source code would trigger a cascade of file changes and thus, a cascade of file detection events and updates sent to the preview service. (So the problem wasn't dotfiles, afaik)...

This PR filters target/build paths for rust-wasm so they are not watched for file changes; namely, this includes the `pkg`, `target`, and  `worker-generated` directories. 

This has been tested on a Linux computer with Javascript, Webpack, and Rust workers and works as expected. Additional manual/real-world testing of this would be appreciated, especially on other operating systems (Mac, Windows)!

This PR also fixes watching of dotfiles for rust live preview.


src/commands/build/watch/watcher.rs
```diff
@@ -33,15 +35,67 @@ pub fn wait_for_changes(
...
+        DebouncedEvent::NoticeWrite(path) => Ok(filter_ignored(path, ignore_dirs)),
+        DebouncedEvent::Write(path) => Ok(filter_ignored(path, ignore_dirs)),
+        DebouncedEvent::NoticeRemove(path) => Ok(filter_ignored(path, ignore_dirs)),
+        DebouncedEvent::Remove(path) => Ok(filter_ignored(path, ignore_dirs)),
+        DebouncedEvent::Create(path) => Ok(filter_ignored(path, ignore_dirs)),
         _ => Ok(None),
     }
 }
+
+// Do not watch dotfiles. Also exclude files within ignored directories.
```
gabbifish:
```suggestion
// Exclude files within ignored directories from live reload.
```

src/commands/build/watch/watcher.rs
```diff
@@ -33,15 +35,67 @@ pub fn wait_for_changes(
...
+        return None;
+    }
+
+    let mut absolute_path: Vec<&str> = path_str.unwrap().split("/./").collect();
+    let relative_path = absolute_path.pop();
+    if relative_path.is_none() {
+        return None;
+    }
+
+    // First check if file in ignored directory.
```
gabbifish:
```suggestion
    // Check if file in ignored directory.
```

src/commands/build/watch/mod.rs
```diff
@@ -34,7 +38,7 @@ pub fn watch_and_build(
                 message::info(&format!("watching {:?}", &JAVASCRIPT_PATH));
 
                 loop {
-                    match wait_for_changes(&watcher_rx, COOLDOWN_PERIOD) {
+                    match wait_for_changes(&watcher_rx, COOLDOWN_PERIOD, None) {
```
ashleygwilliams: what does the third argument in this wait_for_changes function do? looks new?

gabbifish: This argument provides a set of directories whose contents live reload should ignore. So far only Rust Workers have explicit ignored directories.

gabbifish: removed third argument with latest version :)

src/commands/build/watch/mod.rs
```diff
@@ -54,7 +59,28 @@ pub fn watch_and_build(
...
+                let walker = WalkBuilder::new("./")
+                    .overrides(ignored_file_override)
+                    .build();
+
+                for entry in walker {
+                    let entry = entry.unwrap();
+                    if entry.path().is_dir() {
+                        continue;
+                    }
+                    println!("{:?}", entry.path());
```
EverlastingBugstopper: should remove this i think

gabbifish: oooh yes!

gabbifish:
```suggestion
```

src/commands/build/watch/mod.rs
```diff
@@ -15,6 +17,9 @@ pub const COOLDOWN_PERIOD: Duration = Duration::from_millis(2000);
 const JAVASCRIPT_PATH: &str = "./";
 const RUST_PATH: &str = "./";
 
+// Paths to ignore live watching in Rust Workers
+const RUST_IGNORE: &'static [&str] = &["pkg", "target", "worker/generated"];
```
EverlastingBugstopper: 👀 this is quite the syntax!

src/commands/build/watch/mod.rs
```diff
@@ -54,7 +59,27 @@ pub fn watch_and_build(
...
+                let ignored_file_override = ignored_files.build().unwrap();
+
+                let walker = WalkBuilder::new("./")
+                    .overrides(ignored_file_override)
+                    .build();
+
+                for entry in walker {
+                    let entry = entry.unwrap();
+                    if entry.path().is_dir() {
+                        continue;
```
EverlastingBugstopper: what is the purpose of this? does `continue` jump it back up to the start of the loop?

src/commands/build/watch/mod.rs
gabbifish: that's life with const string arrays!

src/commands/build/watch/mod.rs
gabbifish: we don't want to watch over entire directories; this will allow for unwanted files to be watched (and trigger continuous loading). we skip and continue looping.

PR #708, opened 2019-09-24
Insert hashes before extensions in asset manifest to ease content-type detection

### LINKED ISSUE #697: insert hash before extension instead of appending to key
we use mime to find out what content type to respond with in the worker, and by appending a hash to the end of the key, we confuse it and come up with null content type. there are two options here:

1. we can snip the hash off the end of the key in the worker when we got to check the content-type. this is possibly the easiest approach, and would be an issue in the package.
2. we can insert the hash between the file name and its extension the way that other static site generators do (format `path/to/filename.<hash>.ext`).

ashleymichal:
Closes #697 

Appending the content hash to the end of all of the asset paths we are sending up has confused the Worker and caused it to start returning `content-type: null`. This PR changes the fingerprinting logic to insert our hash between the filename and the extension, similar to the way other generators fingerprint.

Tested with the branch `victoria/pathname` of `kv-asset-handler`

### DIFF
--- a/src/commands/kv/bucket/mod.rs
+++ b/src/commands/kv/bucket/mod.rs
@@ -36,7 +36,7 @@ pub fn directory_keys_values(
             let b64_value = base64::encode(&value);
 
             let (url_safe_path, key) =
-                generate_url_safe_key_and_hash(path, directory, Some(b64_value.clone()))?;
+                generate_path_and_key(path, directory, Some(b64_value.clone()))?;
 
             if verbose {
                 message::working(&format!("Parsing {}...", key.clone()));
@@ -66,7 +66,7 @@ fn directory_keys_only(directory: &Path) -> Result<Vec<String>, failure::Error> 
             // Need to base64 encode value
             let b64_value = base64::encode(&value);
 
-            let (_, key) = generate_url_safe_key_and_hash(path, directory, Some(b64_value))?;
+            let (_, key) = generate_path_and_key(path, directory, Some(b64_value))?;
 
             upload_vec.push(key);
         }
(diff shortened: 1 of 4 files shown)


src/commands/kv/bucket/mod.rs
```diff
@@ -113,7 +113,11 @@ pub fn generate_url_safe_key_and_hash(
 
     let path_with_hash = if let Some(value) = value {
         let digest = get_digest(value)?;
-        format!("{}-{}", url_safe_path, digest)
+
+        generate_path_with_hash(path, digest)?
```
EverlastingBugstopper: this `unwrap` makes me uneasy

gabbifish: We should just be able to use `?` instead, correct?

ashleymichal: pushed an update based on in air convo w/ @EverlastingBugstopper

ashleymichal:
failing on windows. will update tomorrow.

src/commands/kv/bucket/mod.rs
```diff
@@ -100,22 +98,26 @@ fn generate_url_safe_path(path: &Path, directory: &Path) -> Result<String, failu
...
-pub fn generate_url_safe_key_and_hash(
+// e.g (sitemap.xml, sitemap.ec717eb2131fdd4fff803b851d2aa5b1dc3e0af36bc3c8c40f2095c747e80d1e.xml)
+pub fn generate_path_and_key(
     path: &Path,
     directory: &Path,
     value: Option<String>,
 ) -> Result<(String, String), failure::Error> {
-    let url_safe_path = generate_url_safe_path(path, directory)?;
+    // strip the bucket directory from both paths for ease of reference.
+    let relative_path = path.strip_prefix(directory).unwrap();
```
EverlastingBugstopper: can we do an if let with a failure in the else

ashleymichal: we can; fwiw this is code moved from a lower function up into this one verbatim, so it's sort of status quo

EverlastingBugstopper: ok :)

src/commands/kv/bucket/mod.rs
```diff
@@ -75,10 +75,8 @@ fn directory_keys_only(directory: &Path) -> Result<Vec<String>, failure::Error>
 }
 
 // Courtesy of Steve Klabnik's PoC :) Used for bulk operations (write, delete)
-fn generate_url_safe_path(path: &Path, directory: &Path) -> Result<String, failure::Error> {
-    let path = path.strip_prefix(directory).unwrap();
-
-    // next, we have to re-build the paths: if we're on Windows, we have paths with
+fn generate_url_safe_path(path: &Path) -> Result<String, failure::Error> {
+    // first, we have to re-build the paths: if we're on Windows, we have paths with
```
gabbifish: Don't we want to continue stripping the prefix? This logic will include the absolute directory and not the relative one. This could be bad if multiple people work on a workers site and have different absolute paths to their static site code.

gabbifish: Oh i see, we moved this!

PR #709, opened 2019-09-24
Custom webpack configuration behavior

### LINKED ISSUE #296: Require webpack_config field in wrangler.toml to use custom webpack configuration
It is not abundantly clear that wrangler will use your custom `webpack.config.js` to package everything. Some users may want to start using wrangler with a project that already has a `webpack.config.js`, but they may not understand that wrangler will use that configuration over its own default. We should require users who want to use their own configuration to specify it in their `wrangler.toml` `webpack_config` field so that there is no confusion that their webpack configuration is impacting their build output.

We need to decide if this is something we want to do, or if we want to leave things like they are and automatically override wrangler’s default webpack behavior when a user has their own `webpack.config.js`

**Edit**

Implementing this would be a breaking change for users that currently use `webpack.config.js` to build their workers, they would need to add a `webpack_config` field to their `wrangler.toml` in order for it to continue working. We would want to add an error message if they have a file `webpack.config.js` to say that if they want it to be used, they should add it to their configuration to override the default.

We're looking at implementing this for the upcoming release because some users will use `webpack.config.js` to build other types of apps alongside their worker. In these cases, the user will want to keep `webpack.config.js` around, but not to build their worker

EverlastingBugstopper:
This PR does two things:

1) If users are deploying a non-site project and have a `webpack.config.js` file, we use it to build their worker. This will be deprecated in 1.6.0 in favor of specifying a `webpack_config` field in `wrangler.toml`.

2) If users are deploying a site project and have a `webpack.config.js` file, we will not use it to build their worker. This is because it's probably for their react app and not the worker.

Fixes #296

### DIFF
--- a/src/commands/build/wranglerjs/mod.rs
+++ b/src/commands/build/wranglerjs/mod.rs
@@ -160,11 +160,12 @@ fn setup_build(target: &Target) -> Result<(Command, PathBuf, Bundle), failure::E
     command.arg(format!("--wasm-binding={}", bundle.get_wasm_binding()));
 
     let webpack_config_path = if let Some(webpack_config) = &target.webpack_config {
-        Some(PathBuf::from(webpack_config.clone()))
+        Some(PathBuf::from(&webpack_config))
     } else if target.site.is_none() {
         message::warn("In Wrangler v1.6.0, you will need to include a webpack_config field in your wrangler.toml to build with a custom webpack configuration.");
         Some(PathBuf::from("webpack.config.js".to_string()))
     } else {
+        // don't use `webpack.config.js` if this project is a site
         None
     };
 
--- a/src/commands/build/wranglerjs/mod.rs
+++ b/src/commands/build/wranglerjs/mod.rs
@@ -159,7 +159,7 @@ fn setup_build(target: &Target) -> Result<(Command, PathBuf, Bundle), failure::E
 
     command.arg(format!("--wasm-binding={}", bundle.get_wasm_binding()));
 
-    let webpack_config_path: Option<PathBuf> = if let Some(webpack_config) = &target.webpack_config {
+    let webpack_config_path = if let Some(webpack_config) = &target.webpack_config {
         // require webpack_config in wrangler.toml to use it in sites
         Some(PathBuf::from(&webpack_config))
     } else if target.site.is_none() {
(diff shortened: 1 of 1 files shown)


ashleymichal:
in case 2: if they specify webpack_config in their toml, do we use it?

EverlastingBugstopper:
yes - if they specify `webpack_config` in their `wrangler.toml` it will be used. the path is relative to `wrangler.toml`

src/commands/build/wranglerjs/mod.rs
```diff
@@ -159,35 +159,53 @@ fn setup_build(target: &Target) -> Result<(Command, PathBuf, Bundle), failure::E
...
-            .webpack_config
-            .clone()
-            .unwrap_or_else(|| "webpack.config.js".to_string()),
-    );
+    let webpack_config_path = if let Some(webpack_config) = &target.webpack_config {
+        Some(PathBuf::from(webpack_config.clone()))
+    } else if target.site.is_none() {
+        message::warn("In Wrangler v1.6.0, you will need to include a webpack_config field in your wrangler.toml to build with a custom webpack configuration.");
+        Some(PathBuf::from("webpack.config.js".to_string()))
+    } else {
```
gabbifish: this conditional sequence took me a while to grok; from what i understand, the final `else` option is for the case when a site is specified and there is no webpack_config, so no webpack config is used? I'd love a comment to clarify that :D

EverlastingBugstopper: cool

src/commands/build/wranglerjs/mod.rs
```diff
@@ -159,35 +159,54 @@ fn setup_build(target: &Target) -> Result<(Command, PathBuf, Bundle), failure::E
...
-    let webpack_config_path = PathBuf::from(
-        &target
-            .webpack_config
-            .clone()
-            .unwrap_or_else(|| "webpack.config.js".to_string()),
-    );
+    let webpack_config_path = if let Some(webpack_config) = &target.webpack_config {
+        Some(PathBuf::from(&webpack_config))
+    } else if target.site.is_none() {
+        message::warn("In Wrangler v1.6.0, you will need to include a webpack_config field in your wrangler.toml to build with a custom webpack configuration.");
```
ashleymichal: i think we also need to switch this warning based on whether there is a wrangler.config.js at the root (what we would use as the default).

EverlastingBugstopper: thanks :) + fixed

Issue #464, opened 2019-08-22
Add way to ignore certain assets


EverlastingBugstopper:
#### Overview and problem statement

Ignore `.DS_Store` and maybe everything in `.gitignore`. Perhaps we need a `.wignore` or something for assets we don't want to upload

ashleymichal:
hmm. piggybacking on gitignore would be great, however i think we would need to hook into git's logic since many users put `.DS_Store` in their root gitignore.

ashleymichal:
on second thought, it is common to gitignore static site public folders since they are repeatedly regenerated from source; so gitignore is not a thing we should do. Perhaps we should just ignore "hidden" files? Should look and see what other solutions do.

ashleymichal:
we should for sure ignore things like `node_modules`

gabbifish:
This will be super useful for .wignore: https://git-scm.com/docs/gitignore

We should have feature parity with gitignore. :)

EverlastingBugstopper:
https://crates.io/crates/ignore

gabbifish:
Closing because this is now tracked in #716

PR #710, opened 2019-09-24
Ignore uploading node_modules

### LINKED ISSUE #667: Sourcemaps are unable to be uploaded when publishing a site with Wrangler
### 🐛 Bug Report

When trying to publish a Gatsby site, the sourcemaps generated for each page are large enough that they make the `publish` step fail. IMO we should ignore the sourcemap files (maybe via a config) or warn on this instead of throwing an error.

```
❯ ~/.cargo/bin/wrangler publish
🌀  Creating namespace for Workers Site "__frontendjobs-closedsource-workers_sites_assets"
Error: File `component---src-pages-post-js-c6d1c3aab5c008b72fa8.js.map-b3fd6703031f027b11dd2dc7e3448fe3838efa53e5c6436c4aa3dd9c721cc7e4` of 3252144 bytes exceeds the maximum value size limit of 2097152 bytes
```

#### Environment

`wrangler` built from master this AM 

#### Steps to reproduce

Run `wrangler publish` on a Workers site with large sourcemaps. Sorry, my project is closed-source, but happy to give access to Wrangler team if needed!

gabbifish:
This PR closes #667. It is a first overture towards one day handling #464, aka adding `.wignore` functionality.

This PR introduces logic for filtering files the bucket `upload` function traverses. Namely, it creates a base for filtering two types of items:
1. directories
1. files with certain prefixes (like dotfiles).

Directories are entirely filtered if they match a certain directory name, in this case `node_modules`. Files are filtered if their prefix matches one that is intended to be filtered.

This reduces the number of times someone is likely to hit the KV maximum size API error via node_module or Gatsby site sourcemap uploads.

### DIFF
--- a/src/commands/kv/bucket/mod.rs
+++ b/src/commands/kv/bucket/mod.rs
@@ -80,10 +80,10 @@ fn directory_keys_only(directory: &Path) -> Result<Vec<String>, failure::Error> 
 
 // todo(gabbi): Replace all the logic below with a proper .wignore implementation
 // when possible.
-const KNOWN_UNNECESSARY_DIRS: &'static [&str] = &[
+const KNOWN_UNNECESSARY_DIRS: &[&str] = &[
     "node_modules", // npm vendoring
 ];
-const KNOWN_UNNECESSARY_FILE_PREFIXES: &'static [&str] = &[
+const KNOWN_UNNECESSARY_FILE_PREFIXES: &[&str] = &[
     ".", // hidden files
 ];
 fn is_ignored(entry: &DirEntry) -> bool {
--- a/src/commands/kv/bucket/mod.rs
+++ b/src/commands/kv/bucket/mod.rs
@@ -84,8 +84,7 @@ const KNOWN_UNNECESSARY_DIRS: &'static [&str] = &[
     "node_modules", // npm vendoring
 ];
 const KNOWN_UNNECESSARY_FILE_PREFIXES: &'static [&str] = &[
-    "component---", // Gatsby sourcemaps
-    ".",            // hidden files
+    ".", // hidden files
 ];
 fn is_ignored(entry: &DirEntry) -> bool {
     let stem = entry.file_name().to_str().unwrap();
--- a/src/commands/kv/bucket/mod.rs
+++ b/src/commands/kv/bucket/mod.rs
@@ -14,7 +14,7 @@ use std::path::Path;
 
 use cloudflare::endpoints::workerskv::write_bulk::KeyValuePair;
 
-use walkdir::WalkDir;
+use walkdir::{DirEntry, WalkDir};
 
 use crate::terminal::message;
 
@@ -26,7 +26,10 @@ pub fn directory_keys_values(
     let mut upload_vec: Vec<KeyValuePair> = Vec::new();
     let mut key_manifest: HashMap<String, String> = HashMap::new();
 
-    for entry in WalkDir::new(directory).into_iter().filter_entry(|e| !is_ignored(e)) {
+    for entry in WalkDir::new(directory)
+        .into_iter()
+        .filter_entry(|e| !is_ignored(e))
+    {
         let entry = entry.unwrap();
         let path = entry.path();
         if path.is_file() {
@@ -48,6 +51,7 @@ pub fn directory_keys_values(
                 expiration_ttl: None,
                 base64: Some(true),
             });
+
             key_manifest.insert(url_safe_path, key);
         }
     }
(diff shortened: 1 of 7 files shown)


src/commands/kv/bucket/upload.rs
```diff
@@ -94,19 +96,31 @@ fn call_put_bulk_api(
...
         }
     }
     filtered_pairs
 }
 
+// todo(gabbi): Replace all the logic below with a proper .wignore implementation
+// when possible.
+const KNOWN_UNNECESSARY_PREFIXES: &'static [&str] = &[
+    "node_modules/", // npm vendoring
+    "component---",  // Gatsby sourcemaps
```
ashleymichal: are these _always_ too big? curious if we should include for a first pass.

src/commands/kv/bucket/upload.rs
```diff
@@ -94,19 +96,31 @@ fn call_put_bulk_api(
...
+];
+fn if_ignored_prefix(key: &str) -> bool {
```
ashleymichal: nit:
```suggestion
fn contains_ignored_prefix(key: &str) -> bool {
```

src/commands/kv/bucket/upload.rs
```diff
@@ -94,19 +96,31 @@ fn call_put_bulk_api(
...
+    for prefix in KNOWN_UNNECESSARY_PREFIXES {
+        if key.starts_with(prefix) {
```
ashleymichal: for this first pass w/ just node_modules, should we ignore any path that includes node_modules?

src/commands/kv/bucket/upload.rs
gabbifish: they're apparently quite the menace (and unnecessary) for gatsby users--there are plenty of js libs that allow folks to skip generating them at all. I think we can get away with ignoring them completely, but I'd love @signalnerve's input before I do anything too rash :)

ashleymichal: yah word; if nothing else we can always do a one-liner with just that.

src/commands/kv/bucket/upload.rs
gabbifish: Hmm, where else would node modules live? I'm not opposed to it, it just seems like unorthodox ignoring behavior (it usually follows prefixes, not the presence of a string)

ashleymichal: i noticed as i was developing https://github.com/cloudflare/wrangler/pull/708 that the key itself still includes the directory name, so this... might not work? or if it does i am confused as to why?

ashleymichal: turns out i broke this.

gabbifish: I added more finessed logic for looking at whether a `node_module` subdirectory existed anywhere within a bucket :) So @ashleymichal we should have your suggested functionality ready to go!

src/commands/kv/bucket/mod.rs
```diff
@@ -26,7 +26,10 @@ pub fn directory_keys_values(
     let mut upload_vec: Vec<KeyValuePair> = Vec::new();
     let mut key_manifest: HashMap<String, String> = HashMap::new();
 
-    for entry in WalkDir::new(directory) {
+    for entry in WalkDir::new(directory)
+        .into_iter()
+        .filter_entry(|e| !is_ignored(e))
```
ashleymichal: ✨

ashleygwilliams:
so i have a approach question - gatsby source maps are not always too big- i think we may just want this to block node_modules for now, and then we can explore the .wranglerignore to ignore too big files, so folks opt in

src/commands/kv/bucket/upload.rs
ashleygwilliams: heh just realized my comment below is on this. i would say we should opt to not do it in this one and then do the .wranglerignore to address these... thoughts?

gabbifish: I'm happy to remove it for now!

gabbifish:
@ashleygwilliams updated this PR to remove the gatsby sourcemap logic. Let me know if this is good to go! :D

ashleymichal:
LGTM! ✅

src/commands/kv/bucket/mod.rs
```diff
@@ -128,3 +160,87 @@ fn get_digest(value: String) -> Result<String, failure::Error> {
...
+
+    #[test]
+    fn it_can_ignore_prefix() {
+        let file_name = ".dotfile";
+        // If test file already exists, delete it.
+        if fs::metadata(file_name).is_ok() {
+            fs::remove_file(file_name).unwrap();
+        }
+
+        fs::File::create(file_name).unwrap();
```
EverlastingBugstopper: do these tests create a real file in the file system or just something in memory? would be great if we didn't add more tests relying on file system but i'm not going to block on this

gabbifish: I tried using tempfiles but had absolutely no success :( It seems that the tempfiles crate is NOT for testing purposes like mine (it produces temp gobbledygook directory names instead of allowing you to create a directory with the name you want, e.g. `node_modules`). So, I had to revert to using a real file in the filesystem.

EverlastingBugstopper: ok :)

ashleygwilliams: agree, but also can say this is a major RM planned for Q4

ashleygwilliams:
WIP- @EverlastingBugstopper to refactor code to remove double asset manifest generation to remove the double error message printing.

PR #714, opened 2019-09-25
Add live reload for static sites

### LINKED ISSUE #653: preview --watch does not work for site

gabbifish:
Fixes #653. This PR adds support for `wrangler preview --watch` to work with static assets.

There is still a quirks I want to work out in the future: is the 2000 ms cooldown period enough to account for large static site generator builds that exceed that time? (I think it should be, we can wait to hear otherwise...)

### DIFF
--- a/src/commands/build/wranglerjs/mod.rs
+++ b/src/commands/build/wranglerjs/mod.rs
@@ -77,6 +77,8 @@ pub fn run_build_and_watch(target: &Target, tx: Option<Sender<()>>) -> Result<()
             if Path::new(&bucket).exists() {
                 watcher.watch(&bucket, RecursiveMode::Recursive)?;
                 info!("watching static sites asset file {:?}", &bucket);
+            } else {
+                failure::bail!("Attempting to watch static assets bucket \"{}\" which doesn't exist", bucket);
             }
         }
 
--- a/src/commands/build/wranglerjs/mod.rs
+++ b/src/commands/build/wranglerjs/mod.rs
@@ -58,6 +58,8 @@ pub fn run_build_and_watch(target: &Target, tx: Option<Sender<()>>) -> Result<()
     let (mut command, temp_file, bundle) = setup_build(target)?;
     command.arg("--watch=1");
 
+    let is_site = target.site.clone();
+
     info!("Running {:?} in watch mode", command);
 
     //Turbofish the result of the closure so we can use ?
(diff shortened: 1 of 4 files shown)


EverlastingBugstopper:
Fixes #653

ashleymichal:
QA-ing this; the output confuses me, esp the duplicate "ignoring" line, and the "Success" message, i'm not sure what has succeeded in each case.


``` sh
$ wrangler-dev preview --watch
🌀  Creating namespace for Workers Site "__hello-react-workers_sites_assets_preview"
💁  ignoring directory node_modules
✨  Success
💁  ignoring directory node_modules
✨  Success
✨  Built successfully, built project size is 4 KiB.
```

src/commands/build/wranglerjs/mod.rs
```diff
@@ -58,6 +58,8 @@ pub fn run_build_and_watch(target: &Target, tx: Option<Sender<()>>) -> Result<()
     let (mut command, temp_file, bundle) = setup_build(target)?;
     command.arg("--watch=1");
 
+    let is_site = target.site.clone();
```
xtuc: This should be maybe_site I think

src/commands/build/wranglerjs/mod.rs
```diff
@@ -68,9 +70,21 @@ pub fn run_build_and_watch(target: &Target, tx: Option<Sender<()>>) -> Result<()
         let mut watcher = notify::watcher(watcher_tx, Duration::from_secs(1))?;
 
         watcher.watch(&temp_file, RecursiveMode::Recursive)?;
-
         info!("watching temp file {:?}", &temp_file);
 
+        if let Some(site) = is_site {
```
xtuc: Similar to above, this could be:
```rs

```

gabbifish:
@ashleymichal ugh something in master is responsible for these double printouts, they did not happen until I merged master in. I'm investigating.

src/commands/build/wranglerjs/mod.rs
ashleygwilliams: why maybe site? i think is_site says if it is a site or not, the maybe will be very tough to understand later. if target.site exists, it's a site. am i missing something perhaps?

src/commands/build/wranglerjs/mod.rs
ashleymichal: agreed but not super critical.

PR #621, opened 2019-09-18
Add readme for sites

victoriabernard92:
Reference documentation for sites using wrangler:

- [ ] double check the output of wrangler init and publish

PR #720, opened 2019-09-26
[Wrangler Ignore option 1] .wranglerignore logic for static asset uploads

### LINKED ISSUE #716: RFC: .wranglerignore
### The Problem
There are static assets directories that may contain lots of content that doesn't need to be uploaded to KV (and likely will trigger KV value size limits). This includes, but is not limited to, things like `node_modules/` and hidden files/directories. Users should be able to specify paths (in the vein of `.gitignore`) that do not get uploaded to KV.

### Option 1: .wranglerignore
We will introduce support for `.wranglerignore` files. These files will be automatically generated with `wrangler init` and `wrangler generate` and placed in the same directory (pwd) as `wrangler.toml`. By default Wrangler will ignore `node_modules` and hidden files (they don't need to be specified in `.wranglerignore`).

These files will be parsed and used exactly like `.gitignore` whenever bucket `upload_files()` is called. The https://docs.rs/ignore/0.4.10/ignore crate will help us a lot here.

### Option 2: include/exclude in wrangler.toml
A more rust-idomatic include/exclude approach is the one used by cargo: https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields-optional

This approach would become part of the `[site]` section of `wrangler.toml`, which would have two new fields, exclude and include. If both are present, wrangler will use include only (same as cargo). The exclude and include fields are used to filter file uploads from whatever directory is specified in the bucket field in wrangler.toml.

The benefits of this approach are that
- Wrangler users will have one fewer file to deal with (no .wranglerignore)
- We can constrain this ignore functionality to just Workers Sites, the only workers application that would use this ignore functionality
- The coupling between Workers sites and include/exclude would neatly be represented in wrangler.toml.

This implementation should be readily doable with https://docs.rs/ignore/0.4.10/ignore/overrides/struct.OverrideBuilder.html.

This approach makes it easier, code-wise, to always ignore node_modules (no messiness about expecting it in the .wranglerignore, we can just always put this logic in the OverrideBuilder).

I advocate for approach number 2. It's fairly easy for me to make implementations of both approaches above, so I can go ahead and the team can play around with the options and tell me which one they like best.

### Proofs of concept
Our two candidates:
1. `.wranglerignore` #720 
1. `include/exclude` in `wrangler.toml` #760

gabbifish:
This PR is the first implementation/option for #716, using `.wranglerignore`. The second option is #760.

This PR will also includes a new docs section for `.wranglerignore` in the sites docs. The docs addition for this is blocked until #621 is merged, but you can read the current draft in a github gist: https://gist.github.com/gabbifish/538edb075a9ffe9501e43ef4da410b22


ashleygwilliams:
fixes : https://github.com/cloudflare/wrangler/issues/464

gabbifish:
This is being closed in favor of #760

PR #726, opened 2019-09-26
Install latest version when binary installing

### LINKED ISSUE #666: may need to update our cargo generate version in generate

EverlastingBugstopper:
This PR fixes a bug where Wrangler would use system installs of `cargo-generate` (or other crates) even if they were out of date.

# Fixes #666 😈

### DIFF
--- a/src/install/mod.rs
+++ b/src/install/mod.rs
@@ -10,6 +10,7 @@ use which::which;
 
 use std::env;
 use std::path::Path;
+use std::process::Command;
 
 use lazy_static::lazy_static;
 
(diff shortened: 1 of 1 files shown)


src/install/mod.rs
```diff
@@ -51,14 +52,46 @@ pub fn install_artifact(
...
+            .output()
+            .expect(&no_version_msg);
+
+        if !tool_version_output.status.success() {
+            let error = String::from_utf8_lossy(&tool_version_output.stderr);
+            log::debug!("could not find version for {}\n{}", tool_name, error);
+            return Ok(None);
+        }
+
+        let installed_tool_version = String::from_utf8_lossy(&tool_version_output.stdout);
```
ashleygwilliams: let's make this logic a function

PR #727, opened 2019-09-26
Use emoji for failure

EverlastingBugstopper:
also don't print "falling back to unauthenticated" when it can't


gabbifish:
So the way this error message will render is `Error: <emoji> message`... I don't know if we want that?

EverlastingBugstopper:
declining

PR #737, opened 2019-09-27
Support pagination of KV namespace API to look at all namespaces

gabbifish:
This PR fixes an issue where Wrangler would not find an existing namespace, because the full list of namespaces is now allowed to be over 20 (and the API once only returned 20 at a time). This means that some of the namespaces Wrangler would want would look like they didn't exist at all!

This would produce a panic when Wrangler expected to find a preexisting namespace and could not.

TODO: We need to publish ashley's cloudflare-rs fork so we can point to that instead of her github repo. After that this PR is no longer WIP.


src/commands/kv/namespace/list.rs
```diff
@@ -1,25 +1,47 @@
+extern crate serde_json;
+
 use cloudflare::endpoints::workerskv::list_namespaces::ListNamespaces;
+use cloudflare::endpoints::workerskv::list_namespaces::ListNamespacesParams;
```
ashleymichal: these two lines, that import `ListNamespaces` and `ListNamespacesParams`, need to come from the dubs crate.

src/commands/kv/namespace/site.rs
```diff
@@ -1,6 +1,5 @@
-use cloudflare::endpoints::workerskv::create_namespace::CreateNamespace;
-use cloudflare::endpoints::workerskv::create_namespace::CreateNamespaceParams;
-use cloudflare::endpoints::workerskv::list_namespaces::ListNamespaces;
+use cloudflare::endpoints::workerskv::create_namespace::{CreateNamespace, CreateNamespaceParams};
+use cloudflare::endpoints::workerskv::list_namespaces::{ListNamespaces, ListNamespacesParams};
```
ashleymichal: same in here: this line with the `ListNamespaces, ListNamespacesParams` imports needs to come from the dubs crate.

Issue #701, opened 2019-09-24
`wrangler subdomain` without args should print your subdomain if you have one


EverlastingBugstopper:
and if you don't have one it should tell you to register one by running `wrangler subdomain <name>`

PR #747, opened 2019-09-29
Print subdomain if `name` arg not provided to `wrangler subdomain`

gusvargas:
Fixes https://github.com/cloudflare/wrangler/issues/701

I added a new method, `Subdomain::get_as_option`, since it seems like the common case is to assert that the subdomain is available.

### DIFF
--- a/src/commands/subdomain.rs
+++ b/src/commands/subdomain.rs
@@ -11,7 +11,7 @@ pub struct Subdomain {
 }
 
 impl Subdomain {
-    pub fn get(account_id: &str, user: &GlobalUser) -> Result<String, failure::Error> {
+    fn request(account_id: &str, user: &GlobalUser) -> Result<Response, failure::Error> {
         let addr = subdomain_addr(account_id);
 
         let client = http::auth_client(None, user);
@@ -27,11 +27,23 @@ impl Subdomain {
             )
         }
 
-        let res: Response = serde_json::from_str(&res.text()?)?;
+        Ok(serde_json::from_str(&res.text()?)?)
+    }
+
+    pub fn get(account_id: &str, user: &GlobalUser) -> Result<String, failure::Error> {
+        let res = Subdomain::request(account_id, user)?;
         Ok(res
             .result
             .expect("Oops! We expected a subdomain name, but found none.")
             .subdomain)
+    }
+
+    pub fn get_as_option(
+        account_id: &str,
+        user: &GlobalUser,
+    ) -> Result<Option<String>, failure::Error> {
+        let res = Subdomain::request(account_id, user)?;
+        Ok(res.result.map(|r| r.subdomain))
     }
 }
 
@@ -58,13 +70,7 @@ fn subdomain_addr(account_id: &str) -> String {
     )
 }
 
-pub fn subdomain(name: &str, user: &GlobalUser, target: &Target) -> Result<(), failure::Error> {
-    if target.account_id.is_empty() {
-        failure::bail!(format!(
-            "{} You must provide an account_id in your wrangler.toml before creating a subdomain!",
-            emoji::WARN
-        ))
-    }
+fn register_domain(name: &str, user: &GlobalUser, target: &Target) -> Result<(), failure::Error> {
     let msg = format!(
         "Registering your subdomain, {}.workers.dev, this could take up to a minute.",
         name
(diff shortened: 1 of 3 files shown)


src/commands/subdomain.rs
```diff
@@ -58,13 +70,34 @@ fn subdomain_addr(account_id: &str) -> String {
...
     if target.account_id.is_empty() {
         failure::bail!(format!(
             "{} You must provide an account_id in your wrangler.toml before creating a subdomain!",
             emoji::WARN
         ))
     }
+    if name.is_none() {
+        return match Subdomain::get_as_option(&target.account_id, user) {
+            Ok(Some(subdomain)) => {
+                println!("{}.workers.dev", subdomain);
```
gusvargas: Should this use `println`, `print`, or `message::info`? Unsure if it'd be useful for this to be piped to stdin – probably not?

src/commands/subdomain.rs
```diff
@@ -58,13 +70,34 @@ fn subdomain_addr(account_id: &str) -> String {
...
+    user: &GlobalUser,
+    target: &Target,
+) -> Result<(), failure::Error> {
     if target.account_id.is_empty() {
         failure::bail!(format!(
             "{} You must provide an account_id in your wrangler.toml before creating a subdomain!",
             emoji::WARN
         ))
     }
+    if name.is_none() {
```
EverlastingBugstopper: This is kind of a nit, but instead of `if name.is_none()` and then `unwrap` later, I think you can do something kinda like this

```rust
let name = if let Some(name) = name {
  name
} else {
  return match Subdomain .......
};
```

I _think_ I like that style better than `unwrap` but I'm still learning Rust :)

gusvargas:
Thanks for the :eyes:. I'll try to get those changes in within the next couple of days.

gusvargas:
@EverlastingBugstopper I've addressed your feedback. I also slightly refactored things so that the states -> behavior is more explicit.

src/commands/subdomain.rs
```diff
@@ -27,12 +27,24 @@ impl Subdomain {
...
             .expect("Oops! We expected a subdomain name, but found none.")
             .subdomain)
     }
+
+    pub fn get_as_option(
+        account_id: &str,
+        user: &GlobalUser,
+    ) -> Result<Option<String>, failure::Error> {
+        let res = Subdomain::request(account_id, user)?;
+        Ok(res.result.map(|r| r.subdomain))
```
EverlastingBugstopper: 🔥

gusvargas:
Thanks for the feedback, @ashleygwilliams – it definitely reads a lot clearer now.

@EverlastingBugstopper `get_as_option` -> `get` makes sense and I agree that it's probably best to do that refactor as part of its own patch.

src/commands/subdomain.rs
```diff
@@ -126,6 +136,44 @@ pub fn subdomain(name: &str, user: &GlobalUser, target: &Target) -> Result<(), f
...
+            let msg = format!("This account already has a registered subdomain. You can only register one subdomain per account. Your subdomain is {}.workers.dev", subdomain);
+            message::user_error(&msg);
+            Ok(())
+        }
+        Err(error) => Err(error),
+    }
+}
+
+pub fn get_subdomain(user: &GlobalUser, target: &Target) -> Result<(), failure::Error> {
+    let subdomain = Subdomain::get_as_option(&target.account_id, user);
```
EverlastingBugstopper: You could simplify this with a `?` at the end of the `let subdomain` line and then the `match` statement could just be `Some(subdomain) => ` and `None =>`

src/commands/subdomain.rs
```diff
@@ -126,6 +136,44 @@ pub fn subdomain(name: &str, user: &GlobalUser, target: &Target) -> Result<(), f
...
 }
 
+pub fn set_subdomain(name: &str, user: &GlobalUser, target: &Target) -> Result<(), failure::Error> {
+    if target.account_id.is_empty() {
+        failure::bail!(format!(
+            "{} You must provide an account_id in your wrangler.toml before creating a subdomain!",
+            emoji::WARN
+        ))
+    }
+    let subdomain = Subdomain::get_as_option(&target.account_id, user);
```
EverlastingBugstopper: Could use `?`

src/commands/subdomain.rs
gusvargas: Ah right – I still haven't internalized all of the useful syntax and idioms. Fixed in https://github.com/cloudflare/wrangler/pull/747/commits/1dd74e1274363d50167619f29efc1cf51938f185.

Issue #63, opened 2019-04-09
refactor all cloudflare API calls to use cloudflare-rs


ashleygwilliams:
as we implemented KV support in Wrangler, we adopted the cloudflare-rs crate (https://github.com/cloudflare/cloudflare-rs) as our official cloudflare api client for use with those endpoints. We should make an effort to move _all_ of our cloudflare api calls into that crate, which would give us uniformity in api request and response handling, as well as a more robust pattern for writing unit tests (the crate comes with an example mock api client).

should likely convert this into a project to track work both in wrangler and in cloudflare-rs.

as per https://github.com/cloudflare/wrangler/pull/57#discussion_r273559460

petebacondarwin:
Closing as the latest Wrangler version is not implemented in Rust and will not use `cloudflare-rs`.

PR #749, opened 2019-09-30
[WIP] Consistency in passing &GlobalUser and refactoring kv::namespace calls

### LINKED ISSUE #623: Refactor kv:namespace create and kv:namespace list
we should refactor some of kv commands to use private functions that return structs rather than always printing to stdout for use by other commands.

gabbifish:
Closes #623.

This PR does two things:
1. Refactors the passing of GlobalUser as a reference to be consistent with other wrangler command invocations
1. Reduces a duplicate implementation of calling ListNamespaces, instead using `kv::commands::namespace::list::call_api`.

I wanted to remove the duplicate implementation of CreateNamespace, but the problem is that `kv::commands::namespace::create::call_api`, if it existed, must return a `failure::Error` because it must creates an API client and THEN calls the API endpoint. But `site()` expects an `ApiError` in response, because it wants access to the API error code to understand when a namespace already exists or whether the user doesn't have Workers KV entitled to their account. I'm trying to think of workarounds here, but they would honestly convolute the `kv::commands::namespace::create` crate; calling `CreateNamespace` directly from `site()` is not the end of the world given how custom its handling logic is.

### DIFF
--- a/src/commands/kv/namespace/site.rs
+++ b/src/commands/kv/namespace/site.rs
@@ -18,7 +18,7 @@ pub fn site(
     preview: bool,
 ) -> Result<WorkersKvNamespace, failure::Error> {
     kv::validate_target(target)?;
-    let client = kv::api_client(user.to_owned())?;
+    let client = kv::api_client(user)?;
 
     let title = if preview {
         format!("__{}-{}", target.name, "workers_sites_assets_preview")
--- a/src/commands/kv/namespace/site.rs
+++ b/src/commands/kv/namespace/site.rs
@@ -14,7 +14,6 @@ pub fn site(
     preview: bool,
 ) -> Result<WorkersKvNamespace, failure::Error> {
     kv::validate_target(target)?;
-    let client = kv::api_client(user)?;
 
     let title = if preview {
         format!("__{}-{}", target.name, "workers_sites_assets_preview")
@@ -22,6 +21,10 @@ pub fn site(
         format!("__{}-{}", target.name, "workers_sites_assets")
     };
 
+    // We call CreateNamespace here directly because we want raw access to the
+    // response codes; the implementations in commands::kv::namespace cannot surface
+    // the raw response
+    let client = kv::api_client(user)?;
     let response = client.request(&CreateNamespace {
         account_identifier: &target.account_id,
         params: CreateNamespaceParams {
--- a/src/commands/kv/namespace/site.rs
+++ b/src/commands/kv/namespace/site.rs
@@ -1,5 +1,4 @@
 use cloudflare::endpoints::workerskv::create_namespace::{CreateNamespace, CreateNamespaceParams};
-use cloudflare::endpoints::workerskv::list_namespaces::{ListNamespaces, ListNamespacesParams};
 use cloudflare::endpoints::workerskv::WorkersKvNamespace;
 use cloudflare::framework::apiclient::ApiClient;
 use cloudflare::framework::response::ApiFailure;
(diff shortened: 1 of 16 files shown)


src/commands/kv/namespace/site.rs
```diff
@@ -45,30 +44,16 @@ pub fn site(
...
-                    let params = ListNamespacesParams {
-                        page: Some(PAGE_NUMBER),
-                        per_page: Some(MAX_NAMESPACES_PER_PAGE),
-                    };
-
-                    let response = client.request(&ListNamespaces {
-                        account_identifier: &target.account_id,
-                        params: params,
-                    });
+                    let namespaces = kv::namespace::list::call_api(target, user)?;
```
ashleymichal: i really like this `::call_api` pattern; or even just `::call`. I'd like to incorporate this in other places too, i think it extends the work you did to divorce api calls from cli output. Should it be implemented for `namespace_create` as well to address the issues in the comment above? Seems like a similar refactor.

src/commands/kv/namespace/site.rs
```diff
@@ -9,23 +8,23 @@ use crate::settings::global_user::GlobalUser;
...
 
     let title = if preview {
         format!("__{}-{}", target.name, "workers_sites_assets_preview")
     } else {
         format!("__{}-{}", target.name, "workers_sites_assets")
     };
 
+    // We call CreateNamespace here directly because we want raw access to the
+    // response codes; the implementations in commands::kv::namespace cannot surface
+    // the raw response
```
ashleymichal: should we implement the same `::call_api` pattern here as you have for the `namespace::list` module below?

src/commands/kv/namespace/site.rs
gabbifish: I tried factoring out `::call_api` for `namespace::create`! :D The limitation I ran into was the fact that if site called `::call_api`, it would expect a `ApiError` instead of a `failure::Error`. Perhaps I can wrap the `ApiError` in a `failure:Error` and perform a match over it, so we can extract the `ApiError` and its error code for use by site.rs.

src/commands/kv/namespace/site.rs
gabbifish: I wanted to, but ran into some challenges--check out the message above. I think I can do this by wrapping the `ApiError` in a `failure::Error` and extracting it via match. I'll give that a shot!

src/commands/kv/namespace/site.rs
ashleymichal: hmm. do you think maybe that deserves its own separate analysis as part of client refactoring? it doesn't have to happen here.

ashleymichal: i think the pattern deserves some meta-level attention as far as its interface and what precisely it should take/return. we can defer if that makes more sense.

gabbifish: I agree--this design decision is beginning to bleed into #63. We should think about what error type we specifically want ALL API calls to surface, and make those consistent. Shall we keep this PR on ice for now and use it to drive discussion?

ashleymichal: i'd say at least get in the `&GlobalUser` changes, I think those are solid. Your call on this part; if we're just going to churn we may as well hold off.

gabbifish: I'll minimize this PR to just those `&GlobalUser` changes an open a new WIP one with a proposed approach to API error handling that we can discuss for #63 and #623.

PR #750, opened 2019-09-30
Refactor site.rs calls to KV API (re-use commands::kv module logic)

### LINKED ISSUE #623: Refactor kv:namespace create and kv:namespace list

gabbifish:
This PR fixes #623 and presents a design for #63 overall. 

The design is as follows:

First, if a call to the cloudflare-rs module is always expected to be wrapped in logic (e.g. the same error handling logic, success parsing logic) it is okay for that call to cloudflare-rs to be part of a larger function that returns the expected `Result<some_type, failure::Error>` type.

There are, however, more complex cases where multiple areas of logic make a call to the same cloudflare-rs function, and each caller handles the API result uniquely. I believe that API endpoints that have multiple callers should have their logic exist in a dedicated file (like commands/kv/namespace/create.rs). This file is composed of two items: 
1. A `call_api` function. This function does the minimal setup for calling cloudflare-rs and calls a endpoint using cloudflare-rs. Its return type MUST be `Result<ApiSuccess<expected_success_type>, ApiFailure>`. This function can be called by multiple callers using the API result in different ways (e.g. have different error handling logic). This return signature allows for maximal clarity about what is being returned by the API call. This also means that other helper functions called during `call_api` should not return any sort of `failure::Error`; the only failure possible is an `ApiFailure`.
1. A wrapper function for the "default" use case for this function. This calls `call_api` and does success and error parsing in the way expected by most caller functions. It must return `Result<some_type, failure::Error>`.

Let me know what you think of this design!

### DIFF
--- a/src/commands/kv/bucket/upload.rs
+++ b/src/commands/kv/bucket/upload.rs
@@ -60,14 +60,14 @@ pub fn upload_files(
     while !(pairs.is_empty() && key_value_batch.is_empty()) {
         if pairs.is_empty() {
             // Last batch to upload
-            call_put_bulk_api(target, &user, namespace_id, &mut key_value_batch)?;
+            call_api(target, &user, namespace_id, &mut key_value_batch)?;
         } else {
             let pair = pairs.pop().unwrap();
             if key_count + 1 > PAIRS_MAX_COUNT
             // Keep upload size small to keep KV bulk API happy
             || key_pair_bytes + pair.key.len() + pair.value.len() > UPLOAD_MAX_SIZE
             {
-                call_put_bulk_api(target, &user, namespace_id, &mut key_value_batch)?;
+                call_api(target, &user, namespace_id, &mut key_value_batch)?;
 
                 // If upload successful, reset counters
                 key_count = 0;
@@ -84,7 +84,7 @@ pub fn upload_files(
     Ok(asset_manifest)
 }
 
-fn call_put_bulk_api(
+fn call_api(
     target: &Target,
     user: &GlobalUser,
     namespace_id: &str,
--- a/src/commands/kv/bucket/upload.rs
+++ b/src/commands/kv/bucket/upload.rs
@@ -22,7 +22,7 @@ const UPLOAD_MAX_SIZE: usize = 50 * 1024 * 1024;
 
 pub fn upload_files(
     target: &Target,
-    user: GlobalUser,
+    user: &GlobalUser,
     namespace_id: &str,
     path: &Path,
     exclude_keys: Option<&HashSet<String>>,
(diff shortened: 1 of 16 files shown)


src/commands/kv/mod.rs
```diff
@@ -63,8 +63,8 @@ pub fn get_namespace_id(target: &Target, binding: &str) -> Result<String, failur
     )
 }
 
-fn api_client(user: GlobalUser) -> Result<HttpApiClient, failure::Error> {
-    Ok(HttpApiClient::new(Credentials::from(user)))
+fn api_client(user: &GlobalUser) -> HttpApiClient {
+    HttpApiClient::new(Credentials::from(user.to_owned()))
```
gabbifish: I don't think we need to return a Result here; we don't seem to unwrap anything here.

ashleymichal: hmmm. Legit!

src/commands/kv/bucket/upload.rs
```diff
@@ -84,15 +84,15 @@ pub fn upload_files(
...
 
-fn call_put_bulk_api(
+fn call_api(
     target: &Target,
-    user: GlobalUser,
+    user: &GlobalUser,
     namespace_id: &str,
     key_value_batch: &mut Vec<KeyValuePair>,
 ) -> Result<(), failure::Error> {
     message::info("Uploading...");
```
ashleymichal: i would argue that `call_api` functions should not use message or any other stdout calls, if at all possible. those should be implemented by the caller.

src/commands/kv/namespace/site.rs
```diff
@@ -1,37 +1,25 @@
-use cloudflare::endpoints::workerskv::create_namespace::{CreateNamespace, CreateNamespaceParams};
-use cloudflare::endpoints::workerskv::list_namespaces::{ListNamespaces, ListNamespacesParams};
 use cloudflare::endpoints::workerskv::WorkersKvNamespace;
-use cloudflare::framework::apiclient::ApiClient;
 use cloudflare::framework::response::ApiFailure;
```
ashleymichal: the only weirdness here is that, now we are wrapping the basic functionality of cloudflare-rs, which had been my initial approach, but because the function signatures still need to reference the cloudflare-rs package for its types, it's an incomplete wrapper.

src/commands/kv/bucket/upload.rs
gabbifish: Agree!

src/commands/kv/namespace/site.rs
gabbifish: I feel like this wrapper will be a necessary evil for functions that handle `ApiFailure` with custom logic. Otherwise, callers using the default logic (which typically just prints things out) will not have to deal with this cloudflare-rs logic and can have their API call completely wrapped.

src/commands/kv/bucket/upload.rs
gabbifish: oh wait, this function shouldn't even be called `call_api`; it doesn't reference cloudflare-rs directly at all. Let me rename this; we should consider this another wrapper function around the true API call.

PR #751, opened 2019-09-30
Deprecate release

### LINKED ISSUE #538: Deprecate --release
Related to changes from #386  - This should be done in version 1.5.0

EverlastingBugstopper:
**Breaking Changes!**

Fixes #538 

This PR removes the ability to pass `--release` to `wrangler publish` .

If we plan on doing a point release before 1.5.0 then this should wait.


src/main.rs
```diff
@@ -510,12 +494,12 @@ fn run() -> Result<(), failure::Error> {
         match kv_matches.subcommand() {
             ("create", Some(create_matches)) => {
                 let env = create_matches.value_of("env");
-                let target = manifest.get_target(env, false)?;
+                let target = manifest.get_target(env)?;
```
ashleymichal: nit: fwiw I prefer this approach (`let env ...; let target = manifest.get_target(env)`) to the one-liner

src/settings/target/mod.rs
```diff
@@ -266,88 +251,54 @@ impl Manifest {
...
-                        if let Some(route) = &self.route {
-                            if !route.is_empty() {
-                                failure::bail!(wdd_failure)
-                            }
+        let wdd_failure = "Your environment should only include `workers_dev` or `route`. If you are trying to publish to workers.dev, remove `route` from your wrangler.toml, if you are trying to publish to your own domain, remove `workers_dev`.";
+        let pick_target_failure =
+            "You must specify either `workers_dev` or `route` and `zone_id` in order to publish.";
+
+        if let Some(env) = &environment {
+            if let Some(wdd) = env.workers_dev {
```
ashleymichal: nit: the short hand keeps tripping me up.
```suggestion
            if let Some(workers_dev) = env.workers_dev {
```

src/settings/target/mod.rs
```diff
@@ -266,88 +251,54 @@ impl Manifest {
...
+                if let Some(route) = &env.route {
```
gabbifish: With the removal of the release flag, I think we can make this truth table logic even simpler. I'm thinking we could have mutable variables, `route` and `workers_dev`, that are initially populated with the default environment's values and then updated with a selected environment's values. This way, we could get rid of these complex nested matches in favor of a simple series of conditionals. What do you think?

src/settings/target/mod.rs
```diff
@@ -208,7 +193,7 @@ impl Manifest {
 
         self.check_private(environment);
 
-        let (route, workers_dev) = self.negotiate_zoneless(environment, release)?;
+        let (route, workers_dev) = self.negotiate_zoneless(environment)?;
```
ashleymichal: i wonder if we could benefit from a type here that's just `DeployConfig`; an enum that can be either Zoneless, or a struct, Zoned, with a zone_id and route, and include it as `target.deploy_config` rather than three attributes of `target` that we incessantly check.

src/settings/target/mod.rs
```diff
@@ -266,88 +251,54 @@ impl Manifest {
         }
     }
 
-    // TODO: when --release is deprecated, this will be much easier
     fn negotiate_zoneless(
         &self,
         environment: Option<&Environment>,
-        release: bool,
     ) -> Result<(Option<String>, bool), failure::Error> {
```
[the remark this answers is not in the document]
ashleymichal: this sounds correct to me

docs/content/environments.md
```diff
@@ -257,7 +253,6 @@ type = "webpack"
 account_id = "12345678901234567890"
 zone_id = "09876543210987654321"
 route = "example.com/*"
-workers_dev = false
```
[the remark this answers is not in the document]
EverlastingBugstopper: Currently we only require `workers_dev` in the top level due to the ambiguous nature of the `--release` flag. Since we are deprecating `--release`, this is no longer necessary since we know that if there is a `route` present, they are not deploying to their workers.dev subdomain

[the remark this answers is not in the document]
EverlastingBugstopper: Agreed we should make this clear in the changelog

src/settings/target/mod.rs
EverlastingBugstopper: I like this - but I think I like Gabbi's approach of making Route an Option instead, gonna do that

src/settings/target/mod.rs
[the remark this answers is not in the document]
EverlastingBugstopper: going to take this approach

src/settings/target/mod.rs
ashleymichal: we could likely do both, but mine is a bit more... involved. gabbi's is likely the better answer

src/settings/target/mod.rs
EverlastingBugstopper: I like this in theory, but everything is an `Option` and we need decent error messages for when there are conflicting values... I'm not exactly sure how to go about this. You can't just `let mut route = environment.route` at the beginning because `environment.route` is an `Option`. and so is `environment.workers_dev`. I think I can simplify it a bit but they will still be match statements

src/settings/target/mod.rs
```diff
@@ -266,88 +247,73 @@ impl Manifest {
...
+        let (workers_dev, route) = if let Some(env) = &environment {
+            log::debug!("env workers_dev: {:?}", env.workers_dev);
+            log::debug!("env route: {:?}", env.route);
+            match (env.workers_dev, env.route.clone()) {
+                (None, Some(route)) => {
+                    if top_workers_dev && route.is_empty() {
+                        failure::bail!(empty_route_failure)
                     } else {
-                        wdd // inherit from top level
+                        (false, Some(route))
```
gabbifish: so if there is a route in the environment, that always overrides whatever setting is in the top level configs. I'd LOVE to see a full, drawn out truth table for this in a comment before this code section, it would be great for maintainers!

src/settings/target/mod.rs
```diff
@@ -266,88 +247,73 @@ impl Manifest {
...
+            }
+            (false, Some(route)) => {
+                if route.is_empty() {
+                    failure::bail!(pick_target_failure)
+                } else {
+                    Ok(Some(route))
+                }
+            }
+            (false, None) => failure::bail!(pick_target_failure),
+        }
```
gabbifish: I feel like this match statement can be completely consolidated with the one above :) 

But beyond this, I think we could do away with matches entirely and just rely on `is_none()` and `is_some()` to deduce which outcome we have for deployment target. I think this would be ultimately more readable than very long match statements and very explicitly display the conditions leading to a truth table assessment (they would all be in the conditional clause, which could really help with readability).

src/settings/target/mod.rs
```diff
@@ -266,88 +247,73 @@ impl Manifest {
...
+            (true, Some(route)) => {
+                if route.is_empty() {
+                    Ok(None)
+                } else {
+                    failure::bail!(conflicting_targets_failure)
+                }
+            }
+            (false, Some(route)) => {
+                if route.is_empty() {
+                    failure::bail!(pick_target_failure)
```
gabbifish: why does this case above not throw the same error?

src/settings/target/mod.rs
EverlastingBugstopper: I don't think that `is_none()` and `is_some()` is what we want here because they always need to be paired with `unwrap()`, not going to be any more readable imo

src/settings/target/mod.rs
EverlastingBugstopper: The first two match statements boil down the two variables and then the last statement figures out what the function should return

src/settings/target/mod.rs
EverlastingBugstopper: agreed! i can draw this out

EverlastingBugstopper:
I've been talking with the Ashley's about this PR and the general feeling is that this is an improvement to what is there already, and this PR is intended to deprecate release and not to entirely rework how we do this. I'm open to getting a refactor in (I think the way to do this is to split stuff out into a deploy config struct of some sort) which will improve readability and maintainability but for now I think I should probably just add some more comments. (yay for small incremental improvements) :) 

@gabbifish if you feel like taking on a big refactor feel free and let me know if you have questions - we have some time before 1.5.0 so I'm going to step back from this PR and work on some other stuff

EverlastingBugstopper:
https://twitter.com/yaahc_/status/1180885289742434305?s=19

src/settings/target/mod.rs
```diff
@@ -270,88 +251,73 @@ impl Manifest {
...
+        let conflicting_targets_failure = "Your environment should only include `workers_dev` or `route`. If you are trying to publish to workers.dev, remove `route` from your wrangler.toml, if you are trying to publish to your own domain, remove `workers_dev`.";
+        let pick_target_failure =
+            "You must specify either `workers_dev` or `route` and `zone_id` in order to publish.";
+        let empty_route_failure =
+            "If you want to deploy to workers.dev, remove `route` from your environment config.";
+
+        log::debug!("top level workers_dev: {:?}", self.workers_dev);
+        log::debug!("top level route: {:?}", self.route);
+
+        // start with top level configuration
```
gabbifish: I still would LOVE a comment that displays all the expected combinations and outputs for the truth table... think you can slide that in before I approve this?

EverlastingBugstopper: oh oops - totally!

PR #757, opened 2019-10-02
Use custom KV api client to support longer timeouts

### LINKED ISSUE #746: Wrangler timeout too short for bulk uploads
#### Environment

* operating system:
* output of `rustc -V`:
* output of `node -v`:
v10.16.2
* output of `wrangler -V`:
👷 ✨  wrangler 1.4.0

#### Steps to reproduce

I've been trying the newly added Workers Sites as it looked awesome. I tried it first with a simple site with about 10 assets. Worked perfectly, first time. Dev, Prod, all fine. Yay. So I then tried it with _another_ site with more pages and many more assets (mainly images). This time, the publish failed :(

It seems to be at the point where it uploads all the site's static files to KV. The time out seems too short ... can it be increased? I guess this is outside of wrangler but maybe that request could be passed on. Since I can see it using the /bulk operation which makes sense. However I'd think someone would only be using the /bulk API if they were uploading a lot and so would need a generous timeout, even _outside_ of Workers Sites.

It's not clear from the error if it is a timeout with one file (like too big, and so timing out) or the entire operation (too many files within its allowed time)

I'd guess there are maybe 500 files.

The publish to successfully complete

$ wrangler publish --env dev
🌀  Using namespace for Workers Site "__name-dev-workers_sites_assets"
💁  ignoring file .DS_Store
💁  ignoring file .DS_Store
💁  Uploading...
Error: ⚠️  Error: https://api.cloudflare.com/client/v4/accounts/longidhere/storage/kv/namespaces/longidhere/bulk: timed out

gabbifish:
This PR closes #746. 

It uses a newer version of cloudflare-rs that allows us to specify custom configurations for our API client. This allows us to specify a longer timeout for the client and wait longer for a bulk API upload to complete. This PR is blocked on cloudflare-rs being re-released so we can point to a version number instead of the latest master commit on the cloudflare-rs repository.

Given that this cloudflare-rs change is breaking (it now returns clients wrapped in a `Result`) we had to change where and how clients were constructed in Wrangler. Namely, clients can no longer be constructed in `call_api` functions anymore, because these functions must return `Result<ApiSuccess<_>, ApiFailure>` (not `failure::Error)`. Instead, clients are now passed in to `call_api` calls.

### DIFF
--- a/src/commands/kv/mod.rs
+++ b/src/commands/kv/mod.rs
@@ -1,8 +1,9 @@
 use std::collections::HashSet;
+use std::time::Duration;
 
 use cloudflare::framework::auth::Credentials;
 use cloudflare::framework::response::ApiFailure;
-use cloudflare::framework::HttpApiClient;
+use cloudflare::framework::{HttpApiClient, HttpApiClientConfig};
 
 use http::status::StatusCode;
 use percent_encoding::{percent_encode, PATH_SEGMENT_ENCODE_SET};
@@ -63,8 +64,14 @@ pub fn get_namespace_id(target: &Target, binding: &str) -> Result<String, failur
     )
 }
 
-fn api_client(user: &GlobalUser) -> HttpApiClient {
-    HttpApiClient::new(Credentials::from(user.to_owned()))
+fn api_client(user: &GlobalUser) -> Result<HttpApiClient, failure::Error> {
+    HttpApiClient::new(
+        Credentials::from(user.to_owned()),
+        HttpApiClientConfig {
+            // Use 3 minute timeout instead of default 30-second one.
+            http_timeout: Duration::from_secs(3 * 60),
+        },
+    )
 }
 
 fn format_error(e: ApiFailure) -> String {
--- a/src/commands/kv/mod.rs
+++ b/src/commands/kv/mod.rs
@@ -68,8 +68,9 @@ fn api_client(user: &GlobalUser) -> Result<HttpApiClient, failure::Error> {
     HttpApiClient::new(
         Credentials::from(user.to_owned()),
         HttpApiClientConfig {
-            // Use 3 minute timeout instead of default 30-second one.
-            http_timeout: Duration::from_secs(3 * 60),
+            // Use 5 minute timeout instead of default 30-second one.
+            // This is useful for bulk upload operations.
+            http_timeout: Duration::from_secs(5 * 60),
         },
     )
 }
@@ -77,7 +78,7 @@ fn api_client(user: &GlobalUser) -> Result<HttpApiClient, failure::Error> {
 fn format_error(e: ApiFailure) -> String {
     match e {
         ApiFailure::Error(status, api_errors) => {
-            give_status_code_context(status);
+            print_status_code_context(status);
             let mut complete_err = "".to_string();
             for error in api_errors.errors {
                 let error_msg =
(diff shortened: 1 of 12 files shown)


src/commands/kv/mod.rs
```diff
@@ -63,8 +64,14 @@ pub fn get_namespace_id(target: &Target, binding: &str) -> Result<String, failur
...
 }
 
-fn api_client(user: &GlobalUser) -> HttpApiClient {
-    HttpApiClient::new(Credentials::from(user.to_owned()))
+fn api_client(user: &GlobalUser) -> Result<HttpApiClient, failure::Error> {
+    HttpApiClient::new(
+        Credentials::from(user.to_owned()),
+        HttpApiClientConfig {
+            // Use 3 minute timeout instead of default 30-second one.
+            http_timeout: Duration::from_secs(3 * 60),
```
ashleymichal: one thing to note: the KV API endpoints vary in _their_ configured timeout; we should make sure we know what it looks like when the KV api times out, vs when we time out, and how wrangler handles that.

gabbifish: I looked into this. :) A timeout from the reqwest client looks like
```
Error: Error: https://api.cloudflare.com/client/v4/accounts/c2c4de4232cdae687dc7a4a0323df63a/storage/kv/namespaces: timed out
```
(^ this is what the original bug reporter observed).

A timeout from Cloudflare's end (e.g. the worker lags on fetching something from the backend) is represented by status code 504. I added new handling logic for this, so that Wrangler users are advised to wait a few seconds and re-run their command.

src/commands/kv/mod.rs
```diff
@@ -119,9 +120,13 @@ fn url_encode_key(key: &str) -> String {
 
 // For handling cases where the API gateway returns errors via HTTP status codes
 // (no KV error code is given).
-fn give_status_code_context(status_code: StatusCode) {
-    if let StatusCode::PAYLOAD_TOO_LARGE = status_code {
-        message::warn("Returned status code 413, Payload Too Large. Make sure your upload is less than 100MB in size")
+fn print_status_code_context(status_code: StatusCode) {
+    match status_code {
+        StatusCode::PAYLOAD_TOO_LARGE =>
```
EverlastingBugstopper: Is the payload in question the one that we upload to the workers api? If so I'm not sure quite how helpful this error messsage is - there isn't really an actionable here that relates to what the user understands wrangler to be doing. "Payload" seems to be an implementation detail.

src/commands/kv/mod.rs
```diff
@@ -119,9 +120,13 @@ fn url_encode_key(key: &str) -> String {
...
+            message::warn("Returned status code 413, Payload Too Large. Make sure your upload is less than 100MB in size"),
+        StatusCode::GATEWAY_TIMEOUT =>
```
EverlastingBugstopper: nit: i'm used to these commands saying "please" try again in a few seconds 😉

src/commands/kv/mod.rs
gabbifish: Ideally this error code is never even hit via wrangler--we actually enforce a maximum upload size of 50MB! This is just conservative error handling out of an abundance of caution.

EverlastingBugstopper: Awesome! Could you add a comment saying this :)

gabbifish: for sure!

src/commands/kv/mod.rs
```diff
@@ -112,9 +120,15 @@ fn url_encode_key(key: &str) -> String {
 
 // For handling cases where the API gateway returns errors via HTTP status codes
 // (no KV error code is given).
-fn give_status_code_context(status_code: StatusCode) {
-    if let StatusCode::PAYLOAD_TOO_LARGE = status_code {
-        message::warn("Returned status code 413, Payload Too Large. Make sure your upload is less than 100MB in size")
+fn print_status_code_context(status_code: StatusCode) {
+    match status_code {
+        // Folks should never hit PAYLOAD_TOO_LARGE, given that Wrangler ensures that bulk file uploads
+        // are max ~50 MB in size. This case is handled anyways out of an abundance of caution.
```
ashleymichal: <3

PR #759, opened 2019-10-03
Use template wrangler.toml when generating

### LINKED ISSUE #315: Don’t infer project type from the template name
Currently in the docs, we suggest that people run `wrangler generate myApp https://github.com/cloudflare/rustwasm-worker-template` to generate our default rust app. This works because there is a code section that checks to see if `rust` is in the name of the template, and if it is, it will assume the user wants a project with `type=rust`. This assumption is shaky and has led to some code sections that are hard to read.

We should stop inferring the project type from the name and either infer from the project structure itself, or require a type to be passed when generating a new project.

There’s a PR for this here: #314 but it breaks one of the commands we suggest on our docs page: `wrangler generate myApp https://github.com/cloudflare/rustwasm-worker-template`

The reason it breaks is because without a specified type, it defaults to webpack, and our build process makes assumptions about where to look for the javascript, so it can’t build the project w/o the type being set to `rust`. 

This change _does_ work however if you simply run `wrangler generate —type rust` and is also more succinct than what we currently have in our docs

Blocked by #313 as it makes changes to this particular section of code

EverlastingBugstopper:
This PR will take every value from a template's wrangler.toml on generate, and if it's not there, it will fill in a default value. It also prints out helpful messages for when users need to fill in account specific information such as account_id, zone_id, route.

Additionally, it stops users from passing a template along with `--site` to `wrangler generate` because it does not behave as the user would expect.

**TODO**: QA wrangler generate 

- with each of our templates
- try making your own repo with a weird toml!

Fixes #315 and #759


src/commands/generate.rs
```diff
@@ -68,10 +61,3 @@ fn command(name: &str, binary_path: PathBuf, args: &[&str], project_type: &Targe
     c.args(args);
     c
 }
-
-fn get_target_type(template: &str) -> TargetType {
-    if template.contains("rust") {
-        return TargetType::Rust;
-    }
-    TargetType::default()
-}
```
EverlastingBugstopper: this is the highlight of my career

EverlastingBugstopper: throwback image

src/settings/target/mod.rs
```diff
@@ -33,6 +33,22 @@ impl Site {
     }
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct TemplateConfig {
+    #[serde(rename = "type")]
+    pub target_type: TargetType,
+    pub webpack_config: Option<String>,
```
gabbifish: perhaps we can explicitly specify that this is a path? I think this would be helpful and provide more clarity.

src/settings/target/mod.rs
```diff
@@ -140,27 +156,48 @@ impl Manifest {
...
         site: Option<Site>,
     ) -> Result<Manifest, failure::Error> {
+        let config_file = config_path.join("wrangler.toml");
+        let template_config_content = fs::read_to_string(&config_file);
+        let template_config = match &template_config_content {
+            Ok(content) => {
+                let config: TemplateConfig = toml::from_str(content)?;
+                if let Some(target_type) = target_type.clone() {
+                    if config.target_type != target_type {
+                        message::warn(&format!("The template recommends the \"{}\" type. Using type \"{}\" may cause errors, we recommend changing the type field in wrangler.toml to \"{}\"", config.target_type, target_type, config.target_type));
```
gabbifish: ideally this edge case never even gets triggered, given that our templates should now have wrangler.toml in them!

EverlastingBugstopper: This case is for when users pass `--type` to `wrangler generate` and the template has a type specified in the toml

EverlastingBugstopper: The behavior is to override the template's toml with the type the user passed in and print this warning. Open to arguments saying you shouldn't be able to override

gabbifish: Makes sense :)

src/settings/target/mod.rs
```diff
@@ -140,27 +156,48 @@ impl Manifest {
...
+                    }
+                }
+                Ok(config)
+            }
+            Err(err) => Err(err),
+        };
+        let template_config = match template_config {
+            Ok(config) => config,
+            Err(err) => {
```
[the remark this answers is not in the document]
EverlastingBugstopper: I feel like we shouldn't punish the user for something they likely didn't configure - the error should be in the debug logs but otherwise it will fall back to the default which is `webpack_config`. This gives wrangler some flexibility when it comes to generating from templates that may not be maintained by us and don't contain a `wrangler.toml`

src/settings/target/mod.rs
EverlastingBugstopper: Out of scope for this - the `webpack_config` type on `Manifest` is an `Option<String>` so that's what it is here

src/settings/target/mod.rs
```diff
@@ -140,27 +156,48 @@ impl Manifest {
...
+        target_type: Option<TargetType>,
         config_path: PathBuf,
         site: Option<Site>,
     ) -> Result<Manifest, failure::Error> {
+        let config_file = config_path.join("wrangler.toml");
+        let template_config_content = fs::read_to_string(&config_file);
+        let template_config = match &template_config_content {
+            Ok(content) => {
+                let config: TemplateConfig = toml::from_str(content)?;
+                if let Some(target_type) = target_type.clone() {
```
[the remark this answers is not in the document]
EverlastingBugstopper: Nope :/

```
error[E0308]: mismatched types
   --> src/settings/target/mod.rs:169:46
    |
169 |                     if config.target_type != target_type {
    |                                              ^^^^^^^^^^^ expected enum `settings::target::target_type::TargetType`, found reference
    |
    = note: expected type `settings::target::target_type::TargetType`
               found type `&settings::target::target_type::TargetType`
```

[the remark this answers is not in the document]
gabbifish: Hehehe something you can do is ```rs
if config.target_type != *target_type
```

File: src/settings/target/mod.rs (hunk 4 above)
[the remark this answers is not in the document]
Comment (gabbifish) [reply src/settings/target/mod.rs]: I feel like being explicit about the error would help users know when a template is broken and requires a ticket/issue... what do you think?

File: src/settings/target/mod.rs (hunk 2 above)
Comment (gabbifish) [reply src/settings/target/mod.rs]: gotcha!

File: src/settings/target/mod.rs (hunk 5 above)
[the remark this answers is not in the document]
Comment (EverlastingBugstopper) [reply src/settings/target/mod.rs]: 🤯

File: src/settings/target/mod.rs (hunk 4 above)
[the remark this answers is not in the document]
Comment (EverlastingBugstopper) [reply src/settings/target/mod.rs]: Iiiii think I disagree. Don't think we should block our users for something that's probably not their fault and I also don't think we should require template creators to have a `wrangler.toml` with these values filled in.

**Comment by victoriabernard92** [comment]
The problems https://github.com/cloudflare/wrangler/issues/315, https://github.com/cloudflare/wrangler/issues/396 and future problems with `generate` could be solved if we just copy over the `wrangler.toml` file. 

The only CON with copying the toml file is the account/zone tags, but the end user must adjust those fields anyways and none fo the templates in the gallery will populate those fields. 

It would dramatically increase the usefulness to the end user and template developer by having all the wrangler config copied in projects. Especially as the wrangler.toml gains complexity (e.g. sites, import/export, buckets.. )

**Comment by EverlastingBugstopper** [comment]
I can see this being useful - we would need to strip namespace ids from any kv namespace, zone id, account id, and replace the name with the name they passed. 

Here's a question though. What happens if they pass `--site` to `wrangler generate` and the template has the site already configured?

**Comment by victoriabernard92** [comment]
> I can see this being useful - we would need to strip namespace ids from any kv namespace, zone id, account id, and replace the name with the name they passed.
> 
> Here's a question though. What happens if they pass `--site` to `wrangler generate` and the template has the site already configured?

Not sure I understand, are you saying like:
```
wrangler generate temp --site https://github...
```
why would someone use `--site` with a git repo on generate? Do we support that now? 

If we simply copied the wrangler.toml from the repo we wouldn't need to worry about the `--site` at all. It should be an invalid flag.

**Comment by EverlastingBugstopper** [comment]
Yeah that's what I mean - I agree that if we copied everything over we should make any `--site` that is not `wrangler generate --site` fail. That should probably already be happening because if you run `wrangler generate my-app https://github.com/cloudflare/worker-template --site` it creates the site directory but completely ignores the template.

**Comment by victoriabernard92** [comment]
I started a proposal with some user stories I'd love your feedback/contribution on @EverlastingBugstopper https://wiki.cfops.it/display/~victoria/Proposal+%3A+init+and+generate

**Comment by EverlastingBugstopper** [comment]
I've updated this PR with relevant info - the entire toml will be copied and if there are values that contain user-specific info, we will warn them about those fields and inform them that they should fill in their own values.

**Comment by EverlastingBugstopper** [comment]
Running into a very interesting error with this

```console
$ wrangler generate my-rust https://github.com/EverlastingBugstopper/rustwasm-worker-template/
🌀  🐑  Generating a new worker project with name 'my-rust'...
🔧   Creating project called `my-rust`...
✨   Done! New project created /Users/averyharnish/Documents/work/_tmp/my-rust
⚠️  The template you generated from includes some account specific information that you may need to update with your own values.
Error: values must be emitted before tables
```

The last message is the most interesting and is related to how serde handles writing out HashMaps to tomls. Not sure what to do about fixing this.

EverlastingBugstopper:
Environment thing still not working but I was able to implement some errors for when user-specific info is included in a template manifest

https://github.com/EverlastingBugstopper/rustwasm-worker-template/ has a `wrangler.toml` that looks like this

```toml
type = "rust"
account_id = "12234"
route = "top-level"

[env.foo]
workers_dev = true

[env.bar]
account_id = "12345"
zone_id = "12345"
route = "bar-env"
```

when I generate with this template, this is the error message

```console
wrangler generate my-rust https://github.com/EverlastingBugstopper/rustwasm-worker-template/
🌀  🐑  Generating a new worker project with name 'my-rust'...
🔧   Creating project called `my-rust`...
✨   Done! New project created /Users/averyharnish/Documents/work/_tmp/my-rust
⚠️  Your generated project contains user-specific information that you will likely need to change.
⚠️  Top level fields: account_id, route
⚠️  [env.bar] fields: account_id, route, zone_id
```

Open to suggestions on better ways to print these warnings

victoriabernard92:
> Open to suggestions on better ways to print these warnings

What about shortening it to help developers action on it?:
> Replace all account-specific info in your wrangler.toml

EverlastingBugstopper:
I've updated the messages to look like this:

```console
$ wrangler generate my-rust https://github.com/EverlastingBugstopper/rustwasm-worker-template/
🌀  🐑  Generating a new worker project with name 'my-rust'...
🔧   Creating project called `my-rust`...
✨   Done! New project created /Users/averyharnish/Documents/work/_tmp/my-rust
⚠️  Replace all account specific info in your wrangler.toml.
⚠️  Your zone_id and account_id can be found in the right sidebar at https://dash.cloudflare.com

Top level fields
  - account_id
  - route
[env.bar]
  - account_id
  - route
  - zone_id
```

I'll be out next week, the work that still needs to be done on this is:

- Add wrangler.tomls to each of our templates (especially rust)
 - I think it makes sense for each of these to have `workers_dev = true` and `route`/`zone_id` to be `= ""` so they can fill those in
- Fix the error message "Error: values must be emitted before tables"
 - this can be reproduced with any toml that has environments in it, but an easy one is `wrangler generate my-rust https://github.com/EverlastingBugstopper/rustwasm-worker-template/`
- QA `wrangler generate` with each of our templates and make sure there are no warnings and that they can all be `wrangler publish`

src/main.rs
```diff
@@ -396,29 +396,27 @@ fn run() -> Result<(), failure::Error> {
...
+        let mut target_type = None;
 
-        let (target_type, template) = if site {
-            // Workers Sites projects are always Webpack for now
-            let target_type = Some(TargetType::Webpack);
-            let template = "https://github.com/cloudflare/worker-sites-template";
-
-            (target_type, template)
+        let template = if site {
+            if template.is_some() {
```
ashleymichal: hesitant to keep this gate but i may be in the minority; folks wanting to write custom site workers should be encouraged, shouldn't they? perhaps rather than gating this we should expose the api? idk.

EverlastingBugstopper: Right now if you try to run this command it is equivalent to `wrangler generate --site` -- if people want to use their own boilerplate instead of ours that's fine, but it won't work with the `--site` flag. (you can try it yourself! the behavior is super wonky)

EverlastingBugstopper: This PR also allows templates to include `[site]` sections in their tomls, so the `--site` flag wouldn't be needed to create that section of the toml if the template was properly configured

ashleymichal: sure, but is that because `--site` is being too special?

EverlastingBugstopper: yes. it's essentially an alias to a specific template, but that's our API for now. open to looking at ways to fix this but it's not gonna make it in this PR

src/settings/target/manifest.rs
```diff
@@ -52,27 +53,64 @@ impl Manifest {
 
     pub fn generate(
         name: String,
-        target_type: TargetType,
+        target_type: Option<TargetType>,
```
ashleymichal: why make this an option type when we end up falling back to a default in most cases?

src/settings/target/template_config.rs
```diff
@@ -0,0 +1,112 @@
+use std::collections::HashMap;
+
+use serde::{Deserialize, Serialize};
+
+use crate::terminal::message;
+
+use super::{Environment, KvNamespace, Site, TargetType};
+
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+pub struct TemplateConfig {
```
ashleymichal: why make a new struct for this rather than using manifest?

src/settings/target/manifest.rs
EverlastingBugstopper: This is an `Option` because `--type` is an optional flag

src/settings/target/template_config.rs
EverlastingBugstopper: oh. that's a really good point. initially i introduced this new struct because it was only taking in `type` and `webpack_config` but now since we're taking everything it makes sense to use `Manifest`

src/settings/target/manifest.rs
ashleymichal: it's always been an optional flag, though; in the past we've simply fallen back to its default value before calling the command.

src/settings/target/template_config.rs
ashleymichal: ah. makes sense. yah i think a lot of logic might be simplified if the logic was just "pull manifest and overwrite some 💩 "

EverlastingBugstopper: Oop - just kidding! spoke too soon! Everything in `TemplateConfig` is an `Option` so that if it's omitted from a given template, wrangler generate does not even give a care. We don't want everything to be an `Option` in `Manifest` though.

src/settings/target/manifest.rs
EverlastingBugstopper: Yes, this PR changes the point in the code where we determine the type. This is now just the value passed in by the flag, then we clone the repo, then we parse the toml and do the type check there. Any type passed in here will override a type that's in the template toml, and warn if it is different.

EverlastingBugstopper:
Thanks so much @victoriabernard92!

src/settings/target/template_config.rs
ashleymichal: alright. i think there are refactors we can do to dedup a lot of this stuff but they are not for this PR.

src/settings/target/manifest.rs
```diff
@@ -52,27 +53,64 @@ impl Manifest {
...
+        let template_config = match template_config {
+            Ok(config) => config,
+            Err(err) => {
+                log::info!("Error parsing template {}", err);
+                log::debug!("template content {:?}", template_config_content);
+                TemplateConfig::default()
+            }
+        };
+
+        let default_workers_dev = if let Some(route) = &template_config.route {
```
gabbifish: why not just do a nested match here instead of using `if let`? I think that would be more readable

gabbifish: rule of thumb I've used: if `if let` is followed by an `else`, that would be better written as a match

src/settings/target/template_config.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub webpack_config: Option<String>,
+    pub workers_dev: Option<bool>,
+    pub zone_id: Option<String>,
+    pub site: Option<Site>,
+}
+
+impl TemplateConfig {
+    pub fn warn_on_account_info(&self) {
+        let mut top_level_fields: Vec<String> = Vec::new();
+        if let Some(account_id) = &self.account_id {
```
gabbifish: If &self.account_id is none, shouldn't we also add this to the list of missing top level fields?

src/settings/target/template_config.rs
gabbifish: I still think there is a way we can use Manifest in place of a new, duplicative TemplateConfig struct... (let's have our cake and eat it too!). There should be a serde option that allows something that isn't present to be deserialized as an empty string (check out https://serde.rs/field-attrs.html, specifically a variant of `#[serde(default)]`.

src/settings/target/template_config.rs
gabbifish: ideally this field won't even be an option though (let's re-use Manifest).

src/commands/build/wranglerjs/mod.rs
```diff
@@ -233,11 +233,7 @@ pub fn scaffold_site_worker(target: &Target) -> Result<(), failure::Error> {
 
     if !Path::new(&build_dir).exists() {
         // TODO: use site.entry_point instead of build_dir explicitly.
-        run_generate(
-            build_dir.file_name().unwrap().to_str().unwrap(),
-            template,
-            &target.target_type,
-        )?;
+        run_generate(build_dir.file_name().unwrap().to_str().unwrap(), template)?;
```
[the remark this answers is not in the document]
EverlastingBugstopper: i'm not sure. only reason this came up in changelog is because of formatting.

src/settings/target/template_config.rs
EverlastingBugstopper: we want to be able to run generate on invalid tomls which is why everything is an option

EverlastingBugstopper: previously wrangler generate worked without a toml altogether, so we want to be able to support those cases

gabbifish: ^ this is why i suggest using serde tricks to handle empty fields without using options :)

EverlastingBugstopper:
I agree that it would be good to re-use Manifest instead of introducing this new config but there are a couple of key differences:

namely `TemplateConfig` does not take `name`, at all, and `account_id` is an `Option`. We create a `Manifest` from a `TemplateConfig`

EverlastingBugstopper:
oops

EverlastingBugstopper:
I've updated the output given by wrangler generate. It will no longer warn people to fill in account_id/zone_id if they have the environment variables. I've also removed the newline at the beginning of the fields. Additionally, the dashboard link is much more helpful than just "dash.cloudflare.com"

```console
$ wrangler generate my-rust https://github.com/EverlastingBugstopper/rustwasm-worker-template/
🔧   Creating project called `my-rust`...
✨   Done! New project created /Users/averyharnish/Documents/work/_tmp/my-rust
🕵️  Your zone_id and account_id can be found in the right sidebar at https://dash.cloudflare.com/?account=workers
🕵️  You will need to update the following fields in the created wrangler.toml file before continuing:
- account_id
- kv-namespace FOO
- kv-namespace BAR
- route

[env.foo]
  - kv-namespace FOOFOO
  - kv-namespace FOOBAR

[env.bar]
  - account_id
  - kv-namespace BARFOO
  - kv-namespace BARBAR
  - route
  - zone_id
```

```console
$ wrangler generate
🔧   Creating project called `worker`...
✨   Done! New project created /Users/averyharnish/Documents/work/_tmp/worker
🕵️  Your zone_id and account_id can be found in the right sidebar at https://dash.cloudflare.com/?account=workers
🕵️  You will need to update the following fields in the created wrangler.toml file before continuing:
- account_id
```

PR #760, opened 2019-10-03
[Wrangler Ignore option 2] exclude/include logic in wrangler.toml for static asset uploads

### LINKED ISSUE #716: RFC: .wranglerignore

gabbifish:
This PR is the secondary implementation/option to close #716. The first option is #720.

Wrangler will look at `wrangler.toml`'s `site` section for fields to ignore:
```toml
[site]
bucket = "./public"
entry-point = "workers-site" 
include = ["included_dir"] # optional
exclude = ["excluded_dir"] # optional
```
The include and exclude fields will apply to the `bucket` directory. Its behavior above is exactly the same as https://github.com/ashleygwilliams/cargo-generate#include--exclude (this is also the standard for cargo).

Keep in mind that, consistent with other implementations of include/exclude, the include field always overrides the exclude one.

This PR also contains a bunch of unit tests to ensure that our exclude/include logic works as expected.


src/commands/kv/bucket/mod.rs
```diff
@@ -15,22 +15,24 @@ use std::path::Path;
...
     verbose: bool,
 ) -> Result<(Vec<KeyValuePair>, AssetManifest), failure::Error> {
     let mut upload_vec: Vec<KeyValuePair> = Vec::new();
     let mut asset_manifest: AssetManifest = AssetManifest::new();
 
-    for entry in WalkDir::new(directory)
-        .into_iter()
-        .filter_entry(|e| !is_ignored(e))
-    {
+    let dir_walker = get_dir_iterator(target, directory)?;
```
EverlastingBugstopper: nit: is it a dir_walker or a dir_iterator?

src/commands/kv/bucket/mod.rs
```diff
@@ -60,9 +62,12 @@ pub fn directory_keys_values(
...
 
 // Returns only the hashed keys for a directory's files.
-fn directory_keys_only(directory: &Path) -> Result<Vec<String>, failure::Error> {
+fn directory_keys_only(target: &Target, directory: &Path) -> Result<Vec<String>, failure::Error> {
     let mut upload_vec: Vec<String> = Vec::new();
-    for entry in WalkDir::new(directory) {
+
+    let dir_walker = get_dir_iterator(target, directory)?;
+
+    for entry in dir_walker {
```
EverlastingBugstopper: what can an `entry` be? mostly just curious

src/commands/kv/bucket/mod.rs
```diff
@@ -79,34 +84,36 @@ fn directory_keys_only(directory: &Path) -> Result<Vec<String>, failure::Error>
...
+        required_override.add(&format!("!{}", ignored))?;
     }
 
-    // Then, ensure files in ignored directories are also ignored.
-    for dir in KNOWN_UNNECESSARY_DIRS {
-        if stem == *dir {
-            // Need to check for full equality here
-            message::info(&format!("ignoring directory {}", dir));
-            return true;
+    if let Some(s) = &target.site {
```
EverlastingBugstopper: nit: could you rename `s` to `site`

src/commands/kv/bucket/mod.rs
gabbifish: i guess they are kind of the same thing--the walker iterates through the directory entries :)

src/commands/kv/bucket/mod.rs
gabbifish: Entry is any item found within a directory, whether it is a dir or file. it also can include symlinks.

src/commands/kv/bucket/mod.rs
gabbifish: sure!

gabbifish:
@EverlastingBugstopper responding to questions:
1. There is no output about what's being ignored, given that the user explicitly already states what should be ignored in the `wrangler.toml` file. I can definitely add some info text about `node_modules` being ignored, though.
1. You can't really use `.gitignore` with the paths like `./img`--it's a bit of an antipattern, you usually just provide `img/`. The same applies for this approach with `exclude/include`. This is because these types of ignore patterns can come from a `.ignore` file anywhere in a project, so using relative paths dependent on `.` is unsafe.
1. I can definitely make the error message a bit better and remind folks to use an array if they just provide a string.
1. I'll add info logs for everything ignored. :)

EverlastingBugstopper:
Makes sense - re no. 2, do you think it makes sense to print a `message::warn` when something that is in an `include` or an `exclude` is not recognized? is that possible? like "hey looks like you wanna ignore this but i have no idea what that even is". maybe not necessary but might be helpful. the info logs might be enough for this

gabbifish:
@EverlastingBugstopper I might wait until we see a ticket related to git entries starting with `./` -- I feel like this case is unlikely, but if an issue is filed, I will definitely add additional hand-holding logic!

Issue #668, opened 2019-09-23
Errors occasionally have double "Error" string


EverlastingBugstopper:
#### Steps to reproduce

Use a bad connection / force timeout on kv uploading

#### What did you expect to see?

```console
$ wrangler publish

🌀  Creating namespace for Workers Site "__www-workers_sites_assets"

Error: https://api.cloudflare.com/client/v4/accounts/f16e4749531f6b3cd0aa857aa043f4f3/storage/kv/namespaces: timed out
```

#### What did you see instead?

```console
$ wrangler publish

🌀  Creating namespace for Workers Site "__www-workers_sites_assets"

Error: Error: https://api.cloudflare.com/client/v4/accounts/f16e4749531f6b3cd0aa857aa043f4f3/storage/kv/namespaces: timed out
```

gabbifish:
This happens because of inconsistent error messages; I think reqwest prepends `Error:` as a prefix when it fails due to things like Timeouts. The KV API itself, however, does not prepend `Error` as a prefix. Wrangler always prepends error messages with `Error: `. This means reqwest errors have Error printed out twice.

 Focusing on bigger fish to fry at the moment, but this is an excellent candidate for a fix if anyone has spare cycles :)

PR #762, opened 2019-10-04
Fix redundant "Error: " specified in #668.

vypr:
I believe this is the only place where such a bug occurs. Mentoring appreciated. :)


gabbifish:
Thanks for your contribution, @vypr! First things first, let's make sure CI is happy with these changes. It looks like you're missing a `)` in your change to the if statement. :)

vypr:
To think I've been doing this as a hobby for 12 years. Thanks for the heads up.

gabbifish:
Hi @vypr! Thank you for your patience! So, upon further inspection, it looks like this duplicate print actually lives in our Rust code, not in the JS in this PR: in #668, the duplicate print occurs when calling to an API endpoint that is only called by our rust code. If you'd like, you can take this opportunity to fix this problem within our Rust code, and I'd be happy to continue providing feedback! This may be a good place to start: https://github.com/cloudflare/wrangler/blob/master/src/commands/kv/mod.rs#L78-L99

vypr:
I'll take a look, thanks again!

gabbifish:
Hi @vypr! Just checking in--have you had a chance to further explore this issue? No worries if not! :D

I'm also planning to close this PR for now, given that it touches javascript code ultimately not related to the issue at hand.

Do let us know if we can help you!

PR #764, opened 2019-10-04
Clean up subdomain creation

### LINKED ISSUE #758: Change signature of Subdomain::get to return an Option
Related to https://github.com/cloudflare/wrangler/pull/747/files

There are a couple of instances in the codebase that still use `get` that fails if there is no subdomain, those should still fail (no behavior changes) but instead of failing in Subdomain::get the failures should be handled wherever it was called

EverlastingBugstopper:
This PR uses the wonderful logic from #747 to check if a subdomain exists before calling the API. Removes the necessity for some logic branches.

Fixes #758


gabbifish:
@EverlastingBugstopper oop! Looks like a small compiler error is breaking most of the builds :) Once this is fixed I'll review!

src/commands/subdomain.rs
```diff
@@ -82,44 +65,24 @@ fn register_subdomain(
...
+    let subdomain = Subdomain {
         subdomain: name.to_string(),
     };
-    let sd_request = serde_json::to_string(&sd)?;
+    let subdomain_request = serde_json::to_string(&subdomain)?;
 
     let client = http::auth_client(None, user);
 
-    let mut res = client.put(&addr).body(sd_request).send()?;
+    let mut res = client.put(&addr).body(subdomain_request).send()?;
```
gabbifish: This call to subdomain API looks redundant--it seems like we make the exact same call later in Subdomain::get(). Is there any way we can consolidate this into just one call?

EverlastingBugstopper: There doesn't really seem to be a way around this - we already make two requests to the API in the existing implementation. In fact I think we make three and this PR reduces it to two. The `put` request doesn't seem to give us info about _why_ the put failed, so we must perform a `get` as well to determine if they already have a subdomain or not.

gabbifish: Ahh! I see. I misread and thought both were GETs. This makes sense now.

gabbifish: a last tiny nit... this would be awesome to have as a Subdomain::put method! What do you think?

PR #765, opened 2019-10-04
Always npm install

### LINKED ISSUE #428: Suggest re-building node_modules on failed builds where appropriate
We assume that the existence of `node_modules` in a `webpack` project means the dependencies are installed and we don't need to run npm install again. However, if the user does ctrl-c during `wrangler build`, `wrangler publish` or `wrangler preview` while the npm install stage is running, it is possible to have a `node_modules` folder with invalid contents.

ashleymichal:
Closes #428 by removing logic to skip npm install step if node_modules is already present. npm install is pretty good about not redundantly downloading packages. tested locally on a mac.


ashleygwilliams:
closing in favor of needing more design for how we can let users "customize builds"

PR #789, opened 2019-10-21
Disallow templates with generate site

### LINKED ISSUE #776: Warn for redundant site generation: generate via git +  --site behavior
After the sync on (10/8) we decided we should throw a warning/error on the behavior of using wrangler generate with a site template alone side the `--site` flag

For example 

```
wrangler generate name  https://github.com/cloudflare/workers-site-template --site
```

EverlastingBugstopper:
This PR adds failure logic when a template is passed to `wrangler generate --site`

**Example**
```console
$ wrangler generate --site name https://github.com/cloudflare/wrangler
Error: You cannot specify a template when generating a Workers Site. If you want to generate site boilerplate, run wrangler generate --site
```

Fixes #776

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -390,12 +390,12 @@ fn run() -> Result<(), failure::Error> {
     } else if let Some(matches) = matches.subcommand_matches("generate") {
         let name = matches.value_of("name").unwrap_or("worker");
         let site = matches.is_present("site");
-        let template_value = matches.value_of("template");
+        let template = matches.value_of("template");
         let mut target_type = None;
 
         let template = if site {
-            if template_value.is_some() {
-                failure::bail!("You cannot specify a template when generating a Workers Site. If you want to generate site boilerplate, run wrangler generate --site")
+            if template.is_some() {
+                failure::bail!("You cannot specify a template when generating a Workers Site. If you'd like to use the default site boilerplate, run wrangler generate --site. If you'd like to use another site boilerplate, omit the --site when running wrangler generate.")
             }
             "https://github.com/cloudflare/worker-sites-template"
         } else {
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -390,17 +390,21 @@ fn run() -> Result<(), failure::Error> {
     } else if let Some(matches) = matches.subcommand_matches("generate") {
         let name = matches.value_of("name").unwrap_or("worker");
         let site = matches.is_present("site");
+        let template_value = matches.value_of("template");
         let mut target_type = None;
 
         let template = if site {
+            if template_value.is_some() {
+                failure::bail!("You cannot specify a template when generating a Workers Site. If you want to generate site boilerplate, run wrangler generate --site")
```
ashleygwilliams: This text is a little confusing to me- I might say:

"You cannot specify a template when generating a Workers Site. If you'd like to use the default site boilerplate, run `wrangler generate --site`. If you'd like to use another site boilerplate, omit the `--site` when running `wrangler generate`."

src/main.rs
```diff
@@ -390,17 +390,21 @@ fn run() -> Result<(), failure::Error> {
     } else if let Some(matches) = matches.subcommand_matches("generate") {
         let name = matches.value_of("name").unwrap_or("worker");
         let site = matches.is_present("site");
+        let template_value = matches.value_of("template");
```
ashleygwilliams: template value is a confusing name to me- maybe template_url?

src/main.rs
```diff
@@ -390,17 +390,21 @@ fn run() -> Result<(), failure::Error> {
     } else if let Some(matches) = matches.subcommand_matches("generate") {
         let name = matches.value_of("name").unwrap_or("worker");
         let site = matches.is_present("site");
+        let template = matches.value_of("template");
         let mut target_type = None;
 
         let template = if site {
+            if template.is_some() {
+                failure::bail!("You cannot specify a template when generating a Workers Site. If you'd like to use the default site boilerplate, run wrangler generate --site. If you'd like to use another site boilerplate, omit the --site when running wrangler generate.")
```
ashleygwilliams: lol i'm still on this, something seems off here

"**You cannot specify a template when generating a Workers Site.** If you'd like to use the default site boilerplate, run wrangler generate --site. If you'd like to use another site boilerplate, omit the --site when running wrangler generate."

I think if we change the bolded part to "You've passed both `--site` and a template to `wrangler generate`- however you can only pass one or the other." 

what do you think?

PR #790, opened 2019-10-21
Verbose file printouts when using wrangler publish with a Worker Site

### LINKED ISSUE #657: Add --verbose option for `wrangler publish`
Given that `wrangler publish` may upload a lot of files when publishing a Workers Site, we can offer it a `--verbose` option similar to #626 that outputs all the files uploaded as part of the `sync()` call made by `publish`.

gabbifish:
This PR closes #657. It allows Wrangler users running `wrangler publish` or `wrangler preview` to see which files are being uploaded to Workers KV:

```console
$ wrangler publish --version
 Using namespace for Workers Site "__blog-workers_sites_assets"
 Preparing to upload updated files...
 Processing ./public/sitemap.xml...
 Processing ./public/page/1/index.html...
 Processing ./public/posts/hahaha.butts/index.html...
 Processing ./public/posts/index.xml...
 ...
 Success
 ⬇️ Installing wasm-pack...
 Built successfully, built project size is 11 KiB.
 Successfully published your script to https://blog.ziggy.workers.dev
```


src/commands/kv/bucket/mod.rs
```diff
@@ -36,6 +36,10 @@ pub fn directory_keys_values(
         let entry = entry.unwrap();
         let path = entry.path();
         if path.is_file() {
+            if print_files {
+                message::working(&format!("Parsing {}...", path.display()));
```
xtuc: Why parsing by the way?

gabbifish: because we technically don't upload until later :P by then, all the files have been batched into vecs for batched KV bulk uploads, so it's not optimal to print out the files there (we'd be iterating over the files again when we can just print out files in this initial interation).

src/main.rs
```diff
@@ -349,6 +349,13 @@ fn run() -> Result<(), failure::Error> {
                         .short("e")
                         .long("env")
                         .takes_value(true)
+                )
+                .arg(
+                    Arg::with_name("print-files")
+                        .short("p")
```
xtuc: I think only commonly used flags should have a short one, `--p` could be something else in the future.

src/commands/kv/bucket/mod.rs
xtuc: I see. I would be fine with `Processing x ...`. `Parsing` the file sounds odd to me, especially if we don't do it at all.

gabbifish: That sounds good! Will replace `Parsing` with `Processing` :)

src/main.rs
EverlastingBugstopper: Is there any reason we are not calling this `--verbose`? `--print-files` seems a bit funky to me. Not a gigantic dealio to me though

src/main.rs
```diff
@@ -475,7 +482,7 @@ fn run() -> Result<(), failure::Error> {
         let env = matches.value_of("env");
         let mut target = manifest.get_target(env)?;
 
-        commands::publish(&user, &mut target)?;
+        commands::publish(&user, &mut target, matches.is_present("print-files"))?;
```
EverlastingBugstopper: nit: Could we separate this `matches.is_present` to a separate line?

src/commands/kv/bucket/mod.rs
```diff
@@ -36,6 +36,10 @@ pub fn directory_keys_values(
         let entry = entry.unwrap();
         let path = entry.path();
         if path.is_file() {
+            if print_files {
+                message::working(&format!("Processing {}...", path.display()));
```
EverlastingBugstopper: Should "Processing" be "Uploading"? We aren't doing any processing on each file.

EverlastingBugstopper: On second thought (personal opinion incoming): I think I'd like to just have each file name on its own line without `Processing` or the `...`.

gabbifish: Heh Sven suggested switching to Processing instead of Parsing! I think either is okay... I have no strong opinions on which verb we use. I can get rid of the ellipses easy.

src/main.rs
gabbifish: i already cargo fmtted this so this should be fine?

gabbifish:
@EverlastingBugstopper what did you think about extending this behavior to `wrangler preview` as well?

src/commands/kv/bucket/mod.rs
EverlastingBugstopper: Yeah - I feel like it might be redundant to have any verb (not hugely important) but seems like we could get away with each line just being the file name by itself with no frills (maybe even no emoji for easy copy pasting?)

EverlastingBugstopper:
I think extending to preview is a great idea - it's probably going to be used more often when debugging in preview than when publishing to be honest. I'm still a bit confused what you mean by "true verbose"

gabbifish:
True verbose refers to actually printing out logs (like our info
logs!)—it’s important that we don’t conflate that and the simple logic here
for printing out file names :)

ashleygwilliams:
@gabbifish is there a reason that we can't just make this what `verbose` would mean for this command? i.e. these are INFO logs? i'm not sure the bepsoke flag is necessary here

i would expected wrangler publish --verbose to print what this does i think

src/main.rs
xtuc: I suggested it because it does only print files. People will probably expect to have a verbose on all commands where this will only be for Worker sites.

gabbifish:
These file printouts are not info logs—instead, they are message::working
printouts. I originally used —verbose but Sven suggested we use something
different so we didn’t conflate these printouts with what we expect from
true verbose (eg. Printing out actual leveled logs). I did like his point,
but am happy to revert to using —verbose if folks prefer it.

ashleygwilliams:
can you summarize @xtuc's point (it's late his time so he may be asleep)?  im' not sure i understand the difference but it seems useful to understand! (i dont see it in the comment history so i'm not sure where the convo happened- would be good to document here!)

xtuc:
I also don't feel strongly about that. My argument was that `--print-files` is self describing, `--verbose` will add some noise in the ouput.

gabbifish:
The file printouts are distinct from info logs—in the code you can see they
are printed out with message::warn. This means that info logs and the file
printouts won’t overlap :D does this make you feel better about us using
—verbose then?

gabbifish:
I'm going to revert to using `--verbose` instead of `--print-files`. Also extending this functionality to preview.

src/main.rs
```diff
@@ -349,6 +356,13 @@ fn run() -> Result<(), failure::Error> {
                         .short("e")
                         .long("env")
                         .takes_value(true)
+                )
+                .arg(
+                    Arg::with_name("verbose")
+                        .short("v")
```
xtuc: I wouldn't use a short alias for this flag. It could be in our way when we want to add a flag that starts with the same name and that is more commonly used.

src/commands/kv/bucket/mod.rs
ashleygwilliams: We aren't parsing the files so we should change this verb.

src/main.rs
ashleygwilliams: agree

src/main.rs
```diff
@@ -460,7 +474,14 @@ fn run() -> Result<(), failure::Error> {
...
         let watch = matches.is_present("watch");
 
-        commands::preview(target, user, method, body, watch)?;
+        commands::preview(
+            target,
+            user,
+            method,
+            body,
+            watch,
+            matches.is_present("verbose"),
```
ashleygwilliams: i think it'd be cleaner to take this matches.is_present out of the functional cal and set it up in a variable, a nit, but we are already passing so much i think it would read better

src/main.rs
```diff
@@ -475,7 +496,7 @@ fn run() -> Result<(), failure::Error> {
         let env = matches.value_of("env");
         let mut target = manifest.get_target(env)?;
 
-        commands::publish(&user, &mut target)?;
+        commands::publish(&user, &mut target, matches.is_present("verbose"))?;
```
ashleygwilliams: same here

src/commands/kv/bucket/mod.rs
gabbifish: How about I use the word "preparing"? I guess we are preparing since we are base64 encoding all file contents :)

src/commands/kv/bucket/mod.rs
gabbifish: How about `Preparing`?

simplenotezy:
Is this is 1.11.0 release yet?

PR #795, opened 2019-10-22
feat(publish): better error msg on needs email verif

### LINKED ISSUE #320: Provide a readable error message for API code 10034 - Please verify your email
```
$ wrangler subdomain jazz
🐌  Registering your subdomain, jazz.workers.dev, this could take up to a minute.
Error: ErrorMessage { msg: "⛔ There was an error creating your requested subdomain.\n Status Code: 400 Bad Request\n Msg: {\"success\":false,\"errors\":[{\"code\":7003,\"message\":\"Could not route to \\/accounts\\/workers\\/subdomain, perhaps your object identifier is invalid?\"},{\"code\":7000,\"message\":\"No route for that URI\"}],\"messages\":[],\"result\":null}" }
```

This should be updated to inform the user that they need to validate their email before they can register the subdomain (maybe give them a link to re-send the email?)

ashleygwilliams:
fixes #320


src/commands/publish/mod.rs
```diff
@@ -118,6 +117,35 @@ fn build_and_publish_script(
     Ok(())
 }
 
+fn error_msg(status: reqwest::StatusCode, text: String) -> String {
+    if text.contains("\"code\": 10034,") {
+        "You need to verify your account's email address before you can publish.".to_string()
```
EverlastingBugstopper: can we add a link to dash.cloudflare.com

ashleygwilliams:
@EverlastingBugstopper is it better now? feels awk to me but i dunno what to update it to

src/commands/publish/mod.rs
```diff
@@ -122,6 +121,33 @@ fn build_and_publish_script(
...
+    {
+      "code": 10034,
+      "message": "workers.api.error.email_verification_required"
+    }
+  ],
+  "messages": []
+}"#
+    .to_string();
+    let result = error_msg(status, text);
+    assert!(result.contains("https://dash.cloudflare.com"));
```
gabbifish: Could we even make the error message above a string const and compare the actual result in this test to this string const?

ashleygwilliams:
@gabbifish i think the idea of actually parsing the error to make it prettier would be a great next step- i'll file an issue!

PR #797, opened 2019-10-23
use cloudflare worker to proxy requests to github and cache distribution binaries

gabbifish:
This PR allows Wrangler installs to access Wrangler binaries via a caching proxy instead of hitting github directly (and possibly getting rate-limited).


xtuc:
Could the worker code live in this repo as well?

src/install/mod.rs
```diff
@@ -151,7 +151,7 @@ fn prebuilt_url(tool_name: &str, owner: &str, version: &str) -> Option<String> {
         };
 
         let url = format!(
-            "https://github.com/{0}/{1}/releases/download/v{2}/{1}-v{2}-{3}.tar.gz",
+            "https://workers.cloudflare.com/get-wrangler-binary/{0}/{1}/v{2}/{3}.tar.gz",
```
xtuc: Are owner + tool_name validated on worker's side? I don't want that people start using our gateway for their own GitHub release

gabbifish:
@xtuc I think storing the worker code within this repo could be potentially awkard; i would suggest we have a monorepo for these worker snippets elsewhere.

src/install/mod.rs
gabbifish: They are validated on worker side.

xtuc:
@gabbifish that's right, maybe just in the cloudflare org.

npm/install-wrangler.js
```diff
@@ -37,7 +37,7 @@ function downloadAsset(version, platform) {
   }
   mkdirSync(dest);
 
-  const url = `https://cloudflare.works/download-wrangler/${ version }/${ platform }`
+  const url = `https://workers.cloudflare.com/get-npm-wrangler-binary/${ version }/${ platform }`
```
EverlastingBugstopper: nit: can we call it `/get-wrangler` instead? i can see a future where we want to download binaries from something other than `npm` (and maybe people wanna use `yarn`) - seems to make sense to just say `get-wrangler` or `get-wrangler-binary`

src/install/mod.rs
```diff
@@ -151,7 +151,7 @@ fn prebuilt_url(tool_name: &str, owner: &str, version: &str) -> Option<String> {
         };
 
         let url = format!(
-            "https://github.com/{0}/{1}/releases/download/v{2}/{1}-v{2}-{3}.tar.gz",
+            "https://workers.cloudflare.com/get-binary/{0}/{1}/v{2}/{3}.tar.gz",
```
EverlastingBugstopper: wait.. what's the difference between this and the `get-npm-wrangler-binary` endpoint?

gabbifish:
This gets the binaries of dependencies (e.g) rust-wasm)

EverlastingBugstopper:
Ah makes sense.

npm/install-wrangler.js
gabbifish: This is used by the npm installer logic in Wrangler, so I still think it's worth keeping the `npm` string.

ashleygwilliams:
@EverlastingBugstopper i'm already on that- think we'll get a pingdom set up

PR #805, opened 2019-10-24
Clean up template config struct

EverlastingBugstopper:
Wanted to break this out separately so it's easier to review and in case there are weird side effects we need to QA. It eliminates the `TemplateConfig` struct and puts some serde defaults in `Manifest`


src/settings/target/manifest.rs
```diff
@@ -92,30 +103,25 @@ impl Manifest {
             None => Some(true),
         };
 
+        template_config.name = name;
```
gabbifish: I'm trying to understand how this logic maps onto the removed logic that creates a new Manifest--mind filling me in?

src/settings/target/manifest.rs
```diff
@@ -16,15 +15,27 @@ use serde::{Deserialize, Serialize};
 use crate::terminal::emoji;
 use crate::terminal::message;
 
+fn default_name() -> String {
```
ashleymichal: curious, why a fn and not a const?

src/settings/target/manifest.rs
EverlastingBugstopper: this will read in the manifest from the template and then override individual fields with what was passed via the command line. so for this - name is taken from `wrangler generate` and overrides whatever name was previously serialized

EverlastingBugstopper: since we're serializing a manifest in the first place i decided to just make it mutable rather than making a new one and cloning everything, should be more performant

gabbifish: Cool!

src/settings/target/manifest.rs
ashleygwilliams: the diff is outdated, but the question is still valuable!

src/settings/target/manifest.rs
```diff
@@ -16,15 +15,27 @@ use serde::{Deserialize, Serialize};
 use crate::terminal::emoji;
 use crate::terminal::message;
 
+fn empty_string() -> String {
```
ashleygwilliams: i'm not convinced these should be functions- but i am open to it. otherwise looks good to me.

EverlastingBugstopper: serde requires them to be functions: 
```
expected function, found enum variant `some_string`
```

PR #816, opened 2019-10-28
Add the ability to preview without opening the browser

### LINKED ISSUE #256: give option to not open browser when previewing
Hi, great talk last Tuesday at cloudflare connect in London on workers and rust.

Wrangler looks awesome, building and testing workers looks like it's got a lot nicer in the last 6 months.

Would be great if I could preview workers without having to open a browser (or have one opened for me) for every preview.

A few things that would make this possible:
* option to not open a browser on `wrangler preview` (obviously)
* connect to the websocket for receiving console output and log it to the terminal - quite how this is displayed is an open question since one can't click to expand output in the terminal as one can in a browser console
* get more details about the request to the worker and response, eg. headers and body - something like the output from httpie.

There's a gist [here](https://gist.github.com/samuelcolvin/bc4783a69ac78e5a09ad7f82ff4b3a92) (old and ugly code, sorry) that demonstrates much of what I'm asking for.

Failing all this, one step would be to reuse the same browser tab for each preview so I don't end up with 50 old cloudflare worker tabs, although maybe that's a limitation of chrome?

EverlastingBugstopper:
Fixes #256 

This PR adds `wrangler preview --headless` which does not open the browser for preview requests. The tests for `wrangler preview` now use this flag.


src/commands/publish/preview/mod.rs
```diff
@@ -34,6 +34,7 @@ pub fn preview(
     body: Option<String>,
     livereload: bool,
     verbose: bool,
+    browser: bool,
```
ashleygwilliams: the flag is "terminal" but this attribute says browser... might be more maintainable if they are aligned- what do you think?

i sorta think that it might be nice to have the flag be browser instead of terminal... backwards compat has us a bit stuck though. gonna think on this- but would love to hear your thoughts!

EverlastingBugstopper: My preference would be that the default opens the browser and that the flag is browser. The reason I implemented it this way is because of the way things are today. That being said, I don't think we really need to worry too much about backwards compatibility wrt preview, but I could be wrong.

ashleymichal: should it be configurable via another method as well? there's the whole idea that if there's a flag for it, there should be a config setting for it.

ashleymichal: also, perhaps it isn't a no arg flag, but a bool value flag, like `--browser=true`, with a default

EverlastingBugstopper:
We've had a bunch of discussion around this that I will attempt to summarize here.

This feature is largely temporary until we deprecate responses being shown in the terminal for `wrangler preview` in favor of a new command `wrangler curl`. 

`--terminal` doesn't quite map properly because if the flag is omitted, the output is opened in the browser, but also printed in the terminal. 

Another proposal to get around this was `--browser=false`, which gets closer to what we want. 

My favorite proposal is `--headless` as it is commonly used to describe this type of functionality, and also doesn't require the user to type `=false`. 

@ashleymichal proposed that we *also* introduce a field in the manifest in addition to an environment variable to configure the default behavior so people can simply `wrangler preview` and it won't open the browser, even if they don't pass a flag. I'm personally not in favor of taking this approach as it will add cruft to people's manifests and their environment variables. We plan on deprecating this feature, and while we can deprecate the feature and the implementation at the same time, instead of simply asking people to type `--headless`, they will need to go through all of their projects and update their manifests. It also doesn't map well to the rest of what we allow to be configured in the manifests as it has nothing to do with deploying, and more to do with the personal preference of the individual developer.

ashleymichal:
at the risk of sounding a bit rude, after all this time of this being a problem, why introduce a stop gap we intend to almost immediately deprecate, when we are actively working on the full-featured command that will replace it and the work is slated to complete this quarter?

ashleygwilliams:
i'm in favor  of `--headless` (make the variable in the code the same, for maintenance sake), eventually deprecate for `wrangler curl`, and don't make it an env var because we will eventually deprecate

if we made it configurable it should be a global config not a per project config, and we dont have that beyond auth yet so i dont want to rush into introducing it

EverlastingBugstopper:
@ashleymichal I think that's a reasonable ask - and is why I'm opposed to introducing manifest changes. I think that this is a small enough change that will alleviate some pain for devs (including our CI running wrangler tests) until we get wrangler curl out.

ashleymichal:
> I also think we'll want to keep --headless around for our own CI when we still want to run tests on preview without opening the browser.

will we maintain long term support for this feature?

EverlastingBugstopper:
> does this imply that will we maintain long term support for this feature?

I think so

PR #826, opened 2019-10-31
Alewis/refactor uploads

ashleymichal:
Some refactors to make testing upload form building easier.

* Use PathBuf instead of String when generating upload forms
* remove call to build command from upload form building logic

### DIFF
--- a/src/commands/publish/mod.rs
+++ b/src/commands/publish/mod.rs
@@ -9,8 +9,6 @@ pub use package::Package;
 use crate::settings::target::KvNamespace;
 use route::Route;
 
-use upload_form::build_upload_form;
-
 use std::path::Path;
 
 use crate::commands;
@@ -93,7 +91,7 @@ fn publish_script(
         http::auth_client(None, user)
     };
 
-    let script_upload_form = build_upload_form(target, asset_manifest)?;
+    let script_upload_form = upload_form::build(target, asset_manifest)?;
 
     let mut res = client
         .put(&worker_addr)
--- a/src/commands/publish/mod.rs
+++ b/src/commands/publish/mod.rs
@@ -9,10 +9,11 @@ pub use package::Package;
 use crate::settings::target::KvNamespace;
 use route::Route;
 
-use upload_form::build_script_and_upload_form;
+use upload_form::build_upload_form;
 
 use std::path::Path;
 
+use crate::commands;
 use crate::commands::kv;
 use crate::commands::kv::bucket::AssetManifest;
 use crate::commands::subdomain::Subdomain;
@@ -43,7 +44,11 @@ pub fn publish(
     }
 
     let asset_manifest = upload_buckets(target, user, verbose)?;
-    build_and_publish_script(&user, &target, asset_manifest)?;
+
+    // Build the script before uploading.
+    commands::build(&target)?;
+
+    publish_script(&user, &target, asset_manifest)?;
 
     Ok(())
 }
@@ -72,7 +77,7 @@ pub fn bind_static_site_contents(
     Ok(())
 }
 
-fn build_and_publish_script(
+fn publish_script(
     user: &GlobalUser,
     target: &Target,
     asset_manifest: Option<AssetManifest>,
@@ -88,7 +93,7 @@ fn build_and_publish_script(
         http::auth_client(None, user)
     };
 
-    let script_upload_form = build_script_and_upload_form(target, asset_manifest)?;
+    let script_upload_form = build_upload_form(target, asset_manifest)?;
 
     let mut res = client
         .put(&worker_addr)
--- a/src/commands/publish/package.rs
+++ b/src/commands/publish/package.rs
@@ -16,8 +16,8 @@ impl Package {
             )
         } else if !build_dir.join(&self.main).exists() {
             failure::bail!(
-                "The entrypoint of your Worker ({:?}) could not be found.",
-                self.main
+                "The entrypoint of your Worker ({}) could not be found.",
+                self.main.display()
             )
         } else {
             Ok(self.main.clone())
@@ -30,15 +30,15 @@ impl Package {
         let manifest_path = pkg_path.join("package.json");
         if !manifest_path.is_file() {
             failure::bail!(
-                "Your JavaScript project is missing a `package.json` file; is `{:?}` the \
+                "Your JavaScript project is missing a `package.json` file; is `{}` the \
                  wrong directory?",
-                pkg_path
+                pkg_path.display()
             )
         }
 
         let package_json: String = fs::read_to_string(manifest_path.clone())?.parse()?;
         let package: Package = serde_json::from_str(&package_json)
-            .unwrap_or_else(|_| panic!("could not parse {:?}", manifest_path));
+            .unwrap_or_else(|_| panic!("could not parse {}", manifest_path.display()));
 
         Ok(package)
     }
--- a/src/commands/publish/package.rs
+++ b/src/commands/publish/package.rs
@@ -12,7 +12,7 @@ impl Package {
     pub fn main(&self, build_dir: &PathBuf) -> Result<PathBuf, failure::Error> {
         if self.main == PathBuf::from("") {
             failure::bail!(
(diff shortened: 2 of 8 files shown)


src/commands/publish/package.rs
```diff
@@ -6,17 +6,17 @@ use serde::{self, Deserialize};
...
-        if self.main == "" {
+    pub fn main(&self, build_dir: &PathBuf) -> Result<PathBuf, failure::Error> {
+        if self.main == PathBuf::from("") {
             failure::bail!(
                 "The `main` key in your `package.json` file is required; please specified the entrypoint of your Worker.",
             )
         } else if !build_dir.join(&self.main).exists() {
             failure::bail!(
-                "The entrypoint of your Worker ({}) could not be found.",
+                "The entrypoint of your Worker ({:?}) could not be found.",
```
gabbifish: We could use `{}` as long as we call `self.main.display()` in the second arg to `failure::bail!`

src/commands/publish/upload_form/project_assets.rs
```diff
@@ -58,7 +60,7 @@ impl ProjectAssets {
         self.script_name.to_string()
     }
 
-    pub fn script_path(&self) -> String {
-        self.script_path.to_string()
+    pub fn script_path(&self) -> PathBuf {
+        self.script_path.clone()
```
gabbifish: do we even need this function anymore, or could we start using `self.script_path` directly and borrow it when we pass it into functions?

src/commands/publish/mod.rs
```diff
@@ -9,10 +9,11 @@ pub use package::Package;
 use crate::settings::target::KvNamespace;
 use route::Route;
 
-use upload_form::build_script_and_upload_form;
+use upload_form::build_upload_form;
```
ashleymichal: this stutter is a bummer.

src/commands/publish/upload_form/wasm_module.rs
```diff
@@ -31,8 +33,8 @@ impl WasmModule {
         Binding::new_wasm_module(name, part)
     }
 
-    pub fn path(&self) -> String {
-        self.path.to_string()
+    pub fn path(&self) -> PathBuf {
+        self.path.clone()
```
gabbifish: same commenet as above for `script_path` :)

src/commands/publish/upload_form/wasm_module.rs
```diff
@@ -1,19 +1,21 @@
...
     filename: String,
     binding: String,
 }
 
 impl WasmModule {
-    pub fn new(path: String, binding: String) -> Result<Self, failure::Error> {
+    pub fn new(path: PathBuf, binding: String) -> Result<Self, failure::Error> {
         let filename = filename_from_path(&path)
-            .ok_or_else(|| format_err!("filename should not be empty: {}", path))?;
+            .ok_or_else(|| format_err!("filename should not be empty: {:?}", path))?;
```
gabbifish: we can use `{}` with `path.display()`

src/commands/publish/mod.rs
gabbifish: hardly the end of the world though!

src/commands/publish/upload_form/project_assets.rs
```diff
@@ -58,7 +60,7 @@ impl ProjectAssets {
         self.script_name.to_string()
     }
 
-    pub fn script_path(&self) -> String {
-        self.script_path.to_string()
+    pub fn script_path(&self) -> PathBuf {
```
ashleymichal: if this is not expected to change, should we return a reference to the attribute itself?

src/commands/publish/upload_form/project_assets.rs
```diff
@@ -58,7 +60,7 @@ impl ProjectAssets {
         self.script_name.to_string()
```
ashleymichal: same question here

src/commands/publish/upload_form/wasm_module.rs
ashleymichal: iirc this sucks to read on windows, suggestions welcome

src/commands/publish/upload_form/wasm_module.rs
ashleymichal: same question re: ref to attribute

src/commands/publish/upload_form/project_assets.rs
gabbifish: yes!!! i think we are on the same page :)

src/commands/publish/upload_form/wasm_module.rs
gabbifish: ^ maybe my `{}` and `path.display()` suggestion above will address this?

src/commands/publish/mod.rs
ashleymichal: yah but the fix was easy so i added another commit

src/commands/publish/package.rs
ashleymichal: TIL!

src/commands/publish/upload_form/project_assets.rs
ashleymichal: yeah is interesting question; in other languages I have been in the habit of defining accessor methods for attributes, but I still feel shaky on the mutability constraints of Rust structs when they are made public.

src/commands/publish/package.rs
ashleymichal: missed this one, force pushing then will merge.

ashleymichal: missed this one. fixing, force-pushing, merging.

truly madly deeply

src/commands/publish/upload_form/wasm_module.rs
EverlastingBugstopper: it should ^ !!

src/commands/publish/package.rs
```diff
@@ -6,18 +6,18 @@ use serde::{self, Deserialize};
...
-    main: String,
+    main: PathBuf,
 }
 impl Package {
-    pub fn main(&self, build_dir: &PathBuf) -> Result<String, failure::Error> {
-        if self.main == "" {
+    pub fn main(&self, build_dir: &PathBuf) -> Result<PathBuf, failure::Error> {
+        if self.main == PathBuf::from("") {
             failure::bail!(
                 "The `main` key in your `package.json` file is required; please specified the entrypoint of your Worker.",
```
EverlastingBugstopper: nit: not in this PR but since we're here - can we change `specified` to `specify`

src/commands/publish/package.rs
```diff
@@ -30,15 +30,15 @@ impl Package {
         let manifest_path = pkg_path.join("package.json");
         if !manifest_path.is_file() {
             failure::bail!(
-                "Your JavaScript project is missing a `package.json` file; is `{:?}` the \
+                "Your JavaScript project is missing a `package.json` file; is `{}` the \
```
EverlastingBugstopper: i'm a bit confused by this error message (again not introduced in this PR) but maybe we could change it to - "make sure you are running wrangler in the right directory") - then again this probably never gets hit because if they're not in the right directory it probably yelled at them about a wrangler.toml

src/commands/publish/upload_form/project_assets.rs
```diff
@@ -10,21 +12,21 @@ use crate::settings::target::KvNamespace;
...
     pub fn new(
-        script_path: String,
+        script_path: PathBuf,
         wasm_modules: Vec<WasmModule>,
         kv_namespaces: Vec<KvNamespace>,
         text_blobs: Vec<TextBlob>,
     ) -> Result<Self, failure::Error> {
         let script_name = filename_from_path(&script_path)
-            .ok_or_else(|| format_err!("filename should not be empty: {}", script_path))?;
+            .ok_or_else(|| format_err!("filename should not be empty: {:?}", script_path))?;
```
EverlastingBugstopper: can we do a `.display` here as well?

src/commands/publish/package.rs
ashleymichal: 🤷‍♀

PR #830, opened 2019-10-31
Fix default value for wrangler generate template

EverlastingBugstopper:


gabbifish:
hmmm... we could also fix the capitalization in these sentences if we wanted to? just a nit.

PR #831, opened 2019-10-31
Put each subcommand in its own directory

EverlastingBugstopper:
This PR puts each subcommand in its own directory - this is the first step to refactoring main.rs


ashleymichal:
> i looove this

came here to say this

Issue #818, opened 2019-10-28
Remove warnings for dead code in cargo test


EverlastingBugstopper:
```console
$ cargo test           
warning: constant item is never used: `BUNDLE_OUT`
  --> tests/utils/mod.rs:14:1
   |
14 | const BUNDLE_OUT: &str = "./worker";
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: function is never used: `fixture_out_path`
  --> tests/utils/mod.rs:37:1
   |
37 | pub fn fixture_out_path(fixture: &str) -> PathBuf {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: function is never used: `webpack_config`
  --> tests/utils/mod.rs:58:1
   |
58 | pub fn webpack_config(fixture: &str, config: &str) {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This code _is_ used which makes this very confusing. [This issue](https://github.com/rust-lang/rust/issues/46379) seems related but I can't make heads nor tails of it.

pradovic:
Hi! 
It looks like the issue is related indeed. However, what is missing in the issue responses is that if your functions are public, this can be used by `pub mod` declaration, as unused warning should be suppressed for public mod/fn. I have opened a PR with the suggested approach: https://github.com/cloudflare/wrangler/pull/832. 
Let me know does it makes sense :) Cheers!

PR #832, opened 2019-11-01
Fix unused code warnings in tests

pradovic:
Hi!
fixes https://github.com/cloudflare/wrangler/issues/818.
If I am correct, the reason for an unused warnings were because the module was not made public. There is a nice example about this behavior here: https://doc.rust-lang.org/1.29.2/book/2018-edition/ch07-02-controlling-visibility-with-pub.html.
Making a module public only in `preview.rs` would be enough, as `build.rs` actually uses all the functions. Basically, each file in `/tests/` is build separately. If the `utils` mod is used in either file, in order to avoid unused code warnings, the file (integration test) must either:
1. Use all functions from the `utils` module or
2. Declare `pub mod utils` because unused warnings are suppressed for pub modules and functions. This approach makes sense imo, because there is no reason to warn about not using some of the public functions of the public, shared module. 

On the good side, it appears that it does not need to be used in every test file.
Let me know what you think :)


pradovic:
@ashleymichal No problem, a pleasure! Thanks!

PR #835, opened 2019-11-02
Introduce enum for GlobalUser to handle distinct authentication methods (token vs global api key)

gabbifish:
This PR introduces a refactor to more cleanly implement API token handling. Instead of inferring authentication method by the presence of fields in a GlobalUser struct, we instead use GlobalUser as an enum and define two explicit types for different authentication methods: KeyAuthUser and TokenAuthUser.


src/http.rs
```diff
@@ -48,24 +48,16 @@ pub fn auth_client(feature: Option<&str>, user: &GlobalUser) -> Client {
...
-                    headers.insert("X-Auth-Email", HeaderValue::from_str(&email).unwrap())
-                }
-                None => None,
-            };
-            match &user.api_key {
-                Some(key) => headers.insert("X-Auth-Key", HeaderValue::from_str(&key).unwrap()),
-                None => None,
-            };
-            None
+    match user {
```
ashleymichal: this is tasty.

src/main.rs
```diff
@@ -401,36 +401,31 @@ fn run() -> Result<(), failure::Error> {
...
-        let mut user = GlobalUser {
-            email: None,
-            api_key: None,
-            api_token: None,
-        };
+        let mut user: GlobalUser;
 
         if !api_key {
             // Default: use API token.
             message::info("Looking to use a Global API Key and email instead? Run \"wrangler config --api-key\". (Not Recommended)");
```
ashleymichal: this message still feels kinda weird to me; we want people to use API token, outputting a message that suggests another method, but then tagging on the caveat of "not recommended", just feels confusing.

src/main.rs
```diff
@@ -401,36 +401,31 @@ fn run() -> Result<(), failure::Error> {
...
         let api_key = matches.is_present("api-key");
 
-        let mut user = GlobalUser {
-            email: None,
-            api_key: None,
-            api_token: None,
-        };
+        let mut user: GlobalUser;
 
         if !api_key {
```
ashleymichal: nit: if not is weird to read, i know it's the "preferred" way but i think flipping the if/else cases is more readable, and also follows the pattern of handling "special" or "weird" or "exit" cases first; "early return".

src/settings/global_user.rs
```diff
@@ -8,11 +8,11 @@ use serde::{Deserialize, Serialize};
...
-#[derive(Clone, Debug, Deserialize, Serialize)]
-pub struct GlobalUser {
-    pub email: Option<String>,
-    pub api_key: Option<String>,
-    pub api_token: Option<String>,
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+#[serde(untagged)]
+pub enum GlobalUser {
+    TokenAuthUser { api_token: String },
+    KeyAuthUser { email: String, api_key: String },
```
EverlastingBugstopper: nit here: can we make this `TokenAuth` and `GlobalKeyAuth` - I feel like `User` is redundant here, and also `Token` and `Key` are very similar

src/main.rs
```diff
@@ -401,36 +401,31 @@ fn run() -> Result<(), failure::Error> {
...
             // Default: use API token.
             message::info("Looking to use a Global API Key and email instead? Run \"wrangler config --api-key\". (Not Recommended)");
             println!("Enter API token: ");
             let mut api_token_str: String = read!("{}\n");
             api_token_str.truncate(api_token_str.trim_end().len());
-            if !api_token_str.is_empty() {
-                user.api_token = Some(api_token_str);
-            }
+            user = GlobalUser::TokenAuthUser {
+                api_token: api_token_str,
```
EverlastingBugstopper: nit: could change `api_token_str` to `api_token` and then you wouldn't need `api_token: api_token_str` you could just do `api_token`

src/main.rs
```diff
@@ -401,36 +401,31 @@ fn run() -> Result<(), failure::Error> {
...
             println!("Enter global API key: ");
             let mut api_key_str: String = read!("{}\n");
             api_key_str.truncate(api_key_str.trim_end().len());
-            if !api_key_str.is_empty() {
-                user.api_key = Some(api_key_str);
-            }
+
+            user = GlobalUser::KeyAuthUser {
+                email: email_str,
+                api_key: api_key_str,
```
EverlastingBugstopper: same nit regarding `_str`

src/main.rs
ashleymichal: i actually don't see any existing `api_token` variable in scope, so the qualifier seems unnecessary

src/main.rs
ashleymichal: i'd even go so far as to name the thing that is `api_key` now to `api_key_flag` or `use_api_key`, something that indicates that it is a boolean flag input

src/settings/global_user.rs
ashleymichal: +1

src/main.rs
gabbifish: how about i actually get the opposite of `api_key` when I pull it from the match, so this if-else doesn't have a `!`?

EverlastingBugstopper:
Do we think it's OK to lump this into the API token feature section of the changelog without a specific callout, or should I put it in the maintenance section? @gabbifish

PR #837, opened 2019-11-04
Add help suggestion on webpack failure

### LINKED ISSUE #428: Suggest re-building node_modules on failed builds where appropriate

ashleymichal:
Fixes #428 by including a suggestion to re-run npm install if webpack exits with an error.

### DIFF
--- a/src/commands/build/wranglerjs/mod.rs
+++ b/src/commands/build/wranglerjs/mod.rs
@@ -122,7 +122,9 @@ fn write_wranglerjs_output(
 ) -> Result<(), failure::Error> {
     if output.has_errors() {
         message::user_error(output.get_errors().as_str());
-        failure::bail!("Webpack returned an error");
+        failure::bail!(
+            "webpack returned an error. You may be able to resolve this issue by running npm install."
+        );
     }
 
     bundle.write(output)?;


src/commands/build/wranglerjs/mod.rs
```diff
@@ -122,7 +122,9 @@ fn write_wranglerjs_output(
 ) -> Result<(), failure::Error> {
     if output.has_errors() {
         message::user_error(output.get_errors().as_str());
-        failure::bail!("Webpack returned an error");
+        failure::bail!(
+            "Webpack returned an error. Try re-running npm install, or yarn check --verify-tree."
```
EverlastingBugstopper: will running `yarn check --verify-tree` create a `yarn.lock` file? i'd hesitate to recommend using `yarn` when wrangler uses `npm`

EverlastingBugstopper: I think I'd just add `"You may be able to resolve this issue by running `npm install`"`

ashleymichal: yah wasn't sure bout the yarn thing, snagged it from the original issue discussion to see how people reacted. i think that's a good text.

ashleymichal:
```suggestion
            "Webpack returned an error. You may be able to resolve this issue by running npm install."
```

EverlastingBugstopper:
webpack builds can fail for a lot of reasons. the biggest one is probably not the node_modules issue, but using a package that is incompatible with `target: webworker`. i wonder if it would be more useful to catch a webpack failure the first time, run npm install, and try rebuilding. If that doesn't work, a better error message would be "this is probably not an issue with wrangler, check to make sure any dependencies are compatible with `target: webworker` or something like that. 

i can open an issue for this if we think it's a good idea cc @ashleymichal

PR #839, opened 2019-11-04
Adds error message for site with non-trailing asterisk

### LINKED ISSUE #814: Warn on route without trailing `*` for Workers Sites
`You probably want to change your route from example.com to example.com/*`

EverlastingBugstopper:
Fixes #814 

This PR adds an error message for users deploying a site to a route without a trailing asterisk.

```console
$ wrangler publish
Error: The route in your wrangler.toml should have a trailing * to apply the Worker on every path.
route = test.averyharnish.com*
```

Open to suggestions on the error message itself, and if we think this should be a `message::warn` instead of a `failure::bail`. The reason I have it bailing here is because I see absolutely no scenario where you would deploy a site without a trailing asterisk. Even for single page applications I believe you still need the worker to run on every route.

### DIFF
--- a/src/commands/publish/mod.rs
+++ b/src/commands/publish/mod.rs
@@ -38,7 +38,7 @@ pub fn publish(
     if let Some(site_config) = target.site.clone() {
         if let Some(route) = &target.route {
             if !route.ends_with("*") {
-                failure::bail!("The route in your wrangler.toml should have a trailing * to apply the Worker on every path, otherwise your site will not behave as expected.\nroute = {}*", route)
+                message::warn(&format!("The route in your wrangler.toml should have a trailing * to apply the Worker on every path, otherwise your site will not behave as expected.\nroute = {}*", route));
             }
         }
         bind_static_site_contents(user, target, &site_config, false)?;
--- a/src/commands/publish/mod.rs
+++ b/src/commands/publish/mod.rs
@@ -36,6 +36,11 @@ pub fn publish(
     validate_worker_name(&target.name)?;
 
     if let Some(site_config) = target.site.clone() {
+        if let Some(route) = &target.route {
+            if !route.ends_with("*") {
+                failure::bail!("The route in your wrangler.toml should have a trailing * to apply the Worker on every path, otherwise your site will not behave as expected.\nroute = {}*", route)
+            }
+        }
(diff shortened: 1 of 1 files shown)


src/commands/publish/mod.rs
```diff
@@ -37,6 +37,11 @@ pub fn publish(
     validate_worker_name(&target.name)?;
 
     if let Some(site_config) = target.site.clone() {
+        if let Some(route) = &target.route {
+            if !route.ends_with("*") {
+                failure::bail!("The route in your wrangler.toml should have a trailing * to apply the Worker on every path.\nroute = {}*", route)
```
gabbifish: I'd use `message::warn` and have this message clarify that without the trailing `*`, your site may not work as expected.

ashleymichal: i agree, i don't think bailing is necessarily the right move.

EverlastingBugstopper: Yeah - if you see my original comment on the PR, I wasn't sure if we should bail or warn here. The reason I opted for bailing is because I'm 99% sure that it just breaks your site no matter what, and if you do this the first time you're migrating your site from something else, your site will go down, even if you see the warning. I'm willing to change it to a warning, but I'm really not sure we want to go that "route" 😛

ashleymichal: #iseewhatyoudidthere

my kneejerk reaction is always to say "listen, we don't necessarily know what a user plans on doing with this". However in this case, you make a pretty strong point that a worker that does not use a wild card route here will likely never serve anything other than the root `index.html`, which would be kind of useless.

either way though, i think gabbi's warning of "might not work as expected" is important.

EverlastingBugstopper: even if they were only serving `index.html` and nothing else (which is highly highly unlikely), it would still be in their best interest to have `/*` - if they have other workers that are more specific elsewhere, they will take precedence. I really really don't think we're going to have users come complaining that they can't deploy a worker without the asterisk aside from the usual complaint that its confusing why its necessary at all, which we aren't solving for currently.

ashleymichal: you've changed my mind. i do like the wording "might not work as expected".

EverlastingBugstopper: @ashleygwilliams I'd love to get your input on this

Issue #439, opened 2019-08-14
wrangler config should test for valid credentials


EverlastingBugstopper:
#### Environment

* operating system:
* output of `rustc -V`: `rustc 1.35.0 (3c235d560 2019-05-20)`
* output of `node -v`: `v12.4.0`
* output of `wrangler -V`: `👷 ✨   wrangler 1.1.1`

#### Steps to reproduce

Leave email + api key fields blank when running `wrangler config`

```bash
❯ wrangler whoami
💁‍  👋  You are logged with the email 'me@cloudflare.com'.
❯ wrangler config
Enter email: 

Enter api key: 

✨   Successfully configured. You can find your configuration file at: /Users/me/.wrangler/config/default.toml
❯ wrangler whoami
💁‍  👋  You are logged with the email ''.
❯ rm /Users/me/.wrangler/config/default.toml 
❯ wrangler whoami
Error: ⚠️  Your global config has an error, run `wrangler config`: missing field `email`
```

#### What did you expect to see?

I either expect to see `wrangler config` return an error when I pass it invalid credentials, and for `wrangler whoami` to error when there are invalid credentials in their `.wrangler/config/default.toml`

#### What did you see instead?

No errors and `wrangler whoami` saying that I am successfully logged in as `email ''`.

#### Other cases

It's not just empty credentials that can be successfully configured, completely fake ones work just fine as well.

```bash
❯ wrangler config
Enter email: 
a@cloudflare.com
Enter api key: 
1234
✨   Successfully configured. You can find your configuration file at: /Users/me/.wrangler/config/default.toml
❯ wrangler whoami
💁‍  👋  You are logged with the email 'a@cloudflare.com'.
```

ashleymichal:
one of the bigger issues here is that some of the error messages that wrangler gives for bad credentials are garbage. we should at least improve those. forcing valid credentials may be out of scope for this release (1.3.0).

gabbifish:
If /user is not ready for use with API tokens, we can use the /user/tokens/verify endpoint as mentioned in https://blog.cloudflare.com/api-tokens-general-availability

Issue #740, opened 2019-09-27
Refactor kv::format_error to cf::format_error


EverlastingBugstopper:
Any API call we make with cloudflare-rs should be formatted in a single function. Looking at you https://github.com/cloudflare/wrangler/blob/master/src/commands/kv/namespace/site.rs#L42

This is related to #360

gabbifish:
The `kv::format_error` function calls upon the `help()` function to print out helpful details when the API returns an error. Currently, this function looks for KV-specific error codes to figure out which help message to print. So, if we generalize this help function to `cf::format_error`, we need to extract this `help()` function and ideally have it provide helpful messaging for other CF API error codes.

ashleymichal:
we may need to define multiple help functions for different apis, as there will be collisions in codes between them, e.g. 
workers api error codes:
https://api.cloudflare.com/#worker-script-errors
workers kv api error codes:
https://api.cloudflare.com/#workers-kv-namespace-errors

gabbifish:
#841 actually already fixed this: it introduces a `http::format_error` function that takes a callback as an argument. This callback function takes an error code and returns a string providing more information about it. This means we now have a general error formatting function for the CF API's error responses, which any subcommand that calls a particular API (e.g. the routes API) can use to format CF API errors.

PR #841, opened 2019-11-04
Refactor cloudflare-rs cli to be called from http.rs

gabbifish:
This is a minor refactor to where we call cloudflare-rs logic and methods for handling errors from this library. We move this logic from commands/kv/mod.rs to http.rs so that we can access the cloudflare-rs library beyond the KV module. This is crucial for #439, given that we want to use shared methods between `wrangler kv*` and `wrangler config` to validate that a user's credentials work. Namely, we want to be able to use the same error handling logic of which KV was once the sole user.


src/http.rs
```diff
@@ -5,6 +5,17 @@ use std::time::Duration;
...
+
+use cloudflare::framework::auth::Credentials;
+use cloudflare::framework::response::ApiFailure;
+use cloudflare::framework::{Environment, HttpApiClient, HttpApiClientConfig};
+
+use crate::terminal::emoji;
+use crate::terminal::message;
+
+////---------------------------OLD API CLIENT CODE---------------------------////
+// todo: remove this and replace it entirely with cloudflare-rs
```
EverlastingBugstopper: nit: todo --> TODO and link https://github.com/cloudflare/wrangler/issues/63

src/http.rs
```diff
@@ -61,3 +72,87 @@ fn add_auth_headers<'a>(headers: &'a mut HeaderMap, user: &GlobalUser) {
...
+            message::warn("Returned status code 413, Payload Too Large. Please make sure your upload is less than 100MB in size"),
+        StatusCode::GATEWAY_TIMEOUT =>
+            message::warn("Returned status code 504, Gateway Timeout. Please try again in a few seconds"),
+        _ => (),
+    }
+}
+
+// kv_help() provides more detailed explanations of Workers KV API error codes.
+// See https://api.cloudflare.com/#workers-kv-namespace-errors for details.
+fn kv_help(error_code: u16) -> &'static str {
```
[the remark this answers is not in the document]
EverlastingBugstopper: I opened this issue a while back that asks that we have all of our API codes handled once we move to cloudflare-rs - if we take the passing function approach, could @gabbifish add a comment on this? https://github.com/cloudflare/wrangler/issues/740

src/http.rs
```diff
@@ -61,3 +72,87 @@ fn add_auth_headers<'a>(headers: &'a mut HeaderMap, user: &GlobalUser) {
...
+        10010 | 10011 | 10012 | 10013 | 10014 | 10018 => {
+            "Run `wrangler kv:namespace list` to see your existing namespaces with IDs"
+        }
+        10009 => "Run `wrangler kv:key list` to see your existing keys", // key errors
+        // TODO: link to more info
+        // limit errors
+        10022 | 10024 | 10030 => "See documentation",
+        // TODO: link to tool for this?
+        // legacy namespace errors
+        10021 | 10035 | 10038 => "Consider moving this namespace",
```
EverlastingBugstopper: I know this isn't part of this PR, but do we know exactly what this means and if we can give them a better error message? This doesn't quite make sense to me

src/http.rs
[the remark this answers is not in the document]
ashleymichal: yeah good call; for posterity's sake i'll clarify that the reason they need to be separate is, unfortunately, the different APIs behind the gateway have overlapping error codes, so we can't reliably match on them everywhere; that is, both Workers and Workers KV APIs have an error code 10000 that mean drastically different things.

src/http.rs
ashleymichal: for reference, the actual error codes for KV are excellent, and tell exactly what the problem is, so we print them to the console first. the messages included here are supplementary to give people actionable suggestions for how to fix their problems, they are _not_ the complete error code.

this is another reason why having separate functions for handling errors per module is useful: the messages that come out of the workers api are rather terse by comparison, so i would want to provide more granular handling there.

src/http.rs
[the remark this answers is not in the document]
ashleymichal: > I figured it would be sort of nice to have distinct error codes explanations for all types of endpoints in one place (readability)

i think at the top level, or in its own directory, per api (i.e. workers, workers kv), is a decent level of consolidation.

src/http.rs
```diff
@@ -61,3 +72,59 @@ fn add_auth_headers<'a>(headers: &'a mut HeaderMap, user: &GlobalUser) {
         }
     }
 }
+
+////---------------------------NEW API CLIENT CODE---------------------------////
+pub fn api_client(
```
ashleymichal: boy do i not want to bikeshed this, but i might suggest lending some specificity for future proofing, like `cf_v4_api_client` or something. non-blocking

src/http.rs
```diff
@@ -61,3 +72,59 @@ fn add_auth_headers<'a>(headers: &'a mut HeaderMap, user: &GlobalUser) {
...
+        Credentials::from(user.to_owned()),
+        config,
+        Environment::Production,
+    )
+}
+
+// Format errors from the cloudflare-rs cli for printing.
+// Optionally takes an argument for providing a function that maps error code numbers to
+// helpful additional information about why someone is getting an error message and how to fix it.
+pub fn format_error(e: ApiFailure, err_helper: Option<&dyn Fn(u16) -> &'static str>) -> String {
```
ashleymichal: v. nice

src/http.rs
gabbifish: ooh, i do like this idea... i defs want us to think about this more as a team. Mind if we make a separate ticket for this?

ashleymichal: yes, though being a soon-to-be ubiquitous public function it should be done sooner than later.

PR #846, opened 2019-11-05
Audit comments

EverlastingBugstopper:
This PR standardizes some comments and fixes up some typos

### DIFF
--- a/src/commands/build/wranglerjs/mod.rs
+++ b/src/commands/build/wranglerjs/mod.rs
@@ -32,7 +32,7 @@ use std::time::Duration;
 // Run the underlying {wranglerjs} executable.
 
 // In Rust we create a virtual file, pass it to {wranglerjs}, run the
- // executable and wait for completion. The file will receive a serialized
+// executable and wait for completion. The file will receive a serialized
 // {WranglerjsOutput} struct.
 // Note that the ability to pass a fd is platform-specific
 pub fn run_build(target: &Target) -> Result<(), failure::Error> {
--- a/src/commands/build/wranglerjs/mod.rs
+++ b/src/commands/build/wranglerjs/mod.rs
@@ -30,8 +30,8 @@ use std::thread;
 use std::time::Duration;
 
 // Run the underlying {wranglerjs} executable.
-//
-// In Rust we create a virtual file, pass the pass to {wranglerjs}, run the
+
+// In Rust we create a virtual file, pass it to {wranglerjs}, run the
 // executable and wait for completion. The file will receive the a serialized
 // {WranglerjsOutput} struct.
 // Note that the ability to pass a fd is platform-specific
@@ -62,7 +62,7 @@ pub fn run_build_and_watch(target: &Target, tx: Option<Sender<()>>) -> Result<()
 
     info!("Running {:?} in watch mode", command);
 
-    //Turbofish the result of the closure so we can use ?
+    // Turbofish the result of the closure so we can use ?
     thread::spawn::<_, Result<(), failure::Error>>(move || {
         let _command_guard = util::GuardedCommand::spawn(command);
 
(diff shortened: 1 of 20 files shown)


src/commands/build/wranglerjs/output.rs
```diff
@@ -6,7 +6,7 @@ use serde::Deserialize;
 use std::io::prelude::*;
 
 // This structure represents the communication between {wranglerjs} and
-// {wrangler}. It is send back after {wranglerjs} completion.
+// {wrangler}. It is sent back after {wranglerjs} completion.
 // FIXME(sven): make this private
```
ashleymichal: i guess while we're at it
```suggestion
// TODO: (sven) make this private
```

src/commands/publish/upload_form/mod.rs
```diff
@@ -57,7 +57,7 @@ pub fn build(
         }
         TargetType::Webpack => {
             log::info!("Webpack project detected. Publishing...");
-            // FIXME(sven): shouldn't new
+            // TODO: (sven) shouldn't new
```
EverlastingBugstopper: @xtuc could you explain this comment?

src/commands/build/wranglerjs/mod.rs
```diff
@@ -30,8 +30,8 @@ use std::thread;
 use std::time::Duration;
 
 // Run the underlying {wranglerjs} executable.
-//
-// In Rust we create a virtual file, pass the pass to {wranglerjs}, run the
+
+// In Rust we create a virtual file, pass it to {wranglerjs}, run the
 // executable and wait for completion. The file will receive the a serialized
```
EverlastingBugstopper:
```suggestion
// executable and wait for completion. The file will receive a serialized
```

src/commands/build/wranglerjs/mod.rs
```diff
@@ -30,9 +30,9 @@ use std::thread;
 use std::time::Duration;
 
 // Run the underlying {wranglerjs} executable.
-//
-// In Rust we create a virtual file, pass the pass to {wranglerjs}, run the
-// executable and wait for completion. The file will receive the a serialized
+
+// In Rust we create a virtual file, pass it to {wranglerjs}, run the
+ // executable and wait for completion. The file will receive a serialized
```
ashleymichal:
```suggestion
// executable and wait for completion. The file will receive a serialized
```

src/commands/publish/upload_form/mod.rs
ashleymichal: i think this should actually be above ln 62, where we call `wranglerjs::Bundle::new` but i'm not certain precisely why.

src/commands/build/wranglerjs/mod.rs
ashleymichal: this will fix the fmt error in CI

EverlastingBugstopper: thanks!!!

src/commands/publish/upload_form/mod.rs
xtuc: My idea was that each backends (rust, webpack, js) returns a Bundle/WorkerBundle/Worker that contains all the bindinds. The creation of the form (code here), would consume it and build a form regardless of the backend used.

If that was implemented building a new bundle wouln't make sense. However, wranglerjs's bundle doesn't contains any state (only constants) so it's fine.

Feel free to remove the comment tho. We could also inline the constants for now.

src/commands/publish/upload_form/mod.rs
```diff
@@ -58,6 +58,7 @@ pub fn build(
         TargetType::Webpack => {
             log::info!("Webpack project detected. Publishing...");
             // TODO: (sven) shouldn't new
+            // see https://github.com/cloudflare/wrangler/pull/846#discussion_r342845421
```
xtuc: I think a dedicated issue would be safer and easier to discuss in

Issue #833, opened 2019-11-01
webpack_config warning doesn't fire on environments


adaptive:
### 🐛 Bug Report
The webpack requirement warning for version 1.6.0 does not fire when using environments.
when `webpack_config = "webpack.config.js"` is missing from wrangler.toml

#### Environment

On wrangler 1.5.0

#### Steps to reproduce

`wrangler publish --env staging`

#### What did you expect to see?
Warning
`In Wrangler v1.6.0, you will need to include a webpack_config field in your wrangler.toml to build with a custom webpack configuration.`

#### What did you see instead?
No warning

PR #847, opened 2019-11-05
Require webpack_config field to build with custom configuration

### LINKED ISSUE #296: Require webpack_config field in wrangler.toml to use custom webpack configuration

EverlastingBugstopper:
Fixes #296 and #833 

As promised - we are deprecating inferred usage of `webpack.config.js` in 1.6.0 and are requiring the `webpack_config` file to be set in order to use a custom webpack configuration file. 

```console
$ wrangler build
⚠️  If you would like to use your own custom webpack configuration, you will need to add this to your wrangler.toml:
webpack_config = "webpack.config.js"
✨  Built successfully, built project size is 535 bytes.
```

```console
$ wrangler build
using custom
✨  Built successfully, built project size is 535 bytes.
```

```console
$ wrangler build --env test
⚠️  If you would like to use your own custom webpack configuration, you will need to add this to your wrangler.toml:
webpack_config = "webpack.config.js"
✨  Built successfully, built project size is 535 bytes.
```

```console
$ wrangler build --env test
using custom config
✨  Built successfully, built project size is 535 bytes.
```
      
```console
$ wrangler build
⚠️  Workers Sites does not support custom webpack configuration files
✨  Built successfully, built project size is 11 KiB.
```


tests/build.rs
```diff
@@ -139,21 +143,8 @@ fn it_fails_with_multiple_webpack_configs() {
...
     single_env_settings! {fixture, r#"
-        type = "Webpack"
-    "#};
-
-    build_fails_with(fixture, "Multiple webpack configurations are not supported. You can specify a different path for your webpack configuration file in wrangler.toml with the `webpack_config` field");
-    utils::cleanup(fixture);
-}
-
-#[test]
-fn it_fails_with_multiple_specify_webpack_configs() {
```
ashleymichal: should this test still exist? it's kind of redundant but i don't think its outcome is changed by this PR.

EverlastingBugstopper: It's the same test as `it_fails_with_multiple_webpack_configs` the only difference previously was one had inferred `webpack.config.js` existed and one had it in the toml - after this breaking change they are all "specify" tests

EverlastingBugstopper:
> the utils removal is a step backwards, kind of, but makes the status quo consistent so is fine.

I brought back the util and moved a couple of other tests to it - the reason I removed it earlier was because I didn't realize that I could use the webpack config file created programatically in the wrangler.toml

motiejunas:
Could you give more details on "Workers Sites does not support custom webpack configuration files"? Basically it means that I cannot use TypeScript for Workers Sites anymore?

ashleymichal:
@motiejunas the error message is historical; when wrangler implicitly looked at the root for a custom webpack config, it made sense to disable the functionality for sites since the root would likely contain a config for the app you were deploying, not for the workers site. with this change, though, we should allow custom webpack configs since you are required to declare them in your wrangler.toml. we'll track it in a separate issue.

PR #851, opened 2019-11-06
add scaffold_worker method to Site, use all the PathBufs

### LINKED ISSUE #622: Site bucket should be PathBuf not String
check out serialization options for this

### LINKED ISSUE #643: Use site.entry_point instead of build_dir for scaffolding sites
Right now i assume the build_dir is determined entirely by entry_point. It's true, but not a good idea to leave coupled.
https://github.com/cloudflare/wrangler/pull/509/files#diff-bc87f262247b9e739224645f50263bd2R197

ashleymichal:
fixes #622
fixes #643 by moving logic to scaffold a site worker out of wranglerjs and onto the site struct.

also some 💅:
* change entry_point to `Option<PathBuf>` instead of `Option<String>`
* organize/tidy `use` statements in build/wranglerjs
* info! -> log::info!
* log before running `generate`, remove `name` argument to `generate::command` that was purely for logging
* use mut Site::default() for bucket tests, set variable fields
* rename site.build_dir() -> site.entry_point(), don't pass 'current_dir''

Happy to extract any controversial refactors to a separate PR, but I think this is all pretty tame.


EverlastingBugstopper:
> change entry_point to Option instead of Option

🤔

ashleymichal:
> > change entry_point to Option instead of Option
> 
> 🤔

oh wow. github markdown made that weird. updated the comment.

src/commands/init/mod.rs
```diff
@@ -19,20 +18,24 @@ pub fn init(
...
-        commands::build::wranglerjs::scaffold_site_worker(&target)?;
+    if site_flag {
+        let site = Site::default();
+        Manifest::generate(
+            name.to_string(),
+            Some(target_type),
+            &config_path,
+            Some(site.clone()),
+        )?;
+        message::success("Succesfully created a `wrangler.toml`");
```
EverlastingBugstopper: nit: move duplicate success messages right above `Ok(())`

src/settings/target/site.rs
```diff
@@ -1,14 +1,18 @@
...
+use crate::commands::generate::run_generate;
+
 const SITE_ENTRY_POINT: &str = "workers-site";
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct Site {
     pub bucket: String,
     #[serde(rename = "entry-point")]
-    pub entry_point: Option<String>,
+    entry_point: Option<PathBuf>,
```
EverlastingBugstopper: this smells like you could fix up `bucket` while you're at https://github.com/cloudflare/wrangler/issues/622

EverlastingBugstopper: (but you don't have to)

src/commands/publish/mod.rs
```diff
@@ -158,7 +157,10 @@ pub fn upload_buckets(
     let mut asset_manifest = None;
     for namespace in &target.kv_namespaces() {
         if let Some(bucket) = &namespace.bucket {
-            if bucket.is_empty() {
+            // We don't want folks setting their bucket to the top level directory,
+            // which is where wrangler commands are always called from.
+            let current_dir = env::current_dir()?;
+            if bucket.as_os_str() == current_dir {
```
EverlastingBugstopper: can we make issue for adding test for this (and maybe other workers sites stuff)

src/settings/target/site.rs
ashleymichal: DID IT

src/commands/publish/mod.rs
ashleymichal: Probably!

src/settings/target/site.rs
```diff
@@ -1,43 +1,64 @@
...
 
         site
     }
 
     // if the user has configured `site.entry-point`, use that
     // as the build directory. Otherwise use the default const
     // SITE_ENTRY_POINT
-    pub fn build_dir(&self, current_dir: PathBuf) -> Result<PathBuf, std::io::Error> {
+    pub fn entry_point(&self) -> Result<PathBuf, std::io::Error> {
+        let current_dir = env::current_dir()?;
```
ashleymichal: eventually methods like these that determine where a file should be found relative to the project root should probably take as an argument the actual project root path, rather than the current directory. for the moment, everything runs relative to root because that is where wrangler.toml is and we assume you are running wrangler adjacent to the toml.

ashleymichal: which makes me want to write a refactor issue

src/commands/publish/mod.rs
ashleymichal: https://github.com/cloudflare/wrangler/issues/858

src/settings/target/site.rs
gabbifish: ^ this would also be a step towards better monorepo support!

ashleymichal: and/or just being able to run wrangler from anywhere inside a project. i'll write the issue and merge this.

ashleymichal: https://github.com/cloudflare/wrangler/issues/860

PR #853, opened 2019-11-06
Clean up wranglerjs temp file

EverlastingBugstopper:
When building a script, wranglerjs creates a temporary file to pass it off to wrangler. Unfortunately, it never removed this file. Now it does.


EverlastingBugstopper:
yes @xtuc watching preview still works for `js`, `webpack`, and `rust` type projects.

PR #854, opened 2019-11-06
Remove the fixtures directory

EverlastingBugstopper:
wave goodbye to the fixtures directory

Wrangler no longer relies on reading fixtures from the file system and instead creates them in the tests themselves.


tests/preview.rs
```diff
@@ -1,77 +1,191 @@
...
-    utils::create_temporary_copy(fixture);
-    settings! {fixture, r#"
+    let fixture = Fixture::new("webpack_simple_js");
+    fixture.scaffold_webpack();
+    fixture.create_wrangler_toml(
+        r#"
         type = "webpack"
-    "#};
+    "#,
+    );
```
gabbifish: do we also want to call "cleanup" at the end of every one of these tests? I feel like that would be advisable.

ashleymichal: he added it to the `preview` helper function: https://github.com/cloudflare/wrangler/pull/854/files#diff-eb986e3abcfb440d6233373e1d67f78dR190

gabbifish: gotcha! I would prefer for cleanup to explicitly happen as its own call instead of being put in preview(), given that preview() should not have cleanup as a side effect. OR change preview() to preview_and_cleanup().

ashleymichal: fwiw, as part of my test refactor work i will likely update fixtures to conditionally clean up based on whether tests pass, per the test fixture library used by wasm_pack.

EverlastingBugstopper: Doesn't make sense for me to call cleanup separately I think - `preview` will both create the files and clean them up. We don't actually create the files until we enter the `preview` function even though we call `fixture.createFile` that's just a string in memory

gabbifish:
Hmm, how about a function name that describes file creation, preview, and
deletion then? Lumping all three of those under preview is still too many
side effects for an otherwise very simply named function

gabbifish:
(Just being nit picky from a maintainability standpoint! I shouldn’t have
to dig for where features live if function names can help make that evident
:)

EverlastingBugstopper:
@gabbifish makes sense! addressed

PR #871, opened 2019-11-14
[WIP] Dev server feature branch

EverlastingBugstopper:
This PR adds `wrangler proxy` as a subcommand and spins up a local webserver for whatever worker you want and then you can curl it.

TODO: live reload, figure out the API we want, clean up the code


EverlastingBugstopper:
#883 is the new feature branch

Issue #874, opened 2019-11-14
[dev] add a flag to customize the IP the proxy listens on


EverlastingBugstopper:
@xtuc says it could be useful for Docker and that we shouldn't just default it to 127.0.0.1

xtuc:
To be clear the default should be localhost. However, the developer might want to bind the proxy on another address.

PR #878, opened 2019-11-14
Refactor proxy and add IP argument

EverlastingBugstopper:
Branches off of #871 and solves #874 

I've introduced a `Proxy` struct but I'm not sure if I like it. Very open to some pointers here, I'm doing a lot of clones because of that pesky `move` and something about `Fn` vs `FnOnce` which has confused me quite a bit.

**Usage:**

---

```console
$ wrangler proxy
Listening on http://localhost:8000
```
worker preview sends origin requests to https://example.com

---

```console
$ wrangler proxy -u http://example.com
Listening on http://localhost:8000
```
worker preview sends origin requests to http://example.com

---

```console
$ wrangler proxy -h averyharnish.com
Listening on http://localhost:8000
```
worker preview sends origin requests to https://averyharnish.com

---

```console
$ wrangler proxy -i 127.0.0.1 -p 8080
Listening on http://127.0.0.1:8080
```
worker preview sends origin requests to https://example.com

---

### DIFF
--- a/src/commands/preview/proxy.rs
+++ b/src/commands/preview/proxy.rs
@@ -4,6 +4,8 @@ extern crate hyper_tls;
 extern crate pretty_env_logger;
 extern crate url;
 
+use std::net::{SocketAddr, ToSocketAddrs};
+
 use chrono::prelude::*;
 
 use hyper::header::{HeaderName, HeaderValue};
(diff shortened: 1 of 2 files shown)


EverlastingBugstopper:
closing this in favor of an async refactor

Issue #867, opened 2019-11-13
Don't use Config crate


EverlastingBugstopper:
It's introducing bugs like #704 and it's also [no longer maintained](https://github.com/mehcode/config-rs/issues/111). EDIT: development seems to have continued, but stagnated just a bit, I still think we should move to our own handling since our use case is straigthforward

We really only use it for merging environment variables with our manifest configuration, which we should be able to do in house by combining `std::env` `toml` and `serde`

Instead of looping over all environment vairables looking for a prefix, we should loop over a whitelist of our supported environment variables.

When constructing `GlobalUser`, we should loop over CF_API_TOKEN, CF_EMAIL, and CF_API_KEY.

When constructing `Manifest`, we should loop over CF_ZONE_ID, CF_ACCOUNT_ID, CF_NAME, CF_TYPE, CF_WORKERS_DEV, CF_ROUTE, and CF_WEBPACK_CONFIG. Note that this leaves out CF_ENV, CF_ROUTES, CF_KV_NAMESPACES, and CF_SITE which we currently try to parse. These are not strings and cannot be set in an environment variable, so we should not attempt to parse them. This will eliminate the parsing bug seen in #704 and will also (very slightly) improve performance on machines with a large number of environment variables.

This refactor would also have the benefit of making our code more readable and easier to document which environment variables are officially supported.

ashleymichal:
there is a struct for this...

EverlastingBugstopper:
i believe i opened this issue before there was - mind updating to point folks at the right place to start for this one?

ashleymichal:
we could mark  as duplicate of https://github.com/cloudflare/wrangler/issues/852

PR #879, opened 2019-11-14
Prefer environment variables over config files

### LINKED ISSUE #834: prefer environment variables over config file; require complete user config
per discussion https://github.com/cloudflare/wrangler/pull/471#discussion_r341769828:

we should be clear about precedence for user configuration; i.e., we need to explain what happens when you have both a config file and environment variables defined, and what happens if you have both token and global api key/email defined.

e.g. (strawman): if you have both a config file and env vars, env vars take precedence; if you have all three env vars defined, api token takes precedence.

ashleymichal:
fixes #834

also some gnarly rust to make it so we can control env vars in unit tests.

I can split this into a few PRs if the various refactors are too hard to grok; else I recommend reading this commit by commit, they are relatively clean.


src/commands/config/mod.rs
```diff
@@ -27,21 +27,18 @@ pub fn global_config(user: &GlobalUser, verify: bool) -> Result<(), failure::Err
...
     }
 
-    let toml = toml::to_string(&user)?;
-
     let config_dir = get_global_config_dir().expect("could not find global config directory");
     fs::create_dir_all(&config_dir)?;
 
-    let config_file = config_dir.join("default.toml");
-    fs::write(&config_file, &toml)?;
+    let config_path = user.to_file(&config_dir)?;
```
ashleymichal: by moving this logic into the GlobalUser struct, we keep the name of the file in the same place.

src/settings/environment.rs
```diff
@@ -0,0 +1,109 @@
+use std::collections::HashMap;
+use std::env;
+
+use config::{Source, Value};
+
+pub trait QueryEnvironment {
```
ashleymichal: This trait wraps calls to the environment, which allows us to also implement a Mock environment for use in unit tests.

src/settings/environment.rs
```diff
@@ -0,0 +1,109 @@
...
+    fn get_var(&self, var: &'static str) -> Result<String, std::env::VarError>;
+
+    fn empty(&self) -> Result<bool, failure::Error>;
```
ashleymichal: this is a convenience method. i'm not... 100% certain about keeping this here, but yeah it does "query" the environment.

src/settings/environment.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+#[derive(Clone, Debug)]
+pub struct Environment {
+    whitelist: Vec<&'static str>,
```
ashleymichal: this whitelist means we don't pull the entire environment in to assess things, which is nice.

src/settings/environment.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+impl Environment {
+    pub fn with_whitelist(whitelist: Vec<&'static str>) -> Self {
+        Environment { whitelist }
```
ashleymichal: a nice convenience constructor. possibly unnecessary

src/settings/global_user.rs
```diff
@@ -1,12 +1,21 @@
 use std::env;
+use std::fs;
 use std::path::{Path, PathBuf};
 
 use cloudflare::framework::auth::Credentials;
-use log::info;
+use config;
```
ashleymichal: i did this to be obnoxious. there's even a commit that says it. it was spiteful and i can revert 😂 but there is an issue out for removing config from the project.

src/settings/global_user.rs
```diff
@@ -1,12 +1,21 @@
...
 use serde::{Deserialize, Serialize};
 
+use crate::settings::{Environment, QueryEnvironment};
 use crate::terminal::emoji;
-use config::{Config, Environment, File};
+
+const DEFAULT_CONFIG_FILE_NAME: &str = "default.toml";
```
ashleymichal: here's our const! but it's also possible that the "get_config_dir()" function + this should all just be one big const. idk.

ashleymichal: update: sort of made it so

src/commands/config/mod.rs
```diff
@@ -1,17 +1,16 @@
...
-use crate::settings::global_user::{get_global_config_dir, GlobalUser};
-
 use cloudflare::endpoints::user::{GetUserDetails, GetUserTokenStatus};
 use cloudflare::framework::apiclient::ApiClient;
 use cloudflare::framework::HttpApiClientConfig;
 
+use crate::http;
+use crate::settings::global_user::{default_config_file, GlobalUser};
+use crate::terminal::message;
+
```
ashleymichal: import shuffling, moving file writes to GlobalUser.to_file

src/commands/config/mod.rs
```diff
@@ -27,13 +26,8 @@ pub fn global_config(user: &GlobalUser, verify: bool) -> Result<(), failure::Err
...
 
-    let toml = toml::to_string(&user)?;
-
-    let config_dir = get_global_config_dir().expect("could not find global config directory");
-    fs::create_dir_all(&config_dir)?;
-
-    let config_file = config_dir.join("default.toml");
-    fs::write(&config_file, &toml)?;
+    let config_file = default_config_file().expect("could not find global config directory");
+    user.to_file(&config_file)?;
```
ashleymichal: moving file writes to GlobalUser.to_file

src/settings/environment.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+}
+
+// Source trait implementation for use with Config::merge
+// until config crate removal is complete. This is effectively a
+// copy of the config crate's impl of Source for its Environment
+// struct, but rather than pulling the whole environment in #collect,
+// we pull in only whitelisted values, and rather than taking a custom
+// prefix, we assume a prefix of `CF_`.
+impl Source for Environment {
```
ashleymichal: keeping the `collect` implementation is nice, though, so perhaps this will not go away in the long term, only behave differently.

src/settings/environment.rs
```diff
@@ -0,0 +1,109 @@
...
+#[derive(Clone, Debug, Default)]
+pub struct MockEnvironment {
+    vars: Vec<(&'static str, &'static str)>,
+}
+
+impl MockEnvironment {
+    pub fn set(&mut self, key: &'static str, value: &'static str) -> &Self {
+        self.vars.push((key, value));
+
+        self
```
ashleymichal: gives nice way to control our mock "environment" while preserving the interface below.

src/commands/config/mod.rs
```diff
@@ -27,13 +26,8 @@ pub fn global_config(user: &GlobalUser, verify: bool) -> Result<(), failure::Err
...
     }
 
-    let toml = toml::to_string(&user)?;
-
-    let config_dir = get_global_config_dir().expect("could not find global config directory");
-    fs::create_dir_all(&config_dir)?;
-
-    let config_file = config_dir.join("default.toml");
-    fs::write(&config_file, &toml)?;
+    let config_file = default_config_file().expect("could not find global config directory");
```
EverlastingBugstopper: is there an actionable here or is this just a bad state? what should the user do to recover?

src/settings/environment.rs
EverlastingBugstopper: great step towards #867

src/commands/config/mod.rs
ashleymichal: that's a great question! atm this is copy 🍝

src/settings/environment.rs
```diff
@@ -0,0 +1,109 @@
...
+            // remove the `CF` prefix before adding to collection
+            let prefix_pattern = "CF_";
+            let key = &key[prefix_pattern.len()..];
+
+            m.insert(key.to_lowercase(), Value::new(Some(&uri), *value));
+        }
+
+        Ok(m)
+    }
+}
```
ashleymichal: i should actually add a couple of regression tests here for the behavior of Environment. if this is the only unit test that uses them, that would be great 😁

src/settings/environment.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    fn collect(&self) -> Result<HashMap<String, Value>, ConfigError> {
+        let mut m = HashMap::new();
+        let uri: String = "env".into();
+
+        for key in &self.whitelist {
+            if let Some(value) = env::var(key).ok() {
+                // remove the `CF` prefix before adding to collection
+                let prefix_pattern = "CF_";
+                let key = &key[prefix_pattern.len()..];
```
EverlastingBugstopper: nit: i think i prefer `let key = key.replace("CF_", "");` here given that `key` is an `&str` we can take advantage of `std::str::replace`

EverlastingBugstopper: either that or it may be worth adding an `if key.starts_with("CF_")` before doing the trim.

ashleymichal: this is ripped directly from `config::Environment`; replace will replace _anywhere_, not just at the front. unlikely to cause problems, but i believe in murphy's law.

src/settings/environment.rs
```diff
@@ -0,0 +1,109 @@
...
+
+        for (key, value) in &self.vars {
+            // remove the `CF` prefix before adding to collection
+            let prefix_pattern = "CF_";
+            let key = &key[prefix_pattern.len()..];
+
+            m.insert(key.to_lowercase(), Value::new(Some(&uri), *value));
+        }
+
+        Ok(m)
```
EverlastingBugstopper: is it possible to re-use the code for `collect` between the real and the mock environment?

src/settings/global_user.rs
```diff
@@ -17,7 +26,90 @@ pub enum GlobalUser {
...
+        config_path: PathBuf,
+    ) -> Result<Self, failure::Error>
+    where
+        T: config::Source + Send + Sync,
+    {
+        if let Some(user) = Self::from_env(environment) {
+            user
+        } else {
+            Self::from_file(config_path)
+        }
```
EverlastingBugstopper: i want to make sure that this does not change previous behavior - did we ever allow reading in an email from an environment variable and an api key from the config file? perhaps someone just wants to override one or the other. not sure if we previously supported that or not (i think we probably did?)

EverlastingBugstopper: i'm not saying we _should_ but we might have, and this would be a breaking change if that were the case

src/settings/global_user.rs
```diff
@@ -17,7 +26,90 @@ pub enum GlobalUser {
...
+
+    // TODO: This fn should return None if the environment has none of the
+    // relevant variables
+    fn from_env<T: 'static + QueryEnvironment>(
+        environment: T,
+    ) -> Option<Result<Self, failure::Error>>
+    where
+        T: config::Source + Send + Sync,
+    {
+        if environment.empty().unwrap() {
```
EverlastingBugstopper: why do we need an `unwrap` here? is it a `Result`? In what scenarios would `environment.empty()` return an error, and can we use `?` to bubble up the error instead of unwrapping it?

src/settings/global_user.rs
```diff
@@ -17,7 +26,90 @@ pub enum GlobalUser {
...
+        // Skip reading global config if non existent
+        // because envs might be provided
+        if config_path.exists() {
+            log::info!(
+                "Config path exists. Reading from config file, {}",
+                config_str
+            );
+            s.merge(config::File::with_name(config_str))?;
+        } else {
+            failure::bail!("config path does not exist {}", config_str);
```
EverlastingBugstopper: how can the user recover from this? can we give them an actionable?

src/settings/environment.rs
ashleymichal: perhaps for the prefix stuff. might also benefit from including a whitelist property in the mock.

src/settings/global_user.rs
```diff
@@ -33,80 +125,122 @@ impl From<GlobalUser> for Credentials {
...
+        log::info!("No WRANGLER_HOME detected");
         dirs::home_dir()
             .expect("oops no home dir")
             .join(".wrangler")
     };
-    let global_config_dir = home_dir.join("config");
-    info!("Using global config dir: {:?}", global_config_dir);
-    Ok(global_config_dir)
+    let global_config_file = home_dir.join("config").join(DEFAULT_CONFIG_FILE_NAME);
+    log::info!("Using global config file: {:?}", global_config_file);
```
EverlastingBugstopper: nit: can we use `.display()` here for our windows folks

src/settings/global_user.rs
ashleymichal: we did; however this is the behavior we landed on in the discussion in #834

src/settings/global_user.rs
ashleymichal: it's not a result, it's an option.

ashleymichal: env::var() has a few different errors. perhaps we should match on some of those in `collect`? because the most likely error is that the env var simply doesn't exist, and that's not an error for us.

src/settings/global_user.rs
ashleymichal: idk, copy pasta.

src/settings/global_user.rs
```diff
@@ -33,80 +125,122 @@ impl From<GlobalUser> for Credentials {
...
             GlobalUser::TokenAuth {
                 api_token: "foo".to_string()
             }
         );
     }
+
+    #[test]
+    fn it_can_prioritize_env_vars() {
+        let api_token = "thisisanapitoken";
+        let api_key = "reallylongglobalyapikey";
```
EverlastingBugstopper: nit: `globalyapikey` -> `globalapikey`

src/settings/global_user.rs
```diff
@@ -33,80 +125,122 @@ impl From<GlobalUser> for Credentials {
...
+
+        let new_user = GlobalUser::build(mock_env, config_dir);
+
+        match new_user {
+            Ok(_) => assert!(false),
+            Err(_) => assert!(true),
+        }
+    }
+
+    fn test_config_dir(
```
EverlastingBugstopper: i think from a readability perspective i would want this function to create the temp directory itself instead of having it passed in by every test, and i'd also want to call it something like `get_user_config_path` to match the name of the argument we're passing to `GlobalUser::build` which is `config_path`

ashleymichal: i would too! the problem is temp dirs die when they go out of scope

ashleymichal:
>  we're not changing prior behavior wrt all or nothing precedence with environment variables and backwards compatibility.

the changes implemented here are changes we agreed upon in the issue.

src/settings/environment.rs
EverlastingBugstopper: yeah the `.replace` probably not the best, but the `starts_with`

src/settings/environment.rs
ashleymichal: you can't really extract calls to `self` tho.

EverlastingBugstopper:
i understand that we've agreed upon this in the issue and i still think that is the way we should be doing things for folks. i just want some confidence that we aren't gonna break people who currently have an api key in their global config and are using an environment variable for email. it doesn't make sense for them to be doing that, but if it's possible today, we shouldn't roll out something that will break them.

src/settings/global_user.rs
ashleymichal: an `.unwrap_or(true)`. incoming.

src/settings/environment.rs
ashleymichal: we might even do some other voodoo here with not explicitly setting the consts to start with `CF_` but programmatically prepending before we do gets so we _know_ it starts with `CF_`

src/settings/environment.rs
gabbifish: This would make a really good inline comment; can we add it?

src/settings/global_user.rs
```diff
@@ -17,7 +26,94 @@ pub enum GlobalUser {
...
+
+        // Skip reading global config if non existent
+        // because envs might be provided
+        if config_path.exists() {
+            log::info!(
+                "Config path exists. Reading from config file, {}",
+                config_str
+            );
+            s.merge(config::File::with_name(config_str))?;
+        } else {
```
gabbifish: Should we just print a warning and not bail here? If we bail here, it seems that the fallback case (allowing people to just use env vars without config files, as they would in CI) would fail.

src/settings/global_user.rs
```diff
@@ -33,80 +129,122 @@ impl From<GlobalUser> for Credentials {
...
+        let tmp_dir = tempdir().unwrap();
+        let tmp_config_path = test_config_dir(&tmp_dir, Some(file_user)).unwrap();
+
+        let new_user = GlobalUser::build(mock_env, tmp_config_path).unwrap();
+
+        assert_eq!(new_user, env_user);
+    }
+
+    #[test]
+    fn it_falls_through_to_config_with_no_env_vars() {
```
gabbifish: Let's add a test to verify that environment variables work when no config directory is set up (the CI use case)

src/settings/global_user.rs
```diff
@@ -17,7 +26,94 @@ pub enum GlobalUser {
...
+    where
+        T: config::Source + Send + Sync,
+    {
+        // if there's some problem with gathering the environment,
+        // or if there are no relevant environment variables set,
+        // fall back to config file.
+        if environment.empty().unwrap_or(true) {
+            None
+        } else {
+            let mut s = config::Config::new();
```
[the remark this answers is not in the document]
ashleymichal: yes that was some of the point

src/settings/environment.rs
```diff
@@ -0,0 +1,139 @@
...
+        Box::new((*self).clone())
+    }
+
+    fn collect(&self) -> Result<HashMap<String, Value>, ConfigError> {
+        let mut m = HashMap::new();
+        let uri: String = "env".into();
+
+        for (key, value) in &self.vars {
+            // remove the `CF` prefix before adding to collection
+            let prefix_pattern = "CF_";
```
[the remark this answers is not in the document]
ashleymichal: looks like it did but didn't get used?

src/settings/global_user.rs
```diff
@@ -1,12 +1,21 @@
...
 use crate::terminal::emoji;
-use config::{Config, Environment, File};
+
+const DEFAULT_CONFIG_FILE_NAME: &str = "default.toml";
+
+const CF_API_TOKEN: &str = "CF_API_TOKEN";
+const CF_API_KEY: &str = "CF_API_KEY";
+const CF_EMAIL: &str = "CF_EMAIL";
+
+static ENV_VAR_WHITELIST: [&str; 3] = [CF_API_TOKEN, CF_API_KEY, CF_EMAIL];
```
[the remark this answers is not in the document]
ashleymichal: i let the compiler guide me to this; any advantage to removing the size declaration if it's unlikely to change?

PR #880, opened 2019-11-15
Some environment variable precedence tests and messaging improvements

EverlastingBugstopper:
This PR is meant to address the concerns raised in #879 

It also adds some visual space to the `wrangler config` command

image

### DIFF
--- a/src/settings/global_user.rs
+++ b/src/settings/global_user.rs
@@ -1,5 +1,6 @@
 use std::env;
 use std::fs;
+use std::io::prelude::*;
 use std::path::{Path, PathBuf};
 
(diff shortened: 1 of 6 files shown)


src/settings/global_user.rs
```diff
@@ -228,10 +244,18 @@ mod tests {
...
 
-        match new_user {
-            Ok(_) => assert!(false),
-            Err(_) => assert!(true),
-        }
+        assert!(new_user.is_err());
+    }
+
+    #[test]
+    fn it_succeeds_with_no_config() {
```
gabbifish: HECK YEAH

PR #882, opened 2019-11-18
[WIP] Use async syntax for proxy

EverlastingBugstopper:
This is functionally the same as the proof of concept in the feature branch except it adds an optional `--ip` flag. It doesn't solve any tickets, but it provides a better structure with which to move forward.


EverlastingBugstopper:
closing in favor of #883 because i am very bad at git

PR #883, opened 2019-11-19
[rc] [dev] Dev Server Feature Branch

### LINKED ISSUE #845: Create a dev server
This issue is meant to be a summary of everything associated with [this milestone](https://github.com/cloudflare/wrangler/milestone/18) - for more detailed information, please look at the issues and PRs listed there.

The overarching goal is to create a "local" development environment for Workers that only requires an Internet connection. Expose the server on a configurable IP and port, and take care of uploading the script to the preview service and sending the necessary authentication headers. 

This feature will proxy requests to rawhttp.cloudflareworkers.com and future work will be built on top of how this service is currently implemented.

The proxy will need to mimic what the UI does:

- Send the same path to execute the worker to the preview endpoint
- Send the same HTTP method to execute the worker to the preview endpoint
- Append `cf-ew-` to all request headers #873 
- Strip `cf-ew-` from response headers #872 
- Pipe console.log output via the chrome devtools protocol #875 

The proxy will depend on the following endpoints:

1) https://rawhttp.cloudflareworkers.com
1) https://api.cloudflare.com/client/v4/accounts/{}/workers/scripts/{}/preview
1) https://cloudflareworkers.com/script

EverlastingBugstopper:
Fixes #845

# Wrangler Dev Server -- _Feature Branch_

**[Testing Instructions](https://gist.github.com/EverlastingBugstopper/6e964c8919afdb97397c74fcb1097ccf)** - these should be included in the rc when it gets cut

---

## Testing

Your local `rustc -V` must be at least 1.39.0. You can update by running `rustup update stable`.

You can install this branch by running `cargo install --git https://github.com/cloudflare/wrangler --branch avery/a-time-sync --debug --force`

_(eventually that will get merged and you can install from this branch)_

Go to a worker directory and run `wrangler dev` and curl/postman away!

```console
$ wrangler dev --help
wrangler-dev 
👂  Start a local server for developing your worker

USAGE:
    wrangler dev [OPTIONS]

FLAGS:
        --help    Prints help information

OPTIONS:
    -h, --host <host>    domain to test behind your worker. defaults to example.com
    -i, --ip <ip>        ip to listsen on. defaults to localhost
    -p, --port <port>    port to listen on. defaults to 8787
```


src/commands/preview/proxy.rs
```diff
@@ -20,104 +18,175 @@ use url::Url;
...
+            Some(host) => host.to_string(),
+            None => "https://example.com".to_string(),
+        };
+
+        let parsed_url = match Url::parse(&host) {
+            Ok(host) => Ok(host),
+            Err(_) => Url::parse(&format!("https://{}", host)),
+        }?;
+
+        let scheme: &str = parsed_url.scheme();
```
steveklabnik: is the type annotation needed here?

src/commands/preview/proxy.rs
```diff
@@ -20,104 +18,175 @@ use url::Url;
...
+            )),
+        }?;
+
+        let proxy = ProxyConfig {
+            listening_address,
+            host,
+            is_https,
+        };
+
+        Ok(proxy)
```
steveklabnik: I don't know what style wrangler uses, but in these cases, i tend to not introduce the `let` above, that is, I just write

    Ok(ProxyConfig {

but I can see things either way

src/commands/preview/proxy.rs
```diff
@@ -20,104 +18,175 @@ use url::Url;
...
+    }
+
+    fn get_listening_address_as_str(&self) -> String {
```
steveklabnik: this name feels weird because it says str but returns String. also, `get_` is not usually a name for getters, see https://rust-lang.github.io/api-guidelines/naming.html?#getter-names-follow-rust-convention-c-getter

it also seems like this method doesn't just return the address as a string, but also turns the ipv6 version into an ipv4 version?

src/commands/preview/proxy.rs
```diff
@@ -20,104 +18,175 @@ use url::Url;
...
-    }?;
-    let scheme: &str = parsed_url.scheme();
-    if scheme != "http" && scheme != "https" {
-        failure::bail!("Your host must be either http or https")
+    let proxy_config = ProxyConfig::new(host, ip, port)?;
+    let https = HttpsConnector::new().expect("TLS initialization failed");
+    let client = Client::builder().build::<_, Body>(https);
+
+    let preview_id = get_preview_id(target, user, &proxy_config)?;
+    let listening_address = &proxy_config.listening_address.clone();
```
steveklabnik: doing `&foo.clone()` always feels a bit weird; I would drop the `&` here and use the `&` at the call site below personal preference.

src/commands/preview/proxy.rs
```diff
@@ -20,104 +18,175 @@ use url::Url;
...
+    let listening_address_str = proxy_config.get_listening_address_as_str();
```
steveklabnik: you only use this variable once, and the `_str` naming is a bit odd; i'd personally put this inline in the `println` below

src/commands/preview/proxy.rs
```diff
@@ -20,104 +18,175 @@ use url::Url;
...
-    let https = HttpsConnector::new(4).expect("TLS initialization failed");
-    let client_main = Client::builder().build::<_, hyper::Body>(https);
+    Ok(())
+}
 
+fn get_preview_url(
+    path_string: &str
+) -> Result<Uri, http::uri::InvalidUri> {
+    let uri_string = format!("https://{}{}", PREVIEW_HOST, path_string);
+    uri_string.parse()
```
steveklabnik: i would write this as a one liner, i'm not sure which is actually better

src/commands/preview/proxy.rs
```diff
@@ -20,104 +18,175 @@ use url::Url;
...
+        HeaderValue::from_static(PREVIEW_HOST),
+    );
+
+    // TODO: remove unwrap
+    parts.headers
+        .insert(HeaderName::from_static("cf-ew-preview"), HeaderValue::from_str(preview_id).unwrap());
+    
+    let req = Request::from_parts(parts, body);
+
+
```
steveklabnik: you've got an extra newline here

EverlastingBugstopper:
thanks for the review @steveklabnik - super super helpful :)

src/commands/preview/proxy.rs
EverlastingBugstopper: This is purely for printing purposes - I want to print `http://localhost:8000` instead of `http://[::1]:8000`. Do you think that will cause problems? Should I be printing this differently?

src/commands/preview/proxy.rs
EverlastingBugstopper: was running into some borrow checker related things there - those nested `move` statements really boggle my brain

src/commands/preview/proxy.rs
steveklabnik: I don't think it would cause problems, I think the hard part is coming up with a good name that communicates what it does. *why* do you want to print it that way?

src/commands/preview/dev_server.rs
```diff
@@ -0,0 +1,190 @@
+use std::net::{SocketAddr, ToSocketAddrs};
```
ashleymichal: because it is so specialized, and could turn into a weird part of the codebase for onboarding, perhaps this module deserves some heavy use of comments.

src/commands/preview/dev_server.rs
```diff
@@ -0,0 +1,190 @@
...
+    is_https: bool,
+}
+
+impl ServerConfig {
+    pub fn new(
+        host: Option<&str>,
+        ip: Option<&str>,
+        port: Option<&str>,
+    ) -> Result<Self, failure::Error> {
+        let port = port.unwrap_or("8000");
```
ashleymichal: nit: i like this pattern for establishing the default values of these arguments, i think it would make things clearer to do the same for all three arguments right here at the top. it would make for more local variables, but would  get that out of the way quicker and i think drop a couple of weird match statements.

src/commands/preview/dev_server.rs
```diff
@@ -0,0 +1,190 @@
...
+
+        let try_address = match ip {
+            Some(ip) => format!("{}:{}", ip, port),
+            None => format!("localhost:{}", port),
+        };
+
+        let mut address_iter = try_address.to_socket_addrs()?;
```
ashleymichal: this seems somewhat arcane; does it make sense to pull it out into a named function?

src/commands/preview/dev_server.rs
```diff
@@ -0,0 +1,190 @@
...
+
+        let listening_address = address_iter
+            .next()
+            .ok_or_else(|| format_err!("Could not parse address {}", try_address))?;
+
+        let host = host.unwrap_or("https://example.com").to_string();
```
ashleymichal: put this at the top

src/commands/preview/dev_server.rs
```diff
@@ -0,0 +1,190 @@
...
+
+        let parsed_url = match Url::parse(&host) {
```
ashleymichal: same here, this seems odd; is this the only reason the parse would fail? this would make another good named function, or a comment might help.

src/commands/preview/dev_server.rs
```diff
@@ -0,0 +1,190 @@
...
+            failure::bail!("Your host scheme must be either http or https")
+        }
+        let is_https = scheme == "https";
+
+        let host = parsed_url.host_str().ok_or(format_err!("Invalid host, accepted formats are example.com, http://example.com, or https://example.com"))?.to_string();
+
+        Ok(ServerConfig {
+            listening_address,
+            host,
+            is_https,
```
ashleymichal: make this a method rather than a struct property

src/commands/preview/dev_server.rs
```diff
@@ -0,0 +1,190 @@
...
+        })
+    }
+
+    fn listening_address_as_string(&self) -> String {
+        self.listening_address
+            .to_string()
+            .replace("[::1]", "localhost")
```
ashleymichal: a comment would be helpful here

src/commands/preview/dev_server.rs
```diff
@@ -0,0 +1,190 @@
...
+    target: Target,
+    user: Option<GlobalUser>,
+    host: Option<&str>,
+    port: Option<&str>,
+    ip: Option<&str>,
+) -> Result<(), failure::Error> {
+    commands::build(&target)?;
+    let server_config = ServerConfig::new(host, ip, port)?;
+    let https = HttpsConnector::new().expect("TLS initialization failed");
+    let client = Client::builder().build::<_, Body>(https);
```
ashleymichal: for sure a comment here, if only for those unfamiliar with the library

src/commands/preview/dev_server.rs
```diff
@@ -0,0 +1,190 @@
...
+            Err(_) => Url::parse(&format!("https://{}", host)),
+        }?;
+
+        let scheme = parsed_url.scheme();
+        if scheme != "http" && scheme != "https" {
+            failure::bail!("Your host scheme must be either http or https")
+        }
+        let is_https = scheme == "https";
+
+        let host = parsed_url.host_str().ok_or(format_err!("Invalid host, accepted formats are example.com, http://example.com, or https://example.com"))?.to_string();
```
ashleymichal: I suggest storing this as a `Url` rather than as a `String`, and adding an accessor method for `host_string` and one for `is_https` rather than storing these two specialized values directly on the struct.

src/commands/preview/dev_server.rs
```diff
@@ -0,0 +1,190 @@
...
+
+        Ok(ServerConfig {
+            listening_address,
+            host,
+            is_https,
+        })
+    }
+
+    fn listening_address_as_string(&self) -> String {
```
ashleymichal: this starts to smell like listening address deserves its own struct with methods. idk tho.

src/commands/preview/dev_server.rs
```diff
@@ -0,0 +1,190 @@
...
+fn get_path_as_str(uri: Uri) -> String {
+    uri.path_and_query()
+        .map(|x| x.as_str())
+        .unwrap_or("")
+        .to_string()
+}
+
+fn preview_request(
+    req: Request<Body>,
+    client: Client<HttpsConnector<HttpConnector>>,
```
ashleymichal: not that i am opposed, but this appears to introduce a new client

EverlastingBugstopper: this reuses the same client we created in the other method - it's a parameter

EverlastingBugstopper:
should probably change the `-h` https://twitter.com/jessfraz/status/1203209066501197824?s=20

ashleymichal:
> should probably change the `-h` https://twitter.com/jessfraz/status/1203209066501197824?s=20

to be fair, as someone else pointed out in the same thread, `-h => host` is older than `-h => help`. i think in this context it is fine.

EverlastingBugstopper:
@ashleymichal all of your previous comments should be addressed now

PR #884, opened 2019-11-19
[dev] Handles cf-ew-raw- headers

### LINKED ISSUE #873: [dev] add cf-ew headers to request
The preview service will send any headers prefixed with `cf-ew-` along to the worker so we should make the proxy modify these headers before sending them.

EverlastingBugstopper:
Fixes #873 

worker code returns stringified headers in the response body:

```js
  let headers = new Map(request.headers)
  return new Response(mapToString(headers))
}
```

```console
$ curl http://localhost:8000 -H "Hi: There"
{"accept":"*/*","hi":"There","host":"example.com","user-agent":"curl/7.64.1","x-forwarded-proto":"https"}
```


src/commands/dev/mod.rs
```diff
@@ -88,6 +88,17 @@ fn preview_request(
     let now: DateTime<Local> = Local::now();
     let preview_id = &preview_id;
 
+    let mut headers: HeaderMap = HeaderMap::new();
+
+    for header in &parts.headers {
+        let (name, value) = header;
+        let forward_header = format!("cf-ew-raw-{}", name);
+        // TODO: remove unwrap
+        let header_name = HeaderName::from_bytes(forward_header.as_bytes()).unwrap();
```
gabbifish: This function returns a Result with error type failure::Error, can we just get rid of this unwrap now? If the error type returned by calling `?` does not work with failure::Error, we can just capture the error and wrap it in a failure::Error

EverlastingBugstopper: This function actually returns a ResponseFuture, and I'm not quite sure how to make that play nicely with the failure crate which is why i have the TODO statements. cc @ashleygwilliams

src/commands/dev/mod.rs
```diff
@@ -47,12 +49,19 @@ pub async fn dev(
...
             Ok::<_, failure::Error>(service_fn(move |req| {
-                preview_request(
-                    req,
-                    client.to_owned(),
-                    preview_id.to_owned(),
-                    server_config.clone(),
-                )
+                let client = client.to_owned();
+                let preview_id = preview_id.to_owned();
+                let server_config = server_config.clone();
```
gabbifish: do we need to clone this twice? (once in line 49, once here?) I feel like we can put variables like client, preview_id, and server_config in the arguments section `move |args|` and avoid unnecessary clones :)

src/commands/dev/mod.rs
```diff
@@ -65,6 +74,22 @@ pub async fn dev(
     Ok(())
 }
 
+fn munge_response_headers(parts: &mut Parts) -> Result<(), failure::Error> {
```
gabbifish: nit: could we use a term more specific than munge, like `strip_cf_prefix_from_headers`?

src/commands/dev/mod.rs
```diff
@@ -89,7 +114,16 @@ fn preview_request(
     let now: DateTime<Local> = Local::now();
     let preview_id = &preview_id;
 
-    parts.uri = get_preview_url(&path).expect("Could not get preview url");
+    let mut headers: HeaderMap = HeaderMap::new();
```
gabbifish: maybe factor this header prefix addition logic out into another function that acts as an "opposite" function to `munge_response_headers`?

PR #886, opened 2019-11-20
[dev] Handle response headers

### LINKED ISSUE #872: [dev] strip cf-ew headers on response
The preview service returns all headers returned from the worker prefixed with `cf-ew-` - these need to be stripped, and the actual other headers returned from the preview service need to be removed.

EverlastingBugstopper:
Fixes #872 

This is now a pretty usable dev server! There's still a bunch of stuff to do but this is what I'd considered an MVP.

---

**worker code:**

```js
})

  return new Response("hello world", { headers: {"x-test": "header"}})
}
```

---

**Proxy and curl running in parallel**

```console
$ curl --head localhost:8000
HTTP/1.1 200 OK
content-length: 11
content-type: text/plain;charset=UTF-8
x-test: header
date: Wed, 20 Nov 2019 17:33:25 GMT
```

```console
$ wrangler dev
Listening on http://localhost:8000
[2019-11-20 11:33:25] "HEAD example.com/ HTTP/1.1"
```


src/commands/dev/mod.rs
```diff
@@ -46,12 +47,32 @@ pub async fn dev(
         let server_config = server_config.clone();
         async move {
             Ok::<_, failure::Error>(service_fn(move |req| {
```
ashleymichal: i think this closure deserves to either be a well named function in its own right, or to have a comment above.

src/commands/dev/mod.rs
```diff
@@ -21,6 +21,7 @@ use crate::settings::target::Target;
 use crate::terminal::emoji;
 
 const PREVIEW_HOST: &str = "rawhttp.cloudflareworkers.com";
+const HEADER_PREFIX: &str = "cf-ew-raw-";
```
ashleymichal: v. nice 👏

src/commands/dev/mod.rs
```diff
@@ -46,12 +47,32 @@ pub async fn dev(
...
-                    req,
-                    client.to_owned(),
-                    preview_id.to_owned(),
-                    server_config.clone(),
-                )
+                let client = client.to_owned();
+                let preview_id = preview_id.to_owned();
+                let server_config = server_config.clone();
+                async move {
+                    let resp = preview_request(req, client, preview_id, server_config).await?;
```
gabbifish: woooah! making moves~!!!! (to async/await)

src/commands/dev/mod.rs
```diff
@@ -46,12 +47,32 @@ pub async fn dev(
...
+                    let mut headers = HeaderMap::new();
+
+                    for header in &parts.headers {
+                        let (name, value) = header;
+                        let name = name.as_str();
+                        if name.starts_with(HEADER_PREFIX) {
+                            let header_name = &name[HEADER_PREFIX.len()..];
+                            // TODO: remove unwrap
+                            let header_name =
+                                HeaderName::from_bytes(header_name.as_bytes()).unwrap();
```
gabbifish: I'd also remove the unwrap in this PR and just wrap whatever the error message is in `failure::Error` (e.g. call `failure::bail!()`)

src/commands/dev/mod.rs
EverlastingBugstopper: i don't think i can do that since hyper doesn't re-export the type `MakeServiceFn` from what I can tell. i've simplified the closure as much as i can by pulling some if it out into a separate function though

EverlastingBugstopper:
@gabbifish @ashleymichal i think i've addressed concerns here :)

PR #893, opened 2019-11-20
Make azure use latest rustc

### LINKED ISSUE #887: Update Azure pipelines to use rust 1.39
None of the proxy branches are building because "async/await is unstable"

EverlastingBugstopper:
Builds are failing for dev server - run `rustup update stable` after curl installer

Fixes #887


EverlastingBugstopper:
yes - initially i was hard coding it but i've since changed my approach. edited description to reflect.

PR #914, opened 2019-11-26
Refactor: Use cloudflare-rs for calls to routes endpoints

ashleymichal:
**NOTE: this PR is blocked as it will not compile until https://github.com/cloudflare/cloudflare-rs/pull/69 is merged and published. To test locally, build cloudflare-rs from the branch alewis/fix#68 and point wrangler's cargo.toml at that `path`**

This PR includes a refactor to use cloudflare-rs as the client for routes calls.

Also some 💅 :
* use `log::info` directly
* pull methods out of Route struct impl into functions in anticipation of additional refactors.
* impl PartialEq for Route rather than defining `Route.match` function
* move std::prelude::* import in global_user to test mod (fmt warning)

### DIFF
--- a/src/commands/publish/mod.rs
+++ b/src/commands/publish/mod.rs
@@ -5,8 +5,6 @@ pub mod upload_form;
 
 pub use package::Package;
 
-use route::Route;
-
 use std::env;
 use std::path::Path;
 
@@ -109,8 +107,8 @@ fn publish_script(
     }
 
     let pattern = if target.route.is_some() {
-        let route = Route::new(&target)?;
-        Route::publish(&user, &target, &route)?;
+        let route = route::Route::new(&target)?;
+        route::publish(&user, &target, &route)?;
         log::info!("publishing to route");
         route.pattern
     } else {
--- a/src/commands/publish/mod.rs
+++ b/src/commands/publish/mod.rs
@@ -4,6 +4,7 @@ mod route;
 pub mod upload_form;
 
 pub use package::Package;
+use route::{publish_route, Route};
 
 use std::env;
 use std::path::Path;
@@ -107,8 +108,8 @@ fn publish_script(
     }
 
     let pattern = if target.route.is_some() {
-        let route = route::Route::new(&target)?;
-        route::publish(&user, &target, &route)?;
+        let route = Route::new(&target)?;
+        publish_route(&user, &target, &route)?;
         log::info!("publishing to route");
         route.pattern
     } else {
--- a/src/commands/publish/route.rs
+++ b/src/commands/publish/route.rs
@@ -5,8 +5,6 @@ use crate::terminal::emoji;
 use reqwest::header::CONTENT_TYPE;
 use serde::{Deserialize, Serialize};
 
-use log::info;
-
 #[derive(Deserialize, Serialize)]
 pub struct Route {
     script: Option<String>,
@@ -28,7 +26,10 @@ impl Route {
         {
             failure::bail!("You must provide a zone_id in your wrangler.toml before publishing!");
         }
-        let msg_config_error = format!("{} Your project config has an error, check your `wrangler.toml`: `route` must be provided.", emoji::WARN);
+        let msg_config_error = format!(
+            "{} Your project config has an error, check your `wrangler.toml`: `route` must be provided.", 
+            emoji::WARN
+        );
         Ok(Route {
             script: Some(target.name.to_string()),
             pattern: target.route.clone().expect(&msg_config_error),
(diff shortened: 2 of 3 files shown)


src/commands/publish/mod.rs
```diff
@@ -5,8 +5,6 @@ pub mod upload_form;
 
 pub use package::Package;
 
-use route::Route;
```
EverlastingBugstopper: can we keep this? `Route::new` feels better than `route::Route::new`

src/commands/publish/mod.rs
```diff
@@ -109,8 +107,8 @@ fn publish_script(
     }
 
     let pattern = if target.route.is_some() {
-        let route = Route::new(&target)?;
-        Route::publish(&user, &target, &route)?;
+        let route = route::Route::new(&target)?;
+        route::publish(&user, &target, &route)?;
```
EverlastingBugstopper: maybe out of scope for this but perhaps we should have this be `route.publish` instead of `route::publish` - since we already pass `&target` to `Route::new`, it seems strange to _also_ pass it to `route::publish`

src/commands/publish/route.rs
```diff
@@ -28,103 +34,71 @@ impl Route {
...
-    ) -> Result<(), failure::Error> {
-        if route.exists(user, target)? {
-            return Ok(());
-        }
-        create(user, target, route)
+pub fn publish(user: &GlobalUser, target: &Target, route: &Route) -> Result<(), failure::Error> {
+    if exists(user, target, route)? {
+        return Ok(());
     }
+    create(user, target, route)
```
EverlastingBugstopper: nit:

this could be 
```rust
if exists(user, target, route)? {
  Ok(())
} else {
  create(user, target, route)
}

File: src/commands/publish/route.rs [hunk 4]
```diff
@@ -28,103 +34,71 @@ impl Route {
...
+}
 
-    pub fn exists(&self, user: &GlobalUser, target: &Target) -> Result<bool, failure::Error> {
-        let routes = get_routes(user, target)?;
+fn exists(user: &GlobalUser, target: &Target, route: &Route) -> Result<bool, failure::Error> {
```
EverlastingBugstopper: nit: should probably name this `route_exists` or maybe put this on `Route` (maybe out of scope)

src/commands/publish/route.rs
```diff
@@ -28,103 +34,71 @@ impl Route {
...
+
+    log::info!("Creating your route {:#?}", &route.pattern,);
+    match client.request(&CreateRoute {
+        zone_identifier: &target.zone_id.as_ref().unwrap(),
+        params: CreateRouteParams {
+            pattern: route.pattern.clone(),
+            script: route.script.clone(),
+        },
+    }) {
+        Ok(_success) => Ok(()),
```
EverlastingBugstopper: nit: `_success` -> `_`

src/commands/publish/route.rs
```diff
@@ -28,103 +34,71 @@ impl Route {
...
-        return Ok(format!(
-            "https://api.cloudflare.com/client/v4/zones/{}/workers/routes",
-            zone_id
-        ));
+fn routes_error_help(error_code: u16) -> &'static str {
+    match error_code {
+        10020 => r#"
+            A worker with a different name was previously deployed to the specified route.
+            If you would like to overwrite that worker,
+            you will need to change `name` in your `wrangler.toml` to match the currently deployed worker,
```
EverlastingBugstopper: Also out of scope for this - I wonder if it's possible to return the name of the worker that's currently associated with that route.

src/commands/publish/mod.rs
ashleymichal: the constructor is likely to get hosed in the next PR, tbh.

src/commands/publish/mod.rs
ashleymichal: in the next PR i am introducing a method to `target`, `routes` that returns a vec of all routes that we will iterate over, so this function will no longer take a route as a parameter, and the Route struct will become a function-less data structure.

src/commands/publish/route.rs
ashleymichal: i took it off of route. i don't think this will stay around very long.

src/commands/publish/route.rs
ashleymichal: i actually have a ticket open to the API with that request, because as it stands today, no it is not possible without making another get request. I will likely be optimizing these calls to make a single get request and keep that set around, in which case we can totally do something like this, but i am also thinking of trying to modify the api to return the appropriate error message.

src/commands/publish/route.rs
ashleymichal: renamed route_exists

src/commands/publish/mod.rs
ashleymichal: renamed this "publish_route"

src/commands/publish/mod.rs
ashleymichal: restored 'use' statement for route module functions to remove this stutter.

ashleymichal:
merging into long-running feature branch, which will remain blocked on cloudflare-rs changes.

PR #916, opened 2019-11-27
Feature: Multi-route support

### LINKED ISSUE #866: RFC: Multi-Route support
# MultiRoute Support
## Progress so far
Lots of great thought put into this PR: https://github.com/cloudflare/wrangler/pull/224, as well as some great points about problems mapping routes to Wrangler projects outlined below.

### `null` Workers
The biggest hurdle we have to handle in Wrangler is “off” routes; routes that are set to a “null” or “no-op” worker. This is tricky because:
1. routes are set at the zone-level, but they can point to any Worker published to the account level. A “many-to-many” relationship.
2. Wrangler projects are scoped to a single Worker which can be deployed to many routes across many zones, including workers.dev. This is a one-to-many relationship. But when defining routes, we need to be able to deal with those “null” workers, which are not represented as a Worker in Wrangler, but rather just an “off” state.

You can’t just compose all the routes for the project, either, because it is likely that there are other scripts on the zone that are not represented in the toml.

##### The Cloudflare Dashboard
The Cloudflare Zoned Dashboard UI actually presents a user with _all_ of the routes configured on their zone, which includes potentially many workers, including routes pointing at a `null` Worker. This approach would be awkward in the context of Wrangler, which is generally concerned with a single worker script, not all scripts associated with an account.

##### Cloudflare Serverless Plugin
The Serverless Plugin doesn't currently concern itself with deleting old routes on deploy; however a later addition to the underlying tooling library includes a function for removing a route, so I imagine some subsequent steps used by the developing team to handle removing routes.
**I believe that implementing this much is sufficient for a first pass at this feature; it will unblock loads of people.**

### PHASE 1: PUT all routes in toml (Parity with Serverless plugin)
This will unblock most users. The downside is it potentially leaves behind unwanted routes, but the `route` option does that today, too. This can support both `on` and `off` routes, with the same downsides. Users will need to handle conflicts in the dashboard (we can provide a quick link by using their zone id to fill in the appropriate url).

### PHASE 1.5: Sync "on" routes (one step above parity with Serverless plugin)
We can be reasonably certain that if a user publishes a script, any routes associated with that script name that are not in their toml can be removed.

### PHASE 2: Guide users through handling conflicts
Any route specified in a user's toml, but associated with a different script, even a null worker, might cause unnecessary problems for users with complex routing tables. There are a couple of ways to provide users with the appropriate feedback:

#### `wrangler publish --dry-run`
**NOTE**: This is a straw man. The wording and presentation of this information is fully up for debate, but I'd like to focus on the concept, namely that by querying the api for the list of routes, Wrangler can provide the user with the information they need to make decisions about their deployments.

Passing a `--dry-run` flag to `wrangler publish` could do the work of diffing the existing routes against the specified routes, outputting a formatted list of anticipated modifications and conflicts:

For example, let's say we are deploying a worker called `cache` to a zone with the following routes already deployed:

```console
"https://vsnginx.workers-tooling.cf":                     "cache"
"https://vsnginx.workers-tooling.cf/cache":               "cache"
"https://workers-tooling.cf/snippets/cache":              "cache"
"https://workers-tooling.cf/snippets/cache*":             "cache"
"qr.workers-tooling.cf/":                                 "qr-generator"
"*serve-assets.workers-tooling.cf/*":                     "serve-assets"
"workers-tooling.cf/foo/*":                               "hello-world-dev-foo"
```

Let's say I define the following routes in my `wrangler.toml` (for the purpose of this illustration, I will ignore the `on/off` concept and just include a list of routes):

```toml
routes = [
    "https://vsnginx.workers-tooling.cf",
    "https://workers-tooling.cf/snippets/cache",
    "https://workers-tooling.cf/snippets/cache*",
    "qr.workers-tooling.cf/",
    "*serve-assets.workers-tooling.cf/*",
    "vsnginx.workers-tooling.cf/*",
    "workers-tooling.cf/foo/*"
]
```

Running `wrangler publish --dry-run` should output the following table illustrating how this will affect the user's routing table:

```console
M "https://vsnginx.workers-tooling.cf":                     "cache"
D "https://vsnginx.workers-tooling.cf/cache":               "cache"
M "https://workers-tooling.cf/snippets/cache":              "cache"
M "https://workers-tooling.cf/snippets/cache*":             "cache"
C "qr.workers-tooling.cf/":                                 "qr-generator"
C "*serve-assets.workers-tooling.cf/*":                     "serve-assets"
A "vsnginx.workers-tooling.cf/*":                           "cache"
C "workers-tooling.cf/foo/*":                               "hello-world-dev-foo"

`M` = `maintained` - this route already points to this script
`D` = `deleted` - this route points to this script but is not in the toml and so will be deleted
`C` = `conflicts` - this route currently points to a different script and may break it
`A` = `added` - this route does not currently exist and will be added
```

This shows me that I have three routes that would potentially break other workers running on my zone (those marked `C`). Which brings us to...

#### `wrangler publish --overwrite-routes`
On a normal `wrangler publish`, wrangler should make the same check against the routes list as in the dry run, and throw an error if any `conflicts` exist, printing the offending route patterns to the console. This gives the user the option to either update their `routes` list, or run wrangler publish again, passing an `--overwrite-routes` flag that will point the conflicting flags to the project script; effectively this is the user saying "I know those routes exist, go ahead and update them anyway".

### PHASE 3: Route subcommands
We can optionally give the user the ability to handle routes individually by providing utilities for CRUD operations on routes, as well as listing routes. I would defer this in favor of the flag until it is requested, but it does give users more granular control over their routing tables.

#### What about those pesky "null" workers?
For a phase 1, we can simply add "off" routes the same way we would "on" routes; that is, just make the same kind of PUT request and let the cards land where they will. I believe this will fail if the route is already pointing to another worker, including the current worker. We could nudge this in the direction of being okay with changing a route from being "on" for this worker to being "off", but not the other way around.

For phase 2, we can treat null workers very similarly to just another conflicting worker:

| v deployed \| in toml >          | Worker "on"   | Worker "off"  |
| -------------------------------- | ------------- | ------------- |
| **route points to this worker**  | Stay the same | Update        |
| **route points to null worker**  | Conflict      | Stay the same |
| **route points to other worker** | Conflict      | Conflict      |

Where a `conflict` will cause wrangler publish to bail unless passed the `--overwrite-routes` flag.

### All or nothing vs one-at-a-time
The above solution assumes that it is better to fail the entire deploy than to deploy only to some routes, which requires that we query the API for the list of routes before taking action. I'm not stoked about this; I would rather build this into the API, but at the moment we are just trying to unblock folks and see if this kind of an approach solves their problems. If we think it's acceptable to just try every route and return the errors to the user, that is still an option (this is how Serverless works), and the fact is that under the hood we will still have to individually POST or PUT every specified route in series rather than defining them in bulk; however defining the interface as a bulk put now makes transitioning the implementation smoother in the future.

### How far do we take this?
I've proposed three phases to this work; the first phase gives us parity with the Serverless Plugin, but without the level of support of the Cloudflare Dashboard UI. This would unblock many users, but may cause problems with more complex multiroute use cases.

The second phase offers what I consider to be the minimum of protection against unexpected behavior. I am happy to write a proof of concept of this solution, but I think it will take more development and is perhaps best suited to a later revision.

The third phase gives routes first class status in wrangler. This one feels like overkill at the moment.

### Wrangler environments
The `route` key in the wrangler.toml cannot be inherited by other environments, since inheriting would stomp the original deployment. `routes` should likewise not be inherited, but be required as an explicit configuration.

ashleymichal:
Addresses phase one of #866 

This PR seeks to unblock users who require Wrangler to publish to multiple routes on a Cloudflare zone. It supports:

* Specifying a list of route patterns in wrangler.toml for publish
* a command to list the routes active on a given zone
* a command to delete a route on a given zone

It does *not* support:

* declaring "null worker" routes
* "syncing" routing tables for a project

## Testing this branch

see https://gist.github.com/ashleymichal/b1a3f0d881fed2544b17c3e7ee1ecc17 for testing instructions.


src/commands/publish/route.rs
```diff
@@ -1,129 +1,124 @@
...
-    script: Option<String>,
-    pub pattern: String,
+use crate::http::{api_client, format_error};
+use crate::settings::global_user::GlobalUser;
+use crate::settings::toml::Route;
+
+pub fn publish_routes(
+    user: &GlobalUser,
+    routes: Vec<Route>,
+    zone_id: &String,
```
EverlastingBugstopper: I think we should be accepting an &str here https://stackoverflow.com/questions/40006219/why-is-it-discouraged-to-accept-a-reference-to-a-string-string-vec-vec-o

src/commands/publish/route.rs
```diff
@@ -1,129 +1,124 @@
...
+        .collect();
+
+    Ok(deployed_routes)
 }
 
-#[derive(Deserialize)]
-struct RoutesResponse {
-    result: Vec<Route>,
-}
+fn fetch_all(user: &GlobalUser, zone_identifier: &String) -> Result<Vec<Route>, failure::Error> {
```
EverlastingBugstopper: same &str comment

src/commands/publish/route.rs
```diff
@@ -1,129 +1,124 @@
...
-        Ok(false)
-    }
+    Ok(routes)
+}
 
-    pub fn matches(&self, route: &Route) -> bool {
-        self.pattern == route.pattern && self.script == route.script
+fn create(
+    user: &GlobalUser,
+    zone_identifier: &String,
```
EverlastingBugstopper: same &str comment

src/commands/publish/route.rs
```diff
@@ -1,129 +1,124 @@
...
+    Same(Route),
+    Conflict(Route),
+    New(Route),
+    Error((Route, String)),
+}
 
-        failure::bail!(msg)
+fn deploy_route(
+    user: &GlobalUser,
+    zone_id: &String,
```
EverlastingBugstopper: same &str comment

src/commands/publish/route.rs
```diff
@@ -1,129 +1,124 @@
...
+    route: &Route,
+    existing_routes: &Vec<Route>,
```
EverlastingBugstopper: if possible - should accept &[Route] for same reason as &str

src/commands/publish/route.rs
ashleymichal: this is a wip.

ashleymichal:
this is only marked `WIP` because we are waiting to release 1.7.0 before merge

EverlastingBugstopper:
can we mark it as blocked instead of `WIP` if it is ready for review?

EverlastingBugstopper:
The tests seem to include a lot of compiler warnings of "this function is never used" which we need to address before merging

ashleymichal:
> The tests seem to include a lot of compiler warnings of "this function is never used" which we ~need to~ really should address before merging

these are a result of the test fixture code being shared between the tests/ directory and the src tests. we can move all the deploy target tests to tests/ though, that might be better.

EverlastingBugstopper:
👍

AaronO:
@ashleygwilliams @EverlastingBugstopper Wanted to try the `1.8.0-rc.0` release, however the install fails (at least on macOS) with a `TAR_BAD_ARCHIVE` exception.

Checking the URL it fetches (https://workers.cloudflare.com/get-npm-wrangler-binary/1.8.0-rc.0/x86_64-apple-darwin), I realized that it's actually a `.zip` file (see the `PK` magic-number):

```
❯ curl https://workers.cloudflare.com/get-npm-wrangler-binary/1.8.0-rc.0/x86_64-apple-darwin 2>/dev/null | head -c2 | hexdump -C
00000000  50 4b                                             |PK|
00000002
```

Anychance you could re-upload the build artifacts as `.tar.gz` or make a corrective `1.8.0-rc.1` release ?

ashleygwilliams:
@AaronO yup! we noticed that as we did the release at the end of the day- we're hoping to fix it this morning :)

lawred2:
> @ashleygwilliams @EverlastingBugstopper I wanted to try the `1.8.0-rc.0` release, however the install fails (at least on macOS) with a `TAR_BAD_ARCHIVE` exception.
> 
> Checking the URL it fetches (https://workers.cloudflare.com/get-npm-wrangler-binary/1.8.0-rc.0/x86_64-apple-darwin), I realized that it's actually a `.zip` file (see the `PK` magic-number):
> 
> ```
> ❯ curl https://workers.cloudflare.com/get-npm-wrangler-binary/1.8.0-rc.0/x86_64-apple-darwin 2>/dev/null | head -c2 | hexdump -C
> 00000000  50 4b                                             |PK|
> 00000002
> ```
> 
> Anychance you could re-upload the build artifacts as `.tar.gz` or make a corrective `1.8.0-rc.1` release ?

fwiw - seeing the same on windows

EverlastingBugstopper:
Hey @AaronO and @lawred2 this should be fixed now 😄 We switched our CI provider which caused some issues for us (which is why we do beta releases 👍 )

PR #917, opened 2019-11-27
[dev] Module refactor

EverlastingBugstopper:
Thanks to @ashleymichal's review, the module structure should be much more readable.


src/commands/dev/server_config/host.rs
```diff
@@ -0,0 +1,37 @@
...
+#[derive(Clone)]
+pub struct Host {
+    url: Url,
+}
+
+impl Host {
+    pub fn new(host: &str) -> Result<Self, failure::Error> {
+        let url = match Url::parse(&host) {
+            Ok(host) => Ok(host),
+            Err(_) => Url::parse(&format!("https://{}", host)),
```
ashleymichal: i'm still not 100% clear on this v. specific catch.

PR #918, opened 2019-11-27
Alewis/refactor routes

ashleymichal:
More intermediate refactors as we move toward multiroute support.

* Move Route struct into settings/target
* Make the routes key in Target a Vec<String> rather than a HashMap
* add a 'routes' method to Target, which will handle both `route` setting and `routes` setting, asserting that only one can be specified in the wrangler.toml, and generalizing them to a Vec<Route>

TODO:

* memoize/cache calls to get all routes; I would like to only refresh these on 10020 errors when we try to create a route, rather than every time we try to upload a specific route
* add tests (this PR)

### DIFF
--- a/src/settings/target/manifest.rs
+++ b/src/settings/target/manifest.rs
@@ -31,7 +31,7 @@ pub struct Manifest {
     pub workers_dev: Option<bool>,
     #[serde(default = "some_string")]
     pub route: Option<String>,
-    pub routes: Option<HashMap<String, String>>,
+    pub routes: Option<Vec<String>>,
     #[serde(default = "some_string")]
     pub zone_id: Option<String>,
     pub webpack_config: Option<String>,
--- a/src/settings/target/target.rs
+++ b/src/settings/target/target.rs
@@ -3,7 +3,6 @@ use super::kv_namespace::KvNamespace;
 use super::site::Site;
 use super::target_type::TargetType;
 
-use std::collections::HashMap;
 use std::env;
 
 use std::path::PathBuf;
@@ -19,7 +18,7 @@ pub struct Target {
     #[serde(rename = "type")]
     pub target_type: TargetType,
     pub route: Option<String>,
-    pub routes: Option<HashMap<String, String>>,
+    pub routes: Option<Vec<String>>,
     pub webpack_config: Option<String>,
     pub zone_id: Option<String>,
     pub site: Option<Site>,
--- a/src/settings/target/target.rs
+++ b/src/settings/target/target.rs
@@ -50,11 +50,14 @@ impl Target {
     pub fn routes(&self) -> Result<Vec<Route>, failure::Error> {
         let mut routes = Vec::new();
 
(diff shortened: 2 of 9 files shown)


src/settings/target/manifest.rs
```diff
@@ -31,7 +31,7 @@ pub struct Manifest {
     pub workers_dev: Option<bool>,
     #[serde(default = "some_string")]
     pub route: Option<String>,
-    pub routes: Option<HashMap<String, String>>,
```
EverlastingBugstopper: interested in why this was a hashmap to begin with - i know we never used it anywhere but still interested

ashleymichal: because technically routes are a map between a pattern and a script, and it is potentially useful within a Wrangler project if you wanted to "negate" a route by setting it to a `null` worker. For the purposes of this refactor, I am starting with the simplest implementation - a list of patterns - to reach parity with serverless.

src/settings/target/target.rs
```diff
@@ -47,4 +46,28 @@ impl Target {
             }
         }
     }
+
+    pub fn routes(&self) -> Result<Vec<Route>, failure::Error> {
+        let mut routes = Vec::new();
+
+        if let Some(single_route) = &self.route {
+            if self.routes.is_some() {
```
EverlastingBugstopper: this case isn't covered in the else and should probably be brought out to the top - 

```rust
if &self.route.is_some() && &self.routes.is_some() {
  failure::bail!("You can specify EITHER `route` or `routes` in your wrangler.toml")
}
```

(also it shouldn't have a trailing semi-colon i think)

src/settings/target/manifest.rs
EverlastingBugstopper: makes sense - will this bite us at all when we try to migrate? will the future require breaking changes to the toml? would like to avoid that as much as possible

src/settings/target/target.rs
ashleymichal: what do you mean "this case isn't covered in the else"?

EverlastingBugstopper: in the `else if let Some(multi_route)` block, we do not check to see if there is also a `route` field. I see now that if they were both there, the first block would catch it, but I still find it a bit strange that we check if both exist after we check if one exists. I think I'd have it as its own separate block like I describe above even though they would be functionally equivalent.

ashleymichal: yeah covering it in both places would be redundant. i'll pull up the first case.

src/settings/target/manifest.rs
ashleymichal: predicting the future of routes is about as easy as predicting next year's weather (probably will be hotter than today, but who knows if it'll be raining). That being said, if it is clear before release that we need this data structure to change we can do that.

ashleymichal:
sob. all my tests are borked. cloudflare-rs should be published sometime today i think

PR #922, opened 2019-12-03
Alewis/publish multiple routes

ashleymichal:
this pr gets us to where we actually publish all specified routes within a wrangler toml.

 todo:
 unit tests
at least one integration test
improve output to show how routes were updated/maintained.

### DIFF
--- a/src/commands/publish/mod.rs
+++ b/src/commands/publish/mod.rs
@@ -16,7 +16,7 @@ use crate::commands::subdomain::Subdomain;
 use crate::commands::validate_worker_name;
 use crate::http;
 use crate::settings::global_user::GlobalUser;
-use crate::settings::target::{KvNamespace, Route, Site, Target};
+use crate::settings::target::{KvNamespace, Site, Target};
 use crate::terminal::{emoji, message};
 
 pub fn publish(
@@ -132,8 +132,8 @@ fn deploy(user: &GlobalUser, target: &Target) -> Result<(), failure::Error> {
             let published_routes = publish_routes(&user, routes, zone_id)?;
 
             message::success(&format!(
-                "Successfully published your script to:\n{}",
-                format_routes(published_routes)
+                "Deployed to the following routes:\n{}",
+                serde_json::to_string(&published_routes)?
             ));
 
             Ok(())
@@ -143,13 +143,6 @@ fn deploy(user: &GlobalUser, target: &Target) -> Result<(), failure::Error> {
             )
         }
     }
-}
-
-// TODO: provide a little more information: created, retained, conflicted, errored
-fn format_routes(routes: Vec<Route>) -> String {
-    let formatted_routes: Vec<String> = routes.iter().map(|r| r.pattern.clone()).collect();
-
-    formatted_routes.join("\n")
 }
 
 fn error_msg(status: reqwest::StatusCode, text: String) -> String {
--- a/src/commands/publish/mod.rs
+++ b/src/commands/publish/mod.rs
@@ -36,7 +36,9 @@ pub fn publish(
     // Build the script before uploading.
     commands::build(&target)?;
 
-    publish_script(&user, &target, asset_manifest)?;
+    upload_script(&user, &target, asset_manifest)?;
+
+    deploy(&user, &target)?;
 
     Ok(())
 }
@@ -65,7 +67,7 @@ pub fn bind_static_site_contents(
     Ok(())
 }
 
-fn publish_script(
+fn upload_script(
     user: &GlobalUser,
     target: &Target,
     asset_manifest: Option<AssetManifest>,
@@ -95,6 +97,10 @@ fn publish_script(
         failure::bail!(error_msg(res_status, res_text))
     }
 
+    Ok(())
+}
+
+fn deploy(user: &GlobalUser, target: &Target) -> Result<(), failure::Error> {
     let routes = target.routes()?;
     log::info!("routes: {:#?}", &routes);
 
--- a/src/commands/publish/mod.rs
+++ b/src/commands/publish/mod.rs
@@ -4,7 +4,7 @@ mod route;
 pub mod upload_form;
 
 pub use package::Package;
-use route::publish_route;
+use route::publish_routes;
 
 use std::env;
 use std::path::Path;
@@ -16,7 +16,7 @@ use crate::commands::subdomain::Subdomain;
 use crate::commands::validate_worker_name;
 use crate::http;
 use crate::settings::global_user::GlobalUser;
-use crate::settings::target::{KvNamespace, Site, Target};
+use crate::settings::target::{KvNamespace, Route, Site, Target};
 use crate::terminal::{emoji, message};
 
 pub fn publish(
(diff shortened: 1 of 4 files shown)


src/commands/publish/mod.rs
```diff
@@ -107,21 +95,53 @@ fn publish_script(
...
+                message::warn(&format!("The route in your wrangler.toml should have a trailing * to apply the Worker on every path, otherwise your site will not behave as expected.\nroute = {}*", route_pattern));
+            }
+        }
+
+        if let Some(zone_id) = &target.zone_id {
+            let published_routes = publish_routes(&user, routes, zone_id)?;
+
+            message::success(&format!(
+                "Successfully published your script to:\n{}",
+                format_routes(published_routes)
```
EverlastingBugstopper: nit and possibly out of scope - wonder if it's possible to implement this as `Display` on a `Routes` type

src/commands/publish/mod.rs
```diff
@@ -107,21 +95,53 @@ fn publish_script(
...
+            ));
+
+            Ok(())
+        } else {
+            failure::bail!("You must provide a zone_id in your wrangler.toml before publishing!")
```
EverlastingBugstopper:
```suggestion
            failure::bail!("You must provide a zone_id in your wrangler.toml before publishing to a route!")
```

Maybe worth putting something in here about the workers.dev alternative as well - not sure

src/commands/publish/mod.rs
```diff
@@ -268,7 +269,7 @@ fn validate_target_required_fields_present(target: &Target) -> Result<(), failur
         {
             missing_fields.push("zone_id")
         };
-        if target.route.as_ref().unwrap_or(&"".to_string()).is_empty() {
+        if route.is_empty() {
```
EverlastingBugstopper: should probably check if `.routes()` is empty, no? This is part of the reason I'm wondering if we should remove `route` from `target` and replace it with `routes`. But yeah if we still want the method we should use it here

src/commands/publish/route.rs
```diff
@@ -4,48 +4,62 @@ use cloudflare::framework::HttpApiClientConfig;
...
+    Ok(deployed_routes)
 }
 
-fn fetch_all(user: &GlobalUser, target: &Target) -> Result<Vec<Route>, failure::Error> {
+fn fetch_all(user: &GlobalUser, zone_identifier: &String) -> Result<Vec<Route>, failure::Error> {
     let client = api_client(user, HttpApiClientConfig::default())?;
 
     let routes: Vec<Route> = match client.request(&ListRoutes {
-        zone_identifier: &target.zone_id.as_ref().unwrap(),
+        zone_identifier: zone_identifier,
```
EverlastingBugstopper:
```suggestion
        zone_identifier,
```

src/commands/publish/route.rs
```diff
@@ -4,48 +4,62 @@ use cloudflare::framework::HttpApiClientConfig;
...
 
+    // TODO: include id info in Route objects
     Ok(routes)
 }
 
-fn create(user: &GlobalUser, target: &Target, route: &Route) -> Result<(), failure::Error> {
+// TODO: merge id info into returned Route object
+fn create(
+    user: &GlobalUser,
+    zone_identifier: &String,
```
EverlastingBugstopper: this is interesting to me! why an `&String` and not an `&str` - and how do you even pass an `&String`?

src/commands/publish/mod.rs
ashleymichal: so i have more commits incoming, and in them this placeholder function goes away and i just serialize to json directly here. do you think i should push those as a separate PR or add them to this?

src/commands/publish/mod.rs
ashleymichal: i think i may be addressing more of the workers.dev stuff in subsequent work; as it is this is just the same message we had before but idk why i changed the wording.

src/commands/publish/mod.rs
ashleymichal: yes, i'll be coming through and cleaning up references to `target.route` in a later PR. good catch!

src/commands/publish/route.rs
ashleymichal: i mean, is the difference between a String and a slice. the String comes from the field on the struct, and i need to use it in two places so i borrow it. 🤷‍♀

src/commands/publish/mod.rs
```diff
@@ -24,22 +24,10 @@ pub fn publish(
...
-
     validate_target_required_fields_present(target)?;
     validate_worker_name(&target.name)?;
 
     if let Some(site_config) = target.site.clone() {
-        if let Some(route) = &target.route {
-            if !route.ends_with("*") {
-                message::warn(&format!("The route in your wrangler.toml should have a trailing * to apply the Worker on every path, otherwise your site will not behave as expected.\nroute = {}*", route));
-            }
-        }
```
ashleymichal: i moved this down into the deploy routes function, since it's only relevant to that path

src/commands/publish/mod.rs
```diff
@@ -107,21 +97,50 @@ fn publish_script(
...
-        publish_route(&user, &target)?
-    } else {
+    Ok(())
+}
+
+fn deploy(user: &GlobalUser, target: &Target) -> Result<(), failure::Error> {
+    let routes = target.routes()?;
+    log::info!("routes: {:#?}", &routes);
+
+    if routes.is_empty() {
```
ashleymichal: i'm nearly inclined to move this all just into the top level command function.

src/commands/publish/mod.rs
```diff
@@ -132,25 +151,6 @@ fn error_msg(status: reqwest::StatusCode, text: String) -> String {
     }
 }
 
-#[test]
```
ashleymichal: nothing about this test changed, i just moved it to the bottom of the module.

src/commands/publish/mod.rs
EverlastingBugstopper: i think i'd rather it be its own function than in the top level

src/commands/publish/mod.rs
```diff
@@ -258,7 +258,7 @@ fn validate_target_required_fields_present(target: &Target) -> Result<(), failur
         None => {}
     }
 
-    let destination = if target.route.is_some() {
+    let destination = if let Some(route) = &target.route {
```
ashleymichal: i could fix this to check `if target.routes()?.is_empty()`, however i'm also inclined to shift some of this logic into something like a `Manifest::get_deploy_config` function instead, so we can keep all the logic around what is needed to publish in one place; right now it is spread across this module and `settings::target`

src/commands/publish/mod.rs
ashleymichal: this would also be more like `for route in routes ...`

src/commands/publish/mod.rs
ashleymichal: i guess what i meant was, all the std out output and switching on routes etc. but i also think i might futz with this so i'm into it staying here

PR #923, opened 2019-12-03
[dev] Output console.log messages

### LINKED ISSUE #875: [dev] Continue implementing devtools protocol work
The preview service exposes a websocket according to [this protocol](https://chromedevtools.github.io/devtools-protocol/)

A good start on this has been made in #923 and future work on implementing types for the protocol is tracked at https://github.com/everlastingbugstopper/chrome-devtools-rs

EverlastingBugstopper:
Starts to fix #875 

Log messages work for everything that is in `chrome-devtools-rs`

```console
$ wrangler dev
👂  Listening on http://localhost:8000
[2019-12-03 13:26:46] "GET example.com/ HTTP/1.1"
"hi there"
```

work on this is being done at https://github.com/everlastingbugstopper/chrome-devtools-rs

If you'd like to help develop, there are some open issues. I'd recommend checking out this branch and cloning that repository. Make sure to switch the dependency out to your local version in wrangler's `Cargo.toml`.

The way I've been developing this is to make a script console.log some stuff and output the `impl Display` to the terminal. the raw JSON coming from the devtools protocol will be printed on a `RUST_LOG=info` along with any serialization errors (this can be a bit noisy as we're not interested in all messages that come from the protocol).

Anyway - there are some naive cases in the library, look for match statements with `_` in them, and try running console.log on some of the object types that are in the library's issue tracker and pull them out of the `_` into another arm of the match statement. I'd recommend looking at both the raw json from RUST_LOG=info and the output that cloudflareworkers.com gives for that object type and trying to match them up. 

Branch off of https://github.com/EverlastingBugstopper/chrome-devtools-rs/tree/avery/object-preview-entries as it has the latest work and we don't want to reimplement it! I think the structs should all be there at this point so it's just a matter of doing the display impls.


Cargo.toml
```diff
@@ -38,7 +38,6 @@ text_io = "0.1.7"
 exitfailure = "0.5.1"
 prettytable-rs = "0.8.0"
 notify = "4.0.12"
-ws = "0.9.0"
```
ashleymichal: we drop this entirely?

src/commands/dev/mod.rs
```diff
@@ -1,19 +1,29 @@
 mod server_config;
+mod socket;
 use server_config::ServerConfig;
 mod headers;
 use headers::{prepend_request_headers_prefix, strip_response_headers_prefix};
 
+extern crate openssl;
+extern crate tokio;
+extern crate url;
+extern crate ws;
```
ashleymichal: curious why these are extern crate rather than use statements?

src/commands/dev/mod.rs
```diff
@@ -1,19 +1,29 @@
...
 use hyper::header::{HeaderName, HeaderValue};
 
 use hyper::http::uri::InvalidUri;
 use hyper::service::{make_service_fn, service_fn};
-use hyper::{Body, Client, Request, Response, Server, Uri};
+use hyper::{Body, Client as HyperClient, Request, Response, Server, Uri};
 
 use hyper_tls::HttpsConnector;
 
+use tokio::runtime::Runtime;
```
ashleymichal: nit: maybe `as TokioRuntime`? not that we have a boatload of runtimes running around, but is a pretty vague constructor name.

src/commands/dev/mod.rs
```diff
@@ -35,14 +45,23 @@ pub async fn dev(
 ) -> Result<(), failure::Error> {
     commands::build(&target)?;
     let server_config = ServerConfig::new(host, ip, port)?;
+    let session_id = get_session_id()?;
+    let preview_id = get_preview_id(target, user, &server_config, &session_id)?;
 
+    thread::spawn(move || socket::listen(session_id));
+
+    let mut rt = Runtime::new()?;
+    rt.block_on(serve(server_config, preview_id))?;
```
ashleymichal: nit: this is probably from some boilerplate, but generally speaking i prefer we use full words for variable names. s/rt/runtime/g

src/commands/dev/mod.rs
```diff
@@ -88,7 +107,7 @@ fn get_path_as_str(uri: &Uri) -> String {
 
 fn preview_request(
     req: Request<Body>,
-    client: Client<HttpsConnector<HttpConnector>>,
+    client: HyperClient<HttpsConnector<HttpConnector>>,
```
ashleymichal: where is this called (she asked herself)

src/commands/dev/socket/client.rs
```diff
@@ -0,0 +1,84 @@
...
+                "id": 2,
+                "method": "Runtime.enable"
+            }"#,
+        );
+        let enable_debugger = WsMessage::text(
+            r#"{
+                "id": 3,
+                "method": "Debugger.enable"
+            }"#,
+        );
```
ashleymichal: they're pretty well contained now, but if we end up generating any more of these outside this function, i would suggest some structs w/ serializers rather than formatted strings.

src/commands/dev/socket/client.rs
```diff
@@ -0,0 +1,84 @@
...
+        let msg = msg.as_text()?;
+        log::info!("{}", msg);
+        let msg: Result<DevtoolsEvent, serde_json::Error> = serde_json::from_str(msg);
+        match msg {
+            Ok(msg) => {
+                match msg {
+                    DevtoolsEvent::ConsoleAPICalled(event) => match event.log_type.as_str() {
+                        "log" => println!("{}", style(event).blue()),
+                        "error" => eprintln!("{}", style(event).red()),
+                        _ => println!("unknown console event: {}", event),
```
ashleymichal: is this unreachable? are there other console event types? should the above strings be made into a deserializable enum?

Cargo.toml
EverlastingBugstopper: nah - i moved it lower

```[dependencies.ws]
version = "0.9.0"
features = ["ssl"]```

File: src/commands/dev/mod.rs (hunk 2 above)
Comment (EverlastingBugstopper) [reply src/commands/dev/mod.rs]: must have had too much coffee! removed - good catch :)

File: src/commands/dev/mod.rs (hunk 3 above)
Comment (EverlastingBugstopper) [reply src/commands/dev/mod.rs]: done

File: src/commands/dev/mod.rs (hunk 5 above)
Comment (EverlastingBugstopper) [reply src/commands/dev/mod.rs]: `client.request(req)` on line `145`

File: src/commands/dev/socket/client.rs (hunk 6 above)
Comment (EverlastingBugstopper) [reply src/commands/dev/socket/client.rs]: yes these will eventually be moved to the devtools crate

File: src/commands/dev/socket/client.rs (hunk 7 above)
Comment (EverlastingBugstopper) [reply src/commands/dev/socket/client.rs]: this is unfinished devtools response stuff

Comment (EverlastingBugstopper) [reply src/commands/dev/socket/client.rs]: and yes - i'm planning on making the log_types an enum in the devtools crate

File: src/commands/dev/socket/client.rs (hunk 6 above)
Comment (gabbifish) [reply src/commands/dev/socket/client.rs]: Agreed, these should be structs whose "method" strings are consts.

File: src/commands/dev/socket/client.rs (hunk 7 above)
Comment (gabbifish) [reply src/commands/dev/socket/client.rs]: do we have a ticket to track the completion of this devtools response work? We should if we want to merge this PR into your feature branch.

File: src/commands/dev/socket/client.rs [hunk 8]
```diff
@@ -0,0 +1,84 @@
...
+                    },
+                    DevtoolsEvent::ExceptionThrown(event) => {
+                        eprintln!("{}", style(event).bold().red())
+                    }
+                }
+                Ok(())
+            }
+            Err(e) => {
+                // this event was not parsed as a DevtoolsEvent
+                log::info!("{}", e);
```
gabbifish: I'd add extra text to this info log saying that the event was not parsed as a DevTools event. Even better, I'd make this logging level WARN. (`log::warn!`)

EverlastingBugstopper:
work is being tracked here @gabbifish: https://github.com/cloudflare/wrangler/issues/875

i don't think it makes sense to make that a log::warn because there are a lot of messages we don't serialize and are safe to ignore. once we flesh that crate out and we should be handling all message types, then i'd say we should make it a log::warn

gabbifish:
Ah, I see--we're still waiting on the complete devtools parsing implementation! In that case, can we make a `//TODO` to make that log a WARN when we know we've covered all possible deserializable devtools values?

PR #925, opened 2019-12-03
[WIP] Add logic to download cloudflared binary from equinox.io

### LINKED ISSUE #924: [wrangler tail logs] Introduce installer logic for cloudflared
We need to ensure that wrangler can download the cloudflared binary if necessary; this will require adding a new set of download logic specific to equinox.io, which is where we host our cloudflared binaries.

gabbifish:
Closes #924.

Right now, this PR does not work as expected; it looks the the binary-install crate does not play well with the .tgz downloads from equinox.io:
```
thread 'main' panicked at 'don't know how to extract https://bin.equinox.io/a/hLuezyJbR6D/cloudflared-2019.11.3-linux-amd64.tgz', /home/gabbi/.cargo/registry/src/github.com-1ecc6299db9ec823/binary-install-0.0.3-alpha/src/lib.rs:125:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2404,6 +2404,7 @@ dependencies = [
  "reqwest 0.9.22 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.101 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.41 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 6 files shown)


gabbifish:
Blocked because we are considering a migration to github for cloudflared binary releases instead of continuing to use equinox.io (which has unreliable uptime guarantees).

EverlastingBugstopper:
@gabbifish should we close this as won't do?

PR #927, opened 2019-12-04
Alewis/improve test fixtures

ashleymichal:
**Note: this PR was created in the context of implementing multiroute support, and so is under that milestone, however nothing here is specific to multiroute support, and so if everyone thinks it makes sense i can cherry pick these changes onto a new branch and pull against master, instead.**

It looks like there's a lot here, but it's pretty repetitive, and all of it is in service of improving the experience of writing tests! 🎉 

This PR introduces a WranglerToml struct and some convenience methods that make instantiating syntax-valid (but not necessarily wrangler-valid) tomls much simpler. The existing tests that use fixtures have been converted to show the difference.

I also noticed while exercising the tests that we were running into some concurrency issues caused by our use of the shared BUILD_LOCK mutex. I copied over the logic from wasm-pack's fixture implementation to put this mutex on each individual fixture instead of, seemingly, the whole process. This appears to have removed the status quo where, if one test failed, many others failed, as well as a weird issue where a test might fail due to an npm install error that I couldn't quite track down but smelled related. To see the difference, check out master, comment out a step in a test that might make it fail (such as writing a wrangler.toml or package.json), and observe that it tends to intermittently crash other tests. The fixture.lock method appears to have fixed that issue, decoupling our tests.

### DIFF
--- a/tests/build.rs
+++ b/tests/build.rs
@@ -5,30 +5,24 @@ pub mod fixture;
 
 use std::process::Command;
 use std::str;
-use std::sync::Mutex;
 
 use assert_cmd::prelude::*;
 use fixture::{Fixture, WranglerToml};
 
-lazy_static! {
-    static ref BUILD_LOCK: Mutex<u8> = Mutex::new(0);
-}
-
 #[test]
 fn it_builds_webpack() {
-    let fixture = Fixture::new("webpack");
+    let fixture = Fixture::new();
     fixture.scaffold_webpack();
 
     let wrangler_toml = WranglerToml::webpack_no_config("test-build-webpack");
     fixture.create_wrangler_toml(wrangler_toml);
 
     build_creates_assets(&fixture, vec!["script.js"]);
-    fixture.cleanup();
 }
 
 #[test]
 fn it_builds_with_webpack_single_js() {
-    let fixture = Fixture::new("webpack_simple_js");
+    let fixture = Fixture::new();
     fixture.create_file(
         "index.js",
         r#"
@@ -51,12 +45,11 @@ fn it_builds_with_webpack_single_js() {
     fixture.create_wrangler_toml(wrangler_toml);
 
     build_creates_assets(&fixture, vec!["script.js"]);
-    fixture.cleanup();
 }
 
 #[test]
 fn it_builds_with_webpack_function_config_js() {
-    let fixture = Fixture::new("webpack_function_config_js");
+    let fixture = Fixture::new();
     fixture.scaffold_webpack();
 
     fixture.create_file(
@@ -70,12 +63,11 @@ fn it_builds_with_webpack_function_config_js() {
     fixture.create_wrangler_toml(wrangler_toml);
 
     build_creates_assets(&fixture, vec!["script.js"]);
-    fixture.cleanup();
 }
 
 #[test]
 fn it_builds_with_webpack_promise_config_js() {
-    let fixture = Fixture::new("webpack_promise_config_js");
+    let fixture = Fixture::new();
     fixture.scaffold_webpack();
 
     fixture.create_file(
@@ -89,12 +81,11 @@ fn it_builds_with_webpack_promise_config_js() {
     fixture.create_wrangler_toml(wrangler_toml);
 
     build_creates_assets(&fixture, vec!["script.js"]);
-    fixture.cleanup();
 }
 
 #[test]
 fn it_builds_with_webpack_function_promise_config_js() {
-    let fixture = Fixture::new("webpack_function_promise_config_js");
+    let fixture = Fixture::new();
     fixture.scaffold_webpack();
 
     fixture.create_file(
@@ -108,12 +99,11 @@ fn it_builds_with_webpack_function_promise_config_js() {
     fixture.create_wrangler_toml(wrangler_toml);
 
     build_creates_assets(&fixture, vec!["script.js"]);
-    fixture.cleanup()
 }
 
 #[test]
 fn it_builds_with_webpack_specify_config() {
-    let fixture = Fixture::new("webpack_specify_config");
+    let fixture = Fixture::new();
     fixture.scaffold_webpack();
 
     fixture.create_file(
@@ -130,12 +120,11 @@ fn it_builds_with_webpack_specify_config() {
     fixture.create_wrangler_toml(wrangler_toml);
 
     build_creates_assets(&fixture, vec!["script.js"]);
-    fixture.cleanup();
 }
 
 #[test]
 fn it_builds_with_webpack_single_js_missing_package_main() {
-    let fixture = Fixture::new("webpack_single_js_missing_package_main");
+    let fixture = Fixture::new();
     fixture.create_empty_js();
 
     fixture.create_file(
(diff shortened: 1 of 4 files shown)


tests/build.rs
```diff
@@ -5,31 +5,24 @@ pub mod fixture;
...
 use std::process::Command;
 use std::str;
-use std::sync::Mutex;
 
 use assert_cmd::prelude::*;
-use fixture::Fixture;
-
-lazy_static! {
-    static ref BUILD_LOCK: Mutex<u8> = Mutex::new(0);
-}
```
ashleymichal: this is now implemented on the Fixture struct

tests/build.rs
```diff
@@ -5,31 +5,24 @@ pub mod fixture;
...
+use fixture::{Fixture, WranglerToml};
 
 #[test]
 fn it_builds_webpack() {
-    let fixture = Fixture::new("webpack");
```
ashleymichal: fixtures no longer use name for their directory name, but instead just accept whatever random directory the OS passes to them.

tests/build.rs
```diff
@@ -5,31 +5,24 @@ pub mod fixture;
...
     fixture.scaffold_webpack();
-    fixture.create_wrangler_toml(
-        r#"
-        type = "webpack"
-    "#,
-    );
+
+    let wrangler_toml = WranglerToml::webpack_no_config("test-build-webpack");
+    fixture.create_wrangler_toml(wrangler_toml);
+
```
ashleymichal: creating a wrangler.toml is now as simple as picking a "template" constructor. these can be modified in place for custom tests, but the constructors that exist are helpful for the existing tests.

tests/build.rs
```diff
@@ -5,31 +5,24 @@ pub mod fixture;
...
     build_creates_assets(&fixture, vec!["script.js"]);
-    fixture.cleanup();
```
ashleymichal: we don't need to explicitly call cleanup any longer, our impl of ManuallyDrop (on the Fixture struct) handles it for us.

tests/build.rs
```diff
@@ -5,31 +5,24 @@ pub mod fixture;
...
-    let fixture = Fixture::new("webpack");
+    let fixture = Fixture::new();
     fixture.scaffold_webpack();
-    fixture.create_wrangler_toml(
-        r#"
-        type = "webpack"
-    "#,
-    );
+
+    let wrangler_toml = WranglerToml::webpack_no_config("test-build-webpack");
```
ashleymichal: moved the human readable names to the wrangler.toml fields instead of naming the temp dirs after them.

tests/build.rs
```diff
@@ -329,21 +279,14 @@ fn it_builds_with_webpack_target_webworker() {
...
 
     build_creates_assets(&fixture, vec!["script.js"]);
-    fixture.cleanup()
 }
 
 fn build_creates_assets(fixture: &Fixture, script_names: Vec<&str>) {
-    // Lock to avoid having concurrent builds
-    let _g = BUILD_LOCK.lock().unwrap();
-
+    let _lock = fixture.lock();
```
ashleymichal: calling .lock on the fixture instance rather than the "global" static shared instance seems to have solved some of the domino effect on failing tests.

tests/fixture/mod.rs
```diff
@@ -74,30 +70,25 @@ impl Fixture {
...
+
+        lazy_static! {
+            static ref ONE_TEST_AT_A_TIME: Mutex<()> = Mutex::new(());
+        }
+
+        ONE_TEST_AT_A_TIME.lock().unwrap_or_else(|e| e.into_inner())
+    }
+}
+
+impl Drop for Fixture {
```
ashleymichal: our impl specifies that if the test succeeds (the test thread does not panic) we drop the temp dir. otherwise it is left in place. the `eprintln!` written out during `Fixture::new()` tells you where to find your failed test assets.

tests/fixture/wrangler_toml.rs
```diff
@@ -0,0 +1,100 @@
...
+    #[serde(rename = "kv-namespaces")]
+    pub kv_namespaces: Option<Vec<KvConfig<'a>>>,
+    pub site: Option<SiteConfig<'a>>,
+}
+
+impl WranglerToml<'_> {
+    pub fn webpack_no_config(name: &str) -> WranglerToml {
+        let mut wrangler_toml = WranglerToml::default();
+        wrangler_toml.name = Some(name);
+        wrangler_toml.workers_dev = Some(true);
```
ashleymichal: this default `wrangler_dev` value is kind of a holdover from how the test fixtures were originally.

tests/fixture/mod.rs
EverlastingBugstopper: omg wow. yes. one million times yes.

tests/build.rs
EverlastingBugstopper: brilliant.

tests/build.rs
EverlastingBugstopper: :octocat:

tests/fixture/wrangler_toml.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub workers_dev: Option<bool>,
+    pub route: Option<&'a str>,
+    pub routes: Option<Vec<&'a str>>,
+    pub zone_id: Option<&'a str>,
+    pub webpack_config: Option<&'a str>,
+    pub private: Option<bool>,
+    pub env: Option<HashMap<&'a str, EnvConfig<'a>>>,
+    #[serde(rename = "kv-namespaces")]
+    pub kv_namespaces: Option<Vec<KvConfig<'a>>>,
+    pub site: Option<SiteConfig<'a>>,
```
EverlastingBugstopper: the only problem i have with this pr is all of the lifetimes - it seems to me that the `<&'a str>` could all be `String` just for readability sake but I might be missing something here. I know we talked about this yesterday but wanted to leave this here regardless.

ashleymichal: i mean honestly, they're all concentrated here, and the only reason to do this is to simplify the calls whenever we add something (i.e. don't need to call to_string on everything). no need to allocate a whole String for each of these when a slice will do. i also think it wouldn't hurt for us to be comfortable with seeing explicit lifetimes in some of our code (again, explicit, because the lifetimes are always there, they are just implicit everywhere).

ashleymichal:
going to force push these commits and pull against master.

tests/build.rs
gabbifish: yeehaw!

tests/fixture/wrangler_toml.rs
gabbifish: i am really on board with the idea of trying to reduce `String` declarations for the sake of good memory practices :)

tests/build.rs
gabbifish: nice :)

tests/fixture/mod.rs
```diff
@@ -1,48 +1,44 @@
...
 
-const BUNDLE_OUT: &str = "./worker";
+const BUNDLE_OUT: &str = "worker";
 
 pub struct Fixture {
-    name: String,
+    // we wrap the fixture's tempdir in a `ManuallyDrop` so that if a test
+    // fails, its directory isn't deleted, and we have a chance to manually
+    // inspect its state and figure out what is going on.
+    dir: ManuallyDrop<TempDir>,
```
gabbifish: Super, super nice! Does this mean testers will need to manually delete these leftover directories if there is a panic?

ashleymichal: yes they will not be cleaned up automatically; however since they are randomly generated by the OS, they will not cause collisions with subsequent test runs. should be trivial to write a shell script to clean them up after, as they consistently go to the same directory.

PR #934, opened 2019-12-06
WIP: open browser on wsl

### LINKED ISSUE #840: WSL preview doesn't open browser, maybe print URL to terminal window as workaround?
#### Environment

* operating system: Windows 10 build 19013
* output of `rustc -V`: rustc 1.36.0
* output of `node -v`: v12.10.0
* output of `wrangler -V`: wrangler 1.5.0

#### Steps to reproduce
`$ wrangler preview --watch`
or
`$ wrangler preview`

A browser opening

First, thank you very much for making this 🙏🏻, it's really a joy to have a tool like this. 🙌🏻 I use it both on my Mac and Windows, but I'm reporting this bug/feature request as a Windows user.

The browser doesn't open. While WSL2 doesn't have xdg-open like the Linux variants, you can open the default browser using cmd.exe because of interop with Windows.

Alt text

So a command of:
**$ cmd.exe /c start http://example.com?foo=bar** would work on WSL2 environment.

Also printing --verbose doesn't print the URL either.
Alt text

I do hope you add this 🙏🏻. Also maybe an interim, have the preview URL printed, so I can copy and paste manually on the browser?

thisguychris:
fixes #840

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,7 @@ sha2 = "0.8.0"
 data-encoding = "2.1.2"
 ignore = "0.4.10"
 tempfile = "3.1.0"
+wsl = "0.1.0"
 
 [dev-dependencies]
 assert_cmd = "0.11.1"
--- a/src/commands/preview/mod.rs
+++ b/src/commands/preview/mod.rs
@@ -122,7 +122,7 @@ fn open_browser(url: &str) -> Result<(), failure::Error> {
         let mac_cmd = format!(r#"open "{}""#, url);
         Command::new("sh").arg("-c").arg(&mac_cmd).output()?
     };
-    
+
     Ok(())
 }
 
--- a/src/commands/preview/mod.rs
+++ b/src/commands/preview/mod.rs
@@ -105,17 +105,24 @@ pub fn preview(
 
 fn open_browser(url: &str) -> Result<(), failure::Error> {
     let _output = if cfg!(target_os = "windows") {
+        println!("windows");
         let url_escaped = url.replace("&", "^&");
         let windows_cmd = format!("start {}", url_escaped);
         Command::new("cmd").args(&["/C", &windows_cmd]).output()?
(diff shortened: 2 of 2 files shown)


thisguychris:
Here's what I have so far, 

It does go to the correct conditional (printing `wsl`), but when I try to invoke `cmd.exe`. Interop seems not to work and cannot find cmd.exe, I tried referencing it directly via /mnt/ to no avail.

If I do invoke `xdg-open`, it does call it since rust treats it as Linux. So I'm not sure what to do at this point. But ultimately, I hope we can make it easier for WSL users to work out of the box.

EverlastingBugstopper:
> If I do invoke xdg-open, it does call it since rust treats it as Linux. So I'm not sure what to do at this point. 

Does it open the browser?

> But ultimately, I hope we can make it easier for WSL users to work out of the box.

I'm hoping that [this feature](https://github.com/cloudflare/wrangler/pull/883) will essentially make `wrangler preview` obsolete and should work out of the box with WSL. I have some testing instructions on that branch if you'd like to play around with it :)

thisguychris:
hey @EverlastingBugstopper sorry I missed this! 🤦🏻‍♂️ I would love to test that branch, is it [this?](https://github.com/cloudflare/wrangler/tree/avery/a-time-sync)

> Does it open the browser?
For `xdg-open` it does open a browser If you set the $BROWSER variable, (one caveat with windows is `xdg-open` doesn't work if the path has spaces on it, even if you escaped it.

So it's either you 1) have it installed outside of Program Files, like "C:\tools\firefox.exe" or 2) just make a symbolic link to the exe.). PR on the way 🔜 for the docs.

EverlastingBugstopper:
That's the correct feature branch, but my latest work is [here](https://github.com/cloudflare/wrangler/tree/avery/websockets)

ashleygwilliams:
closing in favor of a docs fix!

PR #936, opened 2019-12-06
Addresses clippy warnings

EverlastingBugstopper:


tests/fixture/mod.rs
```diff
@@ -21,6 +21,12 @@ pub struct Fixture {
     dir: ManuallyDrop<TempDir>,
 }
```
ashleymichal: curious why clippy wanted this?

PR #937, opened 2019-12-07
Provide helpful error when user accidentally puts kv-namespace information under [site]

### LINKED ISSUE #798: Direct users to add kv-namespaces to specific portion of toml
cc @steveklabnik 

If users add `kv-namespaces` snippet under a `[site]` config, the namespace is not parsed into a Wrangler `Manifest` so the namespace is still inaccessible to a worker.

```
account_id = "f5518bfbf88c230551a64797c221c7da"
name = "website"
type = "webpack"
route = "www.steveklabnik.com/*"
zone_id = "9359b254c46ff9f5c546203686794862"
workers_dev = false

[site]
bucket = "./public"
entry-point = "workers-site"

kv-namespaces = [
         { binding = "new_namespace", id = "5161cf32bf194293a8591c3c65a24c0a" }
]
```

We should do two things here: 

1) If we detect kv-namespaces in the site struct, we should warn them to move it
2) When we ask people to add kv-namespaces to their toml, we should direct them to add it to the top level config, and not below site

**Alternatively** we could just rewrite their toml for them but that gets a little hairy, especially if they have things like comments

gabbifish:
This PR closes #798.

If someone provides a wrangler.toml with
```
account_id = "c2c4de4232cdae687dc7a4a0323df63a"
name = "butt"
type = "webpack"
route = ""
workers_dev = true
zone_id = ""

[site]
bucket = "./public"
entry-point = "workers-site"

kv-namespaces = [
         { binding = "BUTTS", id = "4197135817e648c1957dcc6c3fc3d462" }
]
```

we now provide an error message on any `wrangler` command:
```
Error: kv-namespaces should not live under the [site] table in wrangler.toml; please move it above [site].
```

I wanted to set up TOML serialization so that `[site]` would be an inline TOML table (so there would be no reasonable way to put `kv-namespaces` within `site`!) but alas, inline TOML table serialization isn't supported by the serde TOML serializer yet.


src/settings/target/manifest.rs
```diff
@@ -46,7 +48,16 @@ impl Manifest {
     pub fn new(config_path: &Path) -> Result<Self, failure::Error> {
         let config = read_config(config_path)?;
 
-        let manifest: Manifest = config.try_into()?;
+        let manifest: Manifest = match config.try_into() {
+            Ok(m) => m,
+            Err(e) => {
+                if e.to_string().contains("unknown field `kv-namespaces`") {
```
gabbifish: if there's a nicer way to match on the specific error thrown by `serde` when the `#[serde(deny_unknown_fields)]` condition is triggered, I'm all ears!

PR #941, opened 2019-12-10
[WIP] serialize maps and arrays

EverlastingBugstopper:
Currently blocked on serializing maps - arrays seems to work beautifully but maps don't want to serialize. they're getting parsed as an `Object` not a `Subtype`. if i remove the `preview` field from the `MapData` struct and only print out the description, it will serialize as a `Subtype`, so there's something wonky there. I'm really not sure what it is though because I've modeled MapData just like ArrayData and it works fine. 

Running with `RUST_LOG=info wrangler dev` will print out the message that its trying to serialize. Run the command from within a dir that points to a project that logs out an array and an object like so.

```
})

  console.log(["hello", "world"])
  let map = new Map()
  map.set("hello", "world")
  console.log(map)
  return new Response(`you requested url ${request.url} with method ${request.method}`);
}
```

### DIFF
--- a/src/commands/dev/socket/events/console/messages/mod.rs
+++ b/src/commands/dev/socket/events/console/messages/mod.rs
@@ -25,7 +25,7 @@ pub enum LogMessage {
     BigInt(BigIntData),
     Boolean(BooleanData),
     #[serde(rename = "string")]
-    StringVal(StringData),
+    StringJs(StringData),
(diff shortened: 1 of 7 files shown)


EverlastingBugstopper:
everything has moved to github.com/everlastingbugstopper/chrome-devtools-rs

PR #942, opened 2019-12-10
Fix map serialization

gabbifish:
As it turns out, the devtools protocol does something funny when string types for map keys. 
Take a look:
```
{
  "method": "Runtime.consoleAPICalled",
  "params": {
    "type": "log",
    "args": [
      {
        "type": "object",
        "subtype": "map",
        "className": "Map",
        "description": "Map(1)",
        "objectId": "{\"injectedScriptId\":2,\"id\":2}",
        "preview": {
          "type": "object",
          "subtype": "map",
          "description": "Map(1)",
          "overflow": false,
          "properties": [],
          "entries": [
            {
              "key": {
                "type": "string",
                "description": "hello",
                "overflow": false,
                "properties": []
              },
              "value": {
                "type": "string",
                "description": "world",
                "overflow": false,
                "properties": []
              }
            }
          ]
        }
      }
    ],
....
}
```
The string type here is not the usual expected 
```
{ "name": "0", "type": "string", "value": "hello" }
```
meaning that deserialization with the current logic in this feature branch will not work. If we allow the StringData struct to accept a `description` field to replace a `value` field, we can allow wrangler to deserialize these map keys/values and fix this deserialization problem.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2795,6 +2795,7 @@ dependencies = [
  "reqwest 0.9.22 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.102 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.41 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_with 1.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "sha2 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "tempfile 3.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "text_io 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ openssl = "0.10.26"
 reqwest = "0.9.18"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.39"
+serde_with = "1.3.1"
 toml = "0.5.0"
 uuid = "0.7"
 which = "2.0.1"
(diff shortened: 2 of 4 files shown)


EverlastingBugstopper:
this works for strings, but not other types. this console message will still not serialize properly:

```
})

let obj = { hello: "world", one: "two", fun, o: {
  "somethindg": {
    nest: "again"
  }
}}

  let map = new Map()
  map.set("hello", obj)
  console.log(map)
  return new Response(`you requested url ${request.url} with method ${request.method}`);
}
```

gabbifish:
Yikes... looks like we might have to manually extract what each of these types looks like as a key. All the more reason to honestly make a seperate enum and structs for key types in maps...

Are there docs for what these key types look like represented in JSON?

gabbifish:
JSON representation of objects:
```
{
  "method": "Runtime.consoleAPICalled",
  "params": {
    "type": "log",
    "args": [
      {
        "type": "object",
        "subtype": "map",
        "className": "Map",
        "description": "Map(1)",
        "objectId": "{\"injectedScriptId\":2,\"id\":1}",
        "preview": {
          "type": "object",
          "subtype": "map",
          "description": "Map(1)",
          "overflow": false,
          "properties": [],
          "entries": [
            {
              "key": {
                "type": "string",
                "description": "hello",
                "overflow": false,
                "properties": []
              },
              "value": {
                "type": "object",
                "description": "Object",
                "overflow": false,
                "properties": [
                  {
                    "name": "hello",
                    "type": "string",
                    "value": "world"
                  },
                  {
                    "name": "one",
                    "type": "string",
                    "value": "two"
                  },
                  {
                    "name": "o",
                    "type": "object",
                    "value": "Object"
                  }
                ]
              }
            }
          ]
        }
      }
    ],
    ...
}
```

EverlastingBugstopper:
these are the docs: https://chromedevtools.github.io/devtools-protocol/tot/Runtime#type-RemoteObject

but they are confusing.

EverlastingBugstopper:
Closing this because code now lives in `chrome-devtools-rs` and is fixed

PR #943, opened 2019-12-11
wip: Basic source-map support

xtuc:
Configures webpack to emit an inline source-map.
Worker will send the source-map to the devtools which will resolve source-level
informations.

Note that it only works with the new worker editor, the old editior will be eventually updated too.


xtuc:
We should get the source-map regardless of the webpack config and inject it when publish. `wrangler dev` should be able to use it as well.

wranglerjs/index.js
```diff
@@ -54,6 +54,16 @@ function filterByExtension(ext) {
   }
   config.target = "webworker";
 
+  if (config.devtool !== undefined && config.devtool !== "inline-source-map") {
```
EverlastingBugstopper: i'm not sure if we want to be this opinionated about source maps - i'd say we could throw a warning saying that source maps are highly recommended but i'm not quite sure we want to tell devs that they are wrong for not using source maps.

EverlastingBugstopper: especially given the size constraints we put on our customers

wranglerjs/index.js
```diff
@@ -54,6 +54,16 @@ function filterByExtension(ext) {
...
+    throw error(
+      "Building a Cloudflare Worker with `devtool` " +
+        JSON.stringify(config.devtool) +
+        " is not supported. Wrangler will set `inline-source-map` by default, please remove " +
+        "the `devtool` key in your webpack configuration."
```
EverlastingBugstopper: If we are going to error, perhaps we should say "please remove the devtool key in your webpack configuration, or set it to inline-source-map.

ashleymichal:
similar to the way we minify by default bc it is "production" behavior, i'm not sure i agree that we should send source maps by default, as it is a debugging tool for developers. perhaps i am out of touch with common practice but that's my instinct.

wranglerjs/index.js
xtuc: I believe source-maps will always be accounted in the bundle size, because they will ship along with the worker somehow.

ashleymichal: yah the most obvious way to send source maps along with a worker without tacking on kv (and the latency involved) would be to include them in the bundle as text blob bindings.

aleclarson:
It would be great if a `Error.prototype.stack` shim was used to map stack traces back to their original source in case `console.log(error.stack)` or similar is used.

aleclarson:
> It would be great if a `Error.prototype.stack` shim was used to map stack traces back to their original source in case `console.log(error.stack)` or similar is used.


I tried using [`node-source-map-support`](https://github.com/evanw/node-source-map-support/blob/master/source-map-support.js) but it doesn't know how to load the `worker.js` script in a Cloudflare Worker environment. Is there a way to do that?

EverlastingBugstopper:
Hey @aleclarson - we don't have this on our roadmap right now so I can't give an ETA as to when this will be ready. @xtuc may have some ideas around this but I know he has other things going on right now as well

PR #944, opened 2019-12-11
Alewis/test routes toml

ashleymichal:
Sorry this is a lot, but the high level overview is that this PR adds an enum, `DeployTarget`, and a function `deploy_target` to `Manifest` that extracts the relevant deploy information from the manifest in preparation for publishing. 

This PR covers adding unit tests for pulling deploy targets from the top level of a wrangler toml. There are helper methods added here, also, for things like getting the worker name from the manifest, and lots and lots of test suite helpers.

PR #948, opened 2019-12-11
Add and test environment override logic

ashleymichal:
depends on #944

for realsies mostly unit tests because this logic is like whack a mole. and i have STILL probably missed a few use cases.

### DIFF
--- a/src/settings/toml/deploy_target.rs
+++ b/src/settings/toml/deploy_target.rs
@@ -11,11 +11,7 @@ pub struct RouteConfig {
 impl RouteConfig {
     fn has_conflicting_targets(&self) -> bool {
         if self.workers_dev.unwrap_or_default() {
-            if let Some(pattern) = &self.route {
-                !pattern.is_empty() || self.routes.is_some() // this is all so messy because of deserializer
-            } else {
-                self.routes.is_some()
-            }
+            self.routes_defined()
         } else {
             if let Some(pattern) = &self.route {
                 !pattern.is_empty() && self.routes.is_some()
@@ -25,22 +21,36 @@ impl RouteConfig {
         }
     }
 
+    pub fn routes_defined(&self) -> bool {
+        if let Some(pattern) = &self.route {
+                !pattern.is_empty() || self.routes.is_some() // this is all so messy because of deserializer
+            } else {
+                self.routes.is_some()
+            }
+    }
+
     pub fn is_zoneless(&self) -> bool {
         self.workers_dev.unwrap_or_default() && !self.has_conflicting_targets()
     }
 
-    pub fn is_complete_zoned(&self) -> bool {
-        (self.route.is_some() || self.routes.is_some()) && self.zone_id.is_some()
+    pub fn is_zoned(&self) -> bool {
+        self.routes_defined() || !self.missing_zone_id()
     }
 
     // zone id is another weird one where `Some("")` is treated the same as `None`
     pub fn missing_zone_id(&self) -> bool {
-        let result = !self.workers_dev.unwrap_or_default()
-            && (self.route.is_some() || self.routes.is_some());
         if let Some(zone_id) = &self.zone_id {
-            result && zone_id.is_empty()
+            zone_id.is_empty()
         } else {
-            result
+            true
+        }
+    }
+
+    pub fn workers_dev_false_by_itself(&self) -> bool {
+        if let Some(workers_dev) = self.workers_dev {
+            !workers_dev && !self.routes_defined()
+        } else {
+            false
         }
     }
 }
--- a/src/settings/toml/deploy_target.rs
+++ b/src/settings/toml/deploy_target.rs
@@ -23,10 +23,10 @@ impl RouteConfig {
 
     pub fn routes_defined(&self) -> bool {
         if let Some(pattern) = &self.route {
-                !pattern.is_empty() || self.routes.is_some() // this is all so messy because of deserializer
-            } else {
-                self.routes.is_some()
-            }
+            !pattern.is_empty() || self.routes.is_some() // this is all so messy because of deserializer
+        } else {
+            self.routes.is_some()
+        }
     }
 
     pub fn is_zoneless(&self) -> bool {
--- a/src/settings/toml/environment.rs
+++ b/src/settings/toml/environment.rs
@@ -1,7 +1,8 @@
-use super::kv_namespace::KvNamespace;
-use super::site::Site;
-
 use serde::{Deserialize, Serialize};
+
+use crate::settings::toml::deploy_target::RouteConfig;
+use crate::settings::toml::kv_namespace::KvNamespace;
+use crate::settings::toml::site::Site;
 
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 pub struct Environment {
@@ -16,4 +17,15 @@ pub struct Environment {
     pub site: Option<Site>,
     #[serde(rename = "kv-namespaces")]
     pub kv_namespaces: Option<Vec<KvNamespace>>,
+}
+
+impl Environment {
+    pub fn route_config(&self) -> Result<RouteConfig, failure::Error> {
+        Ok(RouteConfig {
+            workers_dev: self.workers_dev,
+            route: self.route.clone(),
+            routes: self.routes.clone(),
+            zone_id: self.zone_id.clone(),
+        })
+    }
 }
--- a/src/settings/toml/manifest.rs
+++ b/src/settings/toml/manifest.rs
@@ -145,7 +145,29 @@ impl Manifest {
     pub fn deploy_target(&self, env: Option<&str>) -> Result<DeployTarget, failure::Error> {
         let script = self.worker_name(env);
         validate_worker_name(&script)?;
+        if let Some(environment) = self.get_environment(env)? {
+            // if there is a complete environment level deploy target, return that
+            let mut env_route_config = environment.route_config()?;
+            if env_route_config.workers_dev_false_by_itself() {
+                failure::bail!("you must set workers_dev = true or a zoned deploy config.")
+            }
+            if env_route_config.is_zoneless() || env_route_config.is_zoned() {
+                if env_route_config.missing_zone_id() && env_route_config.routes_defined() {
+                    // if there is an incomplete environment level deploy target,
+                    // it is not zoneless, and routes do not inherit,
+                    // so fill in the zone id and build from that
+                    env_route_config.zone_id = self.zone_id.clone();
+                }
+                return DeployTarget::build(&script, &env_route_config);
+            }
+        }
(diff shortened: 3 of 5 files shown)


ashleymichal:
fixup commit contains cargo fmt

src/settings/toml/tests/deploy_target.rs
```diff
@@ -335,3 +335,540 @@ fn it_errors_on_deploy_target_routes_and_workers_dev_true() {
...
+    assert!(actual_deploy_target.is_err());
+}
+
+#[test]
+fn when_top_level_empty_zoneless_env() {
+    let env_name = "test";
+
+    // TODO: THIS TEST SHOULD PASS
+    // when env.workers_dev = false
+    let env_config = EnvConfig::zoneless(false);
```
gabbifish: Shouldn't this then require a zone and a route(s) to pass, then?

src/settings/toml/tests/deploy_target.rs
```diff
@@ -335,3 +335,540 @@ fn it_errors_on_deploy_target_routes_and_workers_dev_true() {
...
+    let script = "top_level_empty_env_zoned_single_route_empty";
+    let test_toml = WranglerToml::webpack_with_env(script, env_name, env_config);
+    let toml_string = toml::to_string(&test_toml).unwrap();
+    let manifest = Manifest::new_from_string(toml_string).unwrap();
+
+    let actual_deploy_target = manifest.deploy_target(Some(env_name));
+
+    assert!(actual_deploy_target.is_err());
+
+    // when zone id is missing, error
```
gabbifish: each of these sub-tests could honestly could happily live as distinct tests

src/settings/toml/tests/deploy_target.rs
```diff
@@ -335,3 +335,540 @@ fn it_errors_on_deploy_target_routes_and_workers_dev_true() {
...
+    assert_eq!(actual_deploy_target, DeployTarget::Zoneless);
+}
+
+#[test]
+fn when_top_level_zoneless_env_zoneless() {
+    let env_name = "test";
+
+    // TODO: SHOULD THIS TEST PASS?
+    // when env.workers_dev = false
+    let env_config = EnvConfig::zoneless(false);
```
gabbifish: this should require a route and zone_id to pass, right? For now, this test as written should fail.

src/settings/toml/tests/deploy_target.rs
```diff
@@ -335,3 +335,540 @@ fn it_errors_on_deploy_target_routes_and_workers_dev_true() {
...
+    let script = "top_level_zoneless_env_zoned_multi_route_empty";
+    let test_toml = WranglerToml::webpack_zoneless_with_env(script, true, env_name, env_config);
+    let toml_string = toml::to_string(&test_toml).unwrap();
+    let manifest = Manifest::new_from_string(toml_string).unwrap();
+
+    let actual_deploy_target = manifest.deploy_target(Some(env_name));
+
+    assert!(actual_deploy_target.is_err());
+
+    // when route key also present, error
```
gabbifish: Nice!

src/settings/toml/tests/deploy_target.rs
```diff
@@ -335,3 +335,540 @@ fn it_errors_on_deploy_target_routes_and_workers_dev_true() {
...
+
+#[test]
+fn when_top_level_zoned_env_zoneless() {
+    let zone_id = "samplezoneid";
+    let pattern = "hostname.tld/*";
+    let env_name = "test";
+
+    // TODO: SHOULD THIS TEST PASS?
+    // when env.workers_dev = false
+    let env_config = EnvConfig::zoneless(false);
```
gabbifish: this should pass and inherit the top level zoned deploy configs, right?

src/settings/toml/deploy_target.rs
```diff
@@ -25,22 +21,36 @@ impl RouteConfig {
         }
     }
 
+    pub fn routes_defined(&self) -> bool {
+        if let Some(pattern) = &self.route {
+            !pattern.is_empty() || self.routes.is_some() // this is all so messy because of deserializer
```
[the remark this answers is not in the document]
ashleymichal: i can check first if self.routes is present which would be better, but i do still need to get the pattern out of self.route and see if it is empty because (today) we allow you to have `route = ""` in your toml.

src/settings/toml/environment.rs
```diff
@@ -17,3 +18,14 @@ pub struct Environment {
     #[serde(rename = "kv-namespaces")]
     pub kv_namespaces: Option<Vec<KvNamespace>>,
 }
+
+impl Environment {
+    pub fn route_config(&self) -> Result<RouteConfig, failure::Error> {
+        Ok(RouteConfig {
+            workers_dev: self.workers_dev,
+            route: self.route.clone(),
```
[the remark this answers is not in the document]
ashleymichal: we do, later. this is kind of an intermediate struct; DeployTarget is what we ultimately put together and send to Publish, and its zoned version does contain JUST a routes vec.

src/settings/toml/manifest.rs
```diff
@@ -145,7 +145,29 @@ impl Manifest {
...
         let script = self.worker_name(env);
         validate_worker_name(&script)?;
+        if let Some(environment) = self.get_environment(env)? {
+            // if there is a complete environment level deploy target, return that
+            let mut env_route_config = environment.route_config()?;
+            if env_route_config.workers_dev_false_by_itself() {
+                failure::bail!("you must set workers_dev = true or a zoned deploy config.")
+            }
+            if env_route_config.is_zoneless() || env_route_config.is_zoned() {
+                if env_route_config.missing_zone_id() && env_route_config.routes_defined() {
```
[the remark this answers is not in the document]
ashleymichal: this IS where we make all those decisions. all of these PRs are replacing the logic in `get_target` with this more focused code, so negotiate_zoneless goes away.

src/settings/toml/tests/deploy_target.rs
ashleymichal: the assert is that this errors...

src/settings/toml/tests/deploy_target.rs
ashleymichal: yeah i think you're right. i can break them out. the descriptions became like...two friggin lines though.

src/settings/toml/tests/deploy_target.rs
ashleymichal: the assert is that this errors.

src/settings/toml/tests/deploy_target.rs
ashleymichal: i discussed this with avery; if you explicitly put `workers_dev = false` in your toml anywhere on its own, it's super ambiguous, and so i feel it should error. i should probably test this against today's behavior, but it's wicked inconsistent. imho.

src/settings/toml/tests/deploy_target.rs
```diff
@@ -335,3 +335,540 @@ fn it_errors_on_deploy_target_routes_and_workers_dev_true() {
 
     assert!(manifest.deploy_target(environment).is_err());
 }
+
+// ENVIRONMENT TESTS
+// Top level empty
+#[test]
+fn when_top_level_empty_env_empty() {
```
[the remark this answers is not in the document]
ashleymichal: i can do that when i break them all out. it's... like two lines worth of friggen description. but maybe with comments i can reduce it. the matrix is wide.

src/settings/toml/tests/deploy_target.rs
gabbifish: Yup! Just responding to the TODO above. I still think this test should fail, as it does now.

src/settings/toml/tests/deploy_target.rs
gabbifish: the todo says this should pass, though. if it does pass, I think it needs to meet the conditions I mentioned in this comment.

src/settings/toml/environment.rs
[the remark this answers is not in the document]
gabbifish: why not do it when creating the struct, and have a special deserializer that handles converting `route` -> `routes` when only one route is present? I think this would reduce confusion and inconsistency :)

src/settings/toml/manifest.rs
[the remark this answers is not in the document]
gabbifish: I think this decision making code could be made a little simpler / more logical for the reasons I wrote up above--what do you think?

src/settings/toml/tests/deploy_target.rs
EverlastingBugstopper: it's not entirely clear what `EnvConfig::zoneless(false)` is doing here - not easy for me to parse out what exactly the toml would look like and what is supposed to fail/succeed. i'm assuming `EnvConfig::zoneless(false)` means that `workers_dev = false`, in which case we might want to rename that method

src/settings/toml/environment.rs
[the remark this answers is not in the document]
ashleymichal: so, the way this logic goes, this is valid:

```toml
route = ""
routes = ["hostname.tld"]
```
i'm not 100% certain how to write the serializer that makes this okay. i'll try in some refactors to make the "RouteConfig" struct go away, but it was in the path of least mental overhead to get from the massive function in manifest to these smaller structs.

src/settings/toml/manifest.rs
```diff
@@ -145,7 +145,29 @@ impl Manifest {
     pub fn deploy_target(&self, env: Option<&str>) -> Result<DeployTarget, failure::Error> {
         let script = self.worker_name(env);
         validate_worker_name(&script)?;
+        if let Some(environment) = self.get_environment(env)? {
+            // if there is a complete environment level deploy target, return that
+            let mut env_route_config = environment.route_config()?;
+            if env_route_config.workers_dev_false_by_itself() {
+                failure::bail!("you must set workers_dev = true or a zoned deploy config.")
```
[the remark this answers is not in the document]
ashleymichal: adding in new PR.

src/settings/toml/manifest.rs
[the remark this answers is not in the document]
ashleymichal: i think... i will try. the deserializer logic might help but it's still wacky.

src/settings/toml/tests/deploy_target.rs
ashleymichal: when you say "this test should fail", you mean this condition should error? the test passing == the condition errors. it's confusing but i just want to  make sure i follow

src/settings/toml/tests/deploy_target.rs
ashleymichal: by "test should pass" i mean the assert should be true; in this case, should error.

ashleymichal:
closing this and addressing issues in fresh pr.

PR #949, opened 2019-12-12
Add Zoneless struct to include script/account id info

ashleymichal:
I noticed that actually zoneless is more than one piece of information as far as what you need to "deploy", so created a struct that carries the script name and the account id, updated tests to use it.

* make webpack_build test toml for "legacy" parsing

### DIFF
--- a/src/settings/toml/deploy_target.rs
+++ b/src/settings/toml/deploy_target.rs
@@ -24,7 +24,8 @@ impl RouteConfig {
 
     pub fn routes_defined(&self) -> bool {
         if let Some(pattern) = &self.route {
-            !pattern.is_empty() || self.routes.is_some() // this is all so messy because of deserializer
+            // this is all so messy because of deserializer
+            !pattern.is_empty() || self.routes.is_some()
         } else {
             self.routes.is_some()
         }
@@ -67,7 +68,7 @@ impl RouteConfig {
 
 impl DeployTarget {
     pub fn build(
-        script: &String,
+        script_name: &String,
         route_config: &RouteConfig,
     ) -> Result<DeployTarget, failure::Error> {
         if route_config.is_zoneless() {
@@ -77,7 +78,7 @@ impl DeployTarget {
                 failure::bail!("field `account_id` is required to deploy to workers.dev");
             }
             let zoneless = Zoneless {
-                script: script.to_string(),
+                script_name: script_name.to_string(),
                 account_id: account_id.to_string(),
             };
 
@@ -101,12 +102,12 @@ impl DeployTarget {
             // TODO: these should be an if/else if block; write deserializer
             // for `route` key that turns `Some("")` into `None`
             if let Some(pattern) = &route_config.route {
-                zoned.add_route(&pattern, script);
+                zoned.add_route(&pattern, script_name);
             }
 
             if let Some(patterns) = &route_config.routes {
                 for pattern in patterns {
-                    zoned.add_route(&pattern, script);
+                    zoned.add_route(&pattern, script_name);
                 }
             }
 
@@ -128,7 +129,7 @@ pub enum DeployTarget {
 #[derive(Clone, Debug, PartialEq)]
 pub struct Zoneless {
     pub account_id: String,
-    pub script: String,
+    pub script_name: String,
 }
 
 #[derive(Clone, Debug, PartialEq)]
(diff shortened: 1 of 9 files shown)


src/settings/toml/deploy_target.rs
```diff
@@ -23,10 +24,10 @@ impl RouteConfig {
 
     pub fn routes_defined(&self) -> bool {
         if let Some(pattern) = &self.route {
-                !pattern.is_empty() || self.routes.is_some() // this is all so messy because of deserializer
-            } else {
-                self.routes.is_some()
-            }
+            !pattern.is_empty() || self.routes.is_some() // this is all so messy because of deserializer
```
EverlastingBugstopper: nit - commment on line above

src/settings/toml/deploy_target.rs
```diff
@@ -61,7 +71,17 @@ impl DeployTarget {
         route_config: &RouteConfig,
     ) -> Result<DeployTarget, failure::Error> {
         if route_config.is_zoneless() {
-            Ok(DeployTarget::Zoneless)
+            // account_id is required
+            let account_id = route_config.account_id.as_ref().unwrap();
```
EverlastingBugstopper: can we just do `let account_id = &route_config.account_id`? if not we should do `as_ref()?` if possible

src/settings/toml/deploy_target.rs
```diff
@@ -101,10 +121,16 @@ impl DeployTarget {
...
 pub enum DeployTarget {
-    Zoneless,
+    Zoneless(Zoneless),
     Zoned(Zoned),
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct Zoneless {
+    pub account_id: String,
+    pub script: String,
```
EverlastingBugstopper: nit: i'd call `script` `name` or even `script_name` to remain consistent with the toml

src/settings/toml/tests/deploy_target.rs
```diff
@@ -31,8 +48,11 @@ fn it_errors_on_zoneless_deploy_target_missing_workers_dev() {
 }
 
 #[test]
-fn it_errors_on_deploy_target_missing_name() {
-    let test_toml = WranglerToml::webpack_zoneless("", true);
+fn it_errors_on_deploy_target_missing_account_id() {
+    let script = "zoneless_no_account_id";
+    let account_id = "account_id";
+    let mut test_toml = WranglerToml::webpack_zoneless(script, account_id, true);
+    test_toml.account_id = None;
```
EverlastingBugstopper: can we pass None directly to `webpack_zoneless`?

src/settings/toml/tests/deploy_target.rs
```diff
@@ -619,7 +661,8 @@ fn when_top_level_zoneless_env_zoned_single_route() {
     let env_config = EnvConfig::zoned_single_route(zone_id, pattern);
 
     let script = "top_level_zoneless_env_zoned_single_route_no_zone_id";
-    let test_toml = WranglerToml::webpack_zoneless_with_env(script, true, env_name, env_config);
+    let test_toml =
```
EverlastingBugstopper: i wonder if it makes sense to just create a struct every time instead of positional parameters - it's kind of hard to read

src/settings/toml/deploy_target.rs
ashleymichal: the issue is that the type of `route_config.account_id` here is `&Option<String>`, not String. so it needs to be unwrapped; but i can't have just a naked string, so it's gotta  be an `as_ref` call, and you can't call `?` on the result of `as_ref`.

i have, however, been considering just making two constructors: one for a zoned and one for a zoneless, because by the time i call this i should know what it's gonna be.

src/settings/toml/tests/deploy_target.rs
ashleymichal: discussed in air; the idea is to have convenience constructors for valid tomls, and manipulate them manually in the case that we want to build "EVIL" tomls

src/settings/toml/tests/deploy_target.rs
ashleymichal: discussed this in air also; creating a struct every time would require that you declare the entire toml, in which case we may as well just go back to writing out entire toml strings. however as a compromise we can reduce the number of positional arguments in some of these

ashleymichal:
closing in favor of cleaner prs: one for functionality, one for tests.

PR #955, opened 2019-12-13
Allow POST-ing of files in `wrangler preview`

### LINKED ISSUE #282: Support binary POST data in preview curl
I've used wrangler to write a worker that processed image uploads, and it would be nice to use `POST @./my_image.jpg` with `wrangler preview`, similar to how curl does it.

gabbifish:
This PR closes #282.

Previously, you could only POST strings directly provided to `wrangler preview post`:
```console
$ wrangler preview post hey --headless
 Your wrangler.toml is missing the following fields: ["account_id"]
 Falling back to unauthenticated preview.
 Your Worker responded with: hey
```

This PR allows you to post files:
```console
$ wrangler preview post @./index.js --headless 
 Your wrangler.toml is missing the following fields: ["account_id"]
 Falling back to unauthenticated preview.
 Your Worker responded with: <contents of ./index.js>
```

If a body argument to `wrangler preview post` starts with `@` but the filename following `@` does not exist, you get the following error message:
```console
$ wrangler preview post @not_real.json --headless
Error: POST value started with @ so wrangler tried to open file "not_real.json". Encountered error "No such file or directory (os error 2)"
```

### DIFF
--- a/src/commands/preview/mod.rs
+++ b/src/commands/preview/mod.rs
@@ -137,8 +137,6 @@ fn post(
             .send(),
         None => client.post(PREVIEW_ADDRESS).header("Cookie", cookie).send(),
     };
-    let msg = format!("POST {}", PREVIEW_ADDRESS);
-    message::preview(&msg);
     Ok(res?.text()?)
 }
 
--- a/src/commands/preview/mod.rs
+++ b/src/commands/preview/mod.rs
@@ -31,7 +31,7 @@ pub fn preview(
     mut target: Target,
     user: Option<GlobalUser>,
     method: HTTPMethod,
-    body: Option<String>,
+    body: Option<Vec<u8>>,
     livereload: bool,
     verbose: bool,
     headless: bool,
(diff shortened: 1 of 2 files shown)


ashleymichal:
can a user escape the `@`?

gabbifish:
@ashleymichal I tried to find how curl handles this and it looks like it actually just strips the first `@` character. Going to make sure this is consistently implemented here (I think at the moment I strip all `@` chars).

gabbifish:
oof, thinking about it, the original #282 proposal to just put `@` in front of the filename is not great for cases where someone does want to post a string starting with `@`. Curl can handle this because curl expects `@` to be provided with a `fileUpload=` prefix. I wouldn't be opposed to creating a separate `--file` argument for the file upload; I don't think the original intention to map curl's approach to wrangler works, and think the explicitness of an argument would be good here.

ashleymichal:
oh even with quotes this doesn't work. unfortunate. unlikely that folks will do it but if there's a convention we could use that would be ideal.

EverlastingBugstopper:
This has been sitting for a while - do we want to get this in or just let wrangler dev handle this case?

EverlastingBugstopper:
I think we should close this and recommend that people use `wrangler dev` for this case cc @ashleymichal

ashleymichal:
@EverlastingBugstopper seems good

PR #956, opened 2019-12-13
Use indicatif to provide a more intuitive representation of upload progress

### LINKED ISSUE #906: Use indicatif for output
Use https://github.com/mitsuhiko/indicatif when uploading KV/Workers Sites

gabbifish:
This PR closes #906. It uses the indicatif crate to represent progress in KV uploads for workers sites. This provides more context to upload progress than `Uploading...` messages.

Now uploads look like
```console
$ wrangler publish
 Using namespace for Workers Site "__blog-workers_sites_assets"
 Uploading site files
████████████████████████████████████████████████████████████████████████ 48/48
 Success
⬇️ Installing wasm-pack...
 Built successfully, built project size is 11 KiB.
 Successfully published your script to https://blog.ziggy.workers.dev

```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -312,6 +312,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "console"
+version = "0.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "clicolors-control 1.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "encode_unicode 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "lazy_static 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.65 (registry+https://github.com/rust-lang/crates.io-index)",
+ "regex 1.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "termios 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "unicode-width 0.1.6 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


gabbifish:
It looks like you were using a pre-existing namespace with all the site files already uploaded (make sure to delete existing namespaces if you're testing this functionality!)

As for the full progress bar, this is because Wrangler already uploads up to 5000 files (or max 50MB files) in one go. You're uploading a tiny site, it seems, so the upload happens in one call. Indicatif will be more useful for our users uploading way more substantial sites.

gabbifish:
Something I can do, though, is not use the indicatif bar if the number of pairs being uploaded == 0. I'll get on that.

EverlastingBugstopper:
I'm wondering if it's better to not display the bar for sites under 5k files. It seems strange to have a progress bar that's always entirely filled. Definitely think it's useful for larger sites though. Will it jump in increments of 5000 because even that might look a little weird.. interested in other opinions about this!

gabbifish:
I like the idea of not displaying the bar for small uploads. This is more preferable than breaking up the API calls into unnecessarily small batches for the sake of a nice looking indicatif bar.

And yes, the bar will jump in batches of 5000 (or whatever files amount to the 50MB payload max for uploads). I don't think that would look weird.

EverlastingBugstopper:
I wonder if it would be worth doing just like a little tiny 1ms sleep for each step in the 5000 or something... /how hard that would be

gabbifish:
That assumes that there's a certain duration for the API call, which is not the case. I'm strongly in favor of displaying accurate information to Wrangler users, even if it's not "pretty". I'd rather peg indicatif monitors to accurate information.

PR #957, opened 2019-12-13
Alewis/custom webpack config for sites

### LINKED ISSUE #905: allow custom webpack config for Workers Sites projects
```
$ wrangler build
⚠️  Workers Sites does not support custom webpack configuration files
✨  Built successfully, built project size is 11 KiB.
```

Thsi error message is historical; when wrangler implicitly looked at the root for a custom webpack config, it made sense to disable the functionality for sites since the root would likely contain a config for the app you were deploying, not for the workers site. with this change, though, we should allow custom webpack configs since you are required to declare them in your wrangler.toml. This should be as simple as removing a conditional somewhere in wranglerjs, we just need to triage it.

_Originally posted by @ashleymichal in https://github.com/cloudflare/wrangler/pull/847#issuecomment-558234664_

ashleymichal:
fixes #905 . need a followup PR for documenting the expected use of these, because it is not immediately obvious to someone not familiar with webpack (like me).

i also want another followup PR to warn when wranglerjs fails with `Entry module not found`, and a user has a site AND a custom webpack config, to tell them to set the context like it is here.


tests/build.rs
```diff
@@ -14,12 +14,40 @@ fn it_builds_webpack() {
...
+}
+
+#[test]
+fn it_builds_webpack_site_with_custom_webpack() {
+    let fixture = Fixture::new_site();
+
+    fixture.create_file(
+        "workers-site/webpack.worker.js",
+        r#"
+        module.exports = { context: __dirname, entry: "./index.js" };
```
EverlastingBugstopper: should `target: webworker` in case somebody tries to copy our test suite instead of looking at our docs :)

ashleymichal: i... O.o

tests/fixture/mod.rs
```diff
@@ -14,28 +14,39 @@ use toml;
...
         eprintln!("Created fixture at {}", dir.path().display());
         Fixture {
             dir: ManuallyDrop::new(dir),
+            output_path: BUNDLE_OUT,
         }
     }
 
+    pub fn new_site() -> Fixture<'static> {
+        let mut fixture = Fixture::new();
+        fixture.output_path = "dist";
```
EverlastingBugstopper: wouldn't this be `/workers-site/worker`?

ashleymichal: heh, we actually output to dist, also. but probably long term this is what we want to assert. will change.

tests/build.rs
ashleymichal: should we do that for all of them though? because none of the rest of our tests do this.

ashleymichal: should we do that for all of them though? because none of the rest of our tests do this.

ashleymichal:
tested with mdx-deck workers sites project with custom webpack to not minify. can confirm this works as expected.

tests/build.rs
EverlastingBugstopper: hmm yeah we probably should... if we dont provide a custom config then it should be using it automatically

PR #963, opened 2019-12-16
[WIP] [dev] Attempt to send cf-ew-raw-host

### LINKED ISSUE #926: [dev] pass host header to preview service
instead of overwriting it, prepend cf-ew-raw-

also check to see if you can curl `subdomain.localhost:8000/test`

EverlastingBugstopper:
Attempt to fix #926  

The preview service doesn't seem to care if I send a `cf-ew-raw-host` header and instead uses the host that we've passed to the preview service when we create a new session.

### DIFF
--- a/src/commands/dev/headers.rs
+++ b/src/commands/dev/headers.rs
@@ -1,6 +1,8 @@
+use crate::commands::dev::server_config::ServerConfig;
+
 const HEADER_PREFIX: &str = "cf-ew-raw-";
 
-use hyper::header::{HeaderMap, HeaderName};
+use hyper::header::{HeaderMap, HeaderName, HeaderValue};
 use hyper::http::request::Parts as RequestParts;
 use hyper::http::response::Parts as ResponseParts;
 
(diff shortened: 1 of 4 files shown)


src/commands/dev/headers.rs
```diff
@@ -20,15 +22,29 @@ pub fn strip_response_headers_prefix(parts: &mut ResponseParts) -> Result<(), fa
     Ok(())
 }
 
-pub fn prepend_request_headers_prefix(parts: &mut RequestParts) {
+pub fn prepend_request_headers_prefix(parts: &mut RequestParts, server_config: &ServerConfig) {
     let mut headers: HeaderMap = HeaderMap::new();
 
     for header in &parts.headers {
         let (name, value) = header;
+        let value = if name == "host" {
```
EverlastingBugstopper: while iterating over all of the headers and prefixing them with `cf-ew-raw-`, if it finds the `host` header, replace `localhost:8787` with whatever they passed in as the host. 

So if they ran `wrangler dev --host example.com --ip localhost --port 8787` and then ran `curl subdomain.localhost:8787`, it would find the incoming `host: subdomain.localhost:8787` and change it to `cf-ew-raw-host: subdomain.example.com`

This doesn't seem to behave as expected as `request.url` still seems to be `example.com` instead of `subdomain.example.com`.

EverlastingBugstopper:
this will not work - closing in favor of future improvements to come after realistic preview

PR #964, opened 2019-12-16
[wip] [dev] Disallow localhost as backend

### LINKED ISSUE #902: [dev] disallow localhost as host
for now it rick rolls you so we should limit that behavior until we solve #901

EverlastingBugstopper:
Fixes #902

```console
$ wrangler dev -h http://localhost:8000
Error: You cannot currently use wrangler dev with a local server as the host. You could use TryCloudflare to generate a public URL that you could then pass as the host to wrangler dev.
See https://developers.cloudflare.com/argo-tunnel/trycloudflare/ for more info
```

This approach is a bit naive at the moment, there are other endpoints that the preview service will return a rick roll embed on.


gabbifish:
If we need to do this, I think we should be at parity with what capn proto checks for in `restrict_peers()`

PR #970, opened 2019-12-17
[WIP] [dev] use native_tls instead of openssl

EverlastingBugstopper:
```
[2019-12-17T21:09:33Z ERROR ws::handler] WS Error <Internal>: Failed to upgrade client to SSL: the handshake process was interrupted
```

[waiting on a response](https://github.com/housleyjk/ws-rs/issues/306) from the friendly `ws-rs` folks

PR #972, opened 2019-12-17
[dev] Disable console log messages for windows

EverlastingBugstopper:
tempfix until https://github.com/cloudflare/wrangler/pull/970 gets figured out


EverlastingBugstopper:
we should just prioritize #970

PR #981, opened 2020-01-03
Fix #958: Do not factor in .gitignore into workers sites upload directory traversal

### LINKED ISSUE #958: Workers Sites uploads do not include files that are in .gitignore
#### Environment

* operating system: windows 10 pro v1909 (18363.535) 

#### Steps to reproduce
Generate workers site project, set account id, api access granted using workers site template permissions. 

Static file export of Wordpress site, placed in ./public folder

Run wrangler publish 

Some css and js and font files are not uploaded (as seen by http 500 error in chrome dev tools for those resources)

I suspect this is because the files are in a deep folder structure, because when these files are moved to root public folder or 1 folder deep, wrangler does not skip them and they are successfully deployed.

I expect ALL files in ./public folder to be deployed
Some files are skipped during wrangler publish process

gabbifish:
Fixes #958. This change ensures that the wrangler include/exclude logic for workers sites bucket directory traversal does NOT take into account .gitignore. A test has been added to ensure this functionality exists.

### DIFF
--- a/src/commands/kv/bucket/mod.rs
+++ b/src/commands/kv/bucket/mod.rs
@@ -140,7 +140,10 @@ fn get_dir_iterator(target: &Target, directory: &Path) -> Result<Walk, failure::
     };
 
     let ignore = build_ignore(target, directory)?;
-    Ok(WalkBuilder::new(directory).overrides(ignore).build())
+    Ok(WalkBuilder::new(directory)
+        .git_ignore(false)
+        .overrides(ignore)
+        .build())
 }
 
 fn build_ignore(target: &Target, directory: &Path) -> Result<Override, failure::Error> {
@@ -255,6 +258,7 @@ mod tests {
     use super::*;
     use regex::Regex;
     use std::fs;
+    use std::io::Write;
     use std::path::{Path, PathBuf};
 
     use crate::settings::toml::{Site, Target, TargetType};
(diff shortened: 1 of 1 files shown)


src/commands/kv/bucket/mod.rs
```diff
@@ -439,6 +443,49 @@ mod tests {
...
+            .map(|entry| entry.unwrap().path().to_owned())
+            .collect();
+
+        assert!(files.contains(&test_path));
+
+        // Why drop()? Well, fs::remove_dir_all on Windows depends on the DeleteFileW syscall.
+        // This syscalldoesn't actually delete a file, but only marks it for deletion. It still
+        // can be alive when we try to delete the parent directory test_dir, causing a "directory
+        // is not empty" error. As a result, we MUST call drop() to close the gitignore file so
+        // that it is not alive when fs::remove_dir_all(test_dir) is called.
```
ashleymichal: 👍

src/commands/kv/bucket/mod.rs
```diff
@@ -439,6 +443,49 @@ mod tests {
...
+
+        let files: Vec<_> = get_dir_iterator(&target, Path::new(&upload_dir))
+            .unwrap()
+            .map(|entry| entry.unwrap().path().to_owned())
+            .collect();
+
+        assert!(files.contains(&test_path));
+
+        // Why drop()? Well, fs::remove_dir_all on Windows depends on the DeleteFileW syscall.
+        // This syscalldoesn't actually delete a file, but only marks it for deletion. It still
```
gabbifish:
```suggestion
        // This syscall doesn't actually delete a file, but only marks it for deletion. It still
```

PR #983, opened 2020-01-06
feat(whoami): return accounts and account ids on whoami

### LINKED ISSUE #630: whoami should display `account_id`

The `account_id` is rather hard to find. It requires you look at the URL on `dash.cloudflare.org`.   At the same time, `account_id` is required to publish anything. 

It would be nice if the `whoami` command displayed the account id in addition to the email.

#### Basic example

```bash
wrangler whoami
💁  👋  You are logged with the email 'ocudigitaltech@gmail.com'. Your account_id is 'b268fbd36acb17898f484b34d0d7'.

### PR DESCRIPTION
**Pull request by ashleygwilliams**
fixes #630 

Screen Shot 2020-01-07 at 12 07 16 PM
Screen Shot 2020-01-07 at 12 09 48 PM
Screen Shot 2020-01-07 at 12 13 10 PM

### DIFF
--- a/src/commands/whoami/mod.rs
+++ b/src/commands/whoami/mod.rs
@@ -1,20 +1,53 @@
+use crate::http;
 use crate::settings::global_user::GlobalUser;
-use crate::terminal::{emoji, message};
+use crate::terminal::emoji;
+
+use cloudflare::endpoints::account;
+use cloudflare::endpoints::account::Account;
+use cloudflare::framework::apiclient::ApiClient;
+use cloudflare::framework::HttpApiClientConfig;
+
+use prettytable::{Cell, Row, Table};
 
 pub fn whoami(user: &GlobalUser) -> Result<(), failure::Error> {
     // If using email + API key for auth, simply prints out email from config file.
-    let email: String = match user {
-        GlobalUser::GlobalKeyAuth { email, .. } => email.to_string(),
-        GlobalUser::TokenAuth { .. } => failure::bail!(
-            "At the moment, Wrangler cannot get user information for users using API tokens"
-        ),
+    let auth: String = match user {
+        GlobalUser::GlobalKeyAuth { email, .. } => {
+            format!("a Global API Key, associated with the email '{}'", email,)
+        }
+        GlobalUser::TokenAuth { .. } => format!("an API Token"),
     };
 
-    let msg = format!(
-        "{} You are logged in with the email '{}'.",
-        emoji::WAVING,
-        email
-    );
-    message::info(&msg);
+    println!("\n{} You are logged in with {}.\n", emoji::WAVING, auth,);
+    let accounts = fetch_accounts(user)?;
+    let table = format_accounts(user, accounts);
+    println!("{}", &table);
     Ok(())
+}
+
+fn fetch_accounts(user: &GlobalUser) -> Result<Vec<Account>, failure::Error> {
+    let client = http::cf_v4_api_client(user, HttpApiClientConfig::default())?;
+    let response = client.request(&account::ListAccounts { params: None });
+    match response {
+        Ok(res) => Ok(res.result),
+        Err(e) => failure::bail!(http::format_error(e, None)),
+    }
+}
+
+fn format_accounts(user: &GlobalUser, accounts: Vec<Account>) -> Table {
+    let mut table = Table::new();
+    let table_head = Row::new(vec![Cell::new("Account Name"), Cell::new("Account ID")]);
+    table.add_row(table_head);
+
+    if let GlobalUser::TokenAuth { api_token: _ } = user {
+        if accounts.is_empty() {
+            println!("Your token is missing the 'Account Settings: Read' permission.\n\nPlease generate and auth with a new token that has these perms to be able to list your accounts.\n");
+        }
+    }
+
+    for account in accounts {
+        let row = Row::new(vec![Cell::new(&account.name), Cell::new(&account.id)]);
+        table.add_row(row);
+    }
+    table
 }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/commands/whoami/mod.rs [hunk 1]
```diff
@@ -1,20 +1,56 @@
...
+    let response = client.request(&account::ListAccounts { params: None })?;
+
+    let mut table = Table::new();
+    let table_head = Row::new(vec![Cell::new("Account Name"), Cell::new("Account ID")]);
+    table.add_row(table_head);
+
+    match user {
+        GlobalUser::TokenAuth { .. } => {
+            if response.result.len() == 0 {
+                println!("Your token is missing the Account Settings: Read permission.\n\nPlease generate and auth with a new token that has these perms to be able to list your accounts.\n");
```
ashleymichal: is this really the only way you learn this? interesting.
are there any other error cases we need to account for in the response? and is it worth including at least info log lines for those errors?

ashleygwilliams: as far as i can tell yes

ashleygwilliams: i dont know what you mean by other errors

ashleymichal: like any non-200 response from the api. we have format_errors that will take a function to match against the code, but you could just pass it a function that prints the error itself.

https://github.com/cloudflare/wrangler/blob/master/src/http.rs#L91

src/commands/whoami/mod.rs
```diff
@@ -1,20 +1,56 @@
...
+fn fetch_accounts(user: &GlobalUser) -> Result<String, failure::Error> {
+    let client = http::cf_v4_api_client(user, HttpApiClientConfig::default())?;
+    let response = client.request(&account::ListAccounts { params: None })?;
+
+    let mut table = Table::new();
+    let table_head = Row::new(vec![Cell::new("Account Name"), Cell::new("Account ID")]);
+    table.add_row(table_head);
+
+    match user {
+        GlobalUser::TokenAuth { .. } => {
```
EverlastingBugstopper: nit: can we use `if let` syntax here instead of `match`

```rust
if let GlobalUser::TokenAuth { .. } = user {
```

src/commands/whoami/mod.rs
```diff
@@ -1,20 +1,56 @@
...
+            if response.result.len() == 0 {
```
EverlastingBugstopper: nit:

```suggestion
            if response.result.is_empty() {
```

src/commands/whoami/mod.rs
EverlastingBugstopper: nit: 

```suggestion
                println!("Your token is missing the "Account Settings: Read" permission.\n\nPlease generate and auth with a new token that has these perms to be able to list your accounts.\n");
```

src/commands/whoami/mod.rs
ashleymichal: i think the match is more appropriate, fwiw.

EverlastingBugstopper: it's only matching on one thing, i think generally `if let` fits that model better, it doesn't need an `else` and would eliminate the `_ => ()`

ashleygwilliams: can you say more about why you want if let?

ashleygwilliams: (there's no assignment here so i am less inclined)

src/commands/whoami/mod.rs
```diff
@@ -1,20 +1,53 @@
...
+}
+
+fn format_accounts(user: &GlobalUser, accounts: Vec<Account>) -> Table {
+    let mut table = Table::new();
+    let table_head = Row::new(vec![Cell::new("Account Name"), Cell::new("Account ID")]);
+    table.add_row(table_head);
+
+    if let GlobalUser::TokenAuth { api_token: _ } = user {
+        if accounts.is_empty() {
+            println!("Your token is missing the 'Account Settings: Read' permission.\n\nPlease generate and auth with a new token that has these perms to be able to list your accounts.\n");
```
EverlastingBugstopper: accessibility nit: auth -> authenticate, perms -> permissions

EverlastingBugstopper: also may be worth linking to the api token page if possible

src/terminal/emoji.rs
```diff
@@ -22,6 +22,7 @@ pub static EYES: Emoji = Emoji("👀 ", "");
 pub static FILES: Emoji = Emoji("🗂️ ", "");
 pub static INBOX: Emoji = Emoji("📥 ", "");
 pub static INFO: Emoji = Emoji("💁 ", "");
+pub static LIST: Emoji = Emoji("📋 ", "");
```
EverlastingBugstopper: is this used anywhere?

ashleygwilliams: i cannot figure out why i added this. i think i was thinking that the interface here should maybe be wrangler account list? i can remove it for now... but also should it be wrangler account list?

src/commands/whoami/mod.rs
```diff
@@ -1,20 +1,53 @@
...
+        GlobalUser::TokenAuth { .. } => format!("an API Token"),
     };
 
-    let msg = format!(
-        "{} You are logged in with the email '{}'.",
-        emoji::WAVING,
-        email
-    );
-    message::info(&msg);
+    println!("\n{} You are logged in with {}.\n", emoji::WAVING, auth,);
```
ashleymichal: why not use message? it's fine but i've been seeing this a lot lately, maybe our message module is not serving us well.

ashleygwilliams: ur not wrong. as said in air- the table would be offset if i used messaged because it would print INFO.

PR #984, opened 2020-01-06
[dev] Adds environment support to dev server

### LINKED ISSUE #962: [dev] support environments
`wrangler dev --env production` - in case there are different build steps/`webpack.config.js` files

EverlastingBugstopper:
Fixes #962 - `wrangler dev --env staging` now works

### DIFF
--- a/src/commands/dev/mod.rs
+++ b/src/commands/dev/mod.rs
@@ -17,7 +17,7 @@ use hyper::{Body, Client as HyperClient, Request, Response, Server, Uri};
 
 use hyper_tls::HttpsConnector;
 
-use tokio::runtime::{Runtime as TokioRuntime};
+use tokio::runtime::Runtime as TokioRuntime;
 
 use uuid::Uuid;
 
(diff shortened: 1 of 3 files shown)


src/commands/dev/mod.rs
```diff
@@ -17,7 +17,7 @@ use hyper::{Body, Client as HyperClient, Request, Response, Server, Uri};
 
 use hyper_tls::HttpsConnector;
 
-use tokio::runtime::{Runtime as TokioRuntime};
+use tokio::runtime::Runtime as TokioRuntime;
```
EverlastingBugstopper: this is an artifact of `cargo fmt`

src/main.rs
```diff
@@ -352,6 +352,13 @@ fn run() -> Result<(), failure::Error> {
                     "{} Start a local server for developing your worker",
                     emoji::EAR
                 ))
+                .arg(
+                    Arg::with_name("env")
+                        .help("environment to build")
+                        .short("e")
+                        .long("env")
+                        .takes_value(true)
+                )
```
EverlastingBugstopper: i already had the `.matches("env")` check and target creation logic, just needed to add the clap argument :)

PR #987, opened 2020-01-06
Add List Routes/Delete Route commands

ashleymichal:
these two commands are handy for unblocking users with complex routing tables.

### DIFF
--- a/src/commands/route/mod.rs
+++ b/src/commands/route/mod.rs
@@ -1,0 +1,54 @@
+extern crate serde_json;
+
+use crate::http;
+
+use cloudflare::endpoints::workers::{DeleteRoute, ListRoutes};
+use cloudflare::framework::apiclient::ApiClient;
+use cloudflare::framework::HttpApiClientConfig;
+
+use crate::settings::global_user::GlobalUser;
+use crate::settings::toml::Target;
+use crate::terminal::message;
+
+pub fn list(target: &Target, user: &GlobalUser) -> Result<(), failure::Error> {
+    let client = http::cf_v4_api_client(user, HttpApiClientConfig::default())?;
+
+    let result = client.request(&ListRoutes {
+        zone_identifier: &target
+            .zone_id
+            .as_ref()
+            .expect("missing zone_id in `wrangler.toml`"),
+    });
+
+    match result {
+        Ok(success) => {
+            let routes = success.result;
+            println!("{}", serde_json::to_string(&routes)?);
+        }
+        // TODO: handle route errors with http::format_error
+        Err(e) => failure::bail!("{}", e),
+    }
+    Ok(())
+}
+
+pub fn delete(target: &Target, user: &GlobalUser, route_id: &str) -> Result<(), failure::Error> {
+    let client = http::cf_v4_api_client(user, HttpApiClientConfig::default())?;
+
+    let result = client.request(&DeleteRoute {
+        zone_identifier: &target
+            .zone_id
+            .as_ref()
+            .expect("missing zone_id in `wrangler.toml`"),
+        identifier: route_id,
+    });
+
+    match result {
+        Ok(success) => {
+            let msg = format!("Successfully deleted route with id {}", success.result.id);
+            message::success(&msg);
+        }
+        // TODO: handle route errors with http::format_error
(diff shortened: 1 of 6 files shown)


src/commands/route/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+    let result = client.request(&ListRoutes {
+        zone_identifier: &zone_identifier,
+    });
+
+    match result {
+        Ok(success) => {
+            let routes = success.result;
+            println!("{}", serde_json::to_string(&routes)?);
+        }
+        // TODO: handle route errors with http::format_error
```
gabbifish: This will be super nice! Do you plan to get this in for this PR or are you holding off until later?

ashleymichal: yah that should go in this PR.

src/main.rs
```diff
@@ -552,6 +576,38 @@ fn run() -> Result<(), failure::Error> {
...
+
+        match route_matches.subcommand() {
+            ("list", Some(_)) => {
+                commands::route::list(zone_id?, &user)?;
+            }
+            ("delete", Some(delete_matches)) => {
+                let route_id = delete_matches.value_of("route_id").unwrap();
+                commands::route::delete(zone_id?, &user, route_id)?;
+            }
+            ("", None) => message::warn("route expects a subcommand"),
```
gabbifish: technically this should never be reached given that `.setting(AppSettings::SubcommandRequiredElseHelp)` is used above. Think we can delete this line?

src/main.rs
```diff
@@ -552,6 +576,38 @@ fn run() -> Result<(), failure::Error> {
...
+
+        let zone_id: Result<String, failure::Error> = if let Some(zone_id) = env_zone_id {
+            Ok(zone_id.to_string())
+        } else if let Some(zone_id) = manifest.zone_id {
+            Ok(zone_id)
+        } else {
+            failure::bail!(
+                "You must specify a zone_id in `wrangler.toml` to use `wrangler route` commands."
+            )
+        };
```
gabbifish: this could honestly be a really nice Manifest method that we call here?

src/commands/route/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+    let client = http::cf_v4_api_client(user, HttpApiClientConfig::default())?;
+
+    let result = client.request(&ListRoutes {
+        zone_identifier: &zone_identifier,
+    });
+
+    match result {
+        Ok(success) => {
+            let routes = success.result;
+            println!("{}", serde_json::to_string(&routes)?);
```
gabbifish: Huh! looks like this API call doesn't use pagination in its results, which surprised me :D Printing this out as vanilla JSON should work just fine then (no need to consolidate multiple pages of results to print them out).

src/main.rs
ashleymichal: this is copied from the kv subcommands, which were copied from the clap documentation. atm i'm kinda just going on what i know works, i'm not 100% certain why they specify this.

src/main.rs
ashleymichal: it's a one-off, so i'm inclined not to shove this into Manifest, or at minimum to wait until it repeats. It already kinda bothers me that deploy target is there, and honestly i think if i go for refactors i will turn that method inside out.

src/commands/route/mod.rs
ashleymichal: is this a call to action? 😅 if so idk what it is

src/commands/route/mod.rs
ashleymichal: ah i remember why this was gnarly. i wanted to refactor the logic of extracting an error out of `http::format_error` into its own fn, so that this doesn't break json parsers like jq when it errors.

src/commands/route/mod.rs
gabbifish: no call to action! just an observation :)

src/main.rs
gabbifish: ok! we can keep this for now for consistency, though this is something we can definitely add to a clean up effort when it arrives.

src/commands/route/mod.rs
ashleymichal: done

PR #988, opened 2020-01-07
Fix installs on Windows

EverlastingBugstopper:
windows did not like my fancy `node -e require`

### DIFF
--- a/npm/install-wrangler.js
+++ b/npm/install-wrangler.js
@@ -1,0 +1,4 @@
+#!/usr/bin/env node
+
+const { install } = require("./binary");
+install()
--- a/npm/uninstall-wrangler.js
+++ b/npm/uninstall-wrangler.js
@@ -1,0 +1,4 @@
+#!/usr/bin/env node
+
+const { uninstall } = require("./binary");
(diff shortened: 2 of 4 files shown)


npm/install-wrangler.js
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env node
+
+const { install } = require("./binary");
+install()
```
ashleymichal: no EOL i guess makes gh grumpy

npm/uninstall-wrangler.js
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env node
+
+const { uninstall } = require("./binary");
+uninstall()
```
ashleymichal: same

EverlastingBugstopper:
@ashleymichal i used the `-e` so i wouldn't have to break things out into separate files that are two lines long lol

npm/install-wrangler.js
EverlastingBugstopper: gh can have a little eol as a treat

Issue #888, opened 2019-11-20
[dev] "invalid preview session" after long period of inactivity


EverlastingBugstopper:
The proxy will need to re-create a preview session if the current one expires

EverlastingBugstopper:
this may be solved by #991

PR #991, opened 2020-01-08
[dev] fix windows builds and timeouts

EverlastingBugstopper:
This fixes the issues we were having with logs not working on windows!


EverlastingBugstopper:
keepalive timer requires that we send a heartbeat message every so often and i'm gonna need `tokio-tungstenite` for that. probably for the best as it seems to be the recommended websocket library

EverlastingBugstopper:
One thing I have a question about is my use of two separate tokio runtimes here. I'm wondering if there's a better pattern I should be using.

EverlastingBugstopper:
I think this might fix #888

steveklabnik:
Yeah, two runtimes aren't ideal. I haven't dug into the code yet but https://docs.rs/tokio/0.2.9/tokio/task/fn.spawn.html is probably what you'd do

EverlastingBugstopper:
So it turns out this builds on Windows but seems to not read incoming socket messages. I've opened [this issue](https://github.com/snapview/tokio-tungstenite/issues/77) to see if the awesome folks over at `tokio-tungstenite` have any ideas.

EverlastingBugstopper:
@steveklabnik re: your comment about `spawn`

The following code starts and ends the process almost immediately, it seems to spawn the two tasks and then exit.

```rust
    let mut runtime = TokioRuntime::new().unwrap();

    // listen for devtools messages
    runtime.spawn(async move {
        socket::listen(session_id).await
    });

    // handle incoming HTTP requests
    runtime.spawn(async move {
        serve(server_config, preview_id).await
    })?;
```

The following code seems to work at first, provides a "Listening" message, but fails to handle any incoming HTTP requests:

```rust
    let mut runtime = TokioRuntime::new().unwrap();

    // listen for devtools messages
    runtime.spawn(async move {
        socket::listen(session_id).await
    });

    // handle incoming HTTP requests
    runtime.block_on(async move {
        serve(server_config, preview_id).await
    })?;
```

steveklabnik:
I believe that's because of this: 
image

that is, your runtime goes out of scope at the end of the function, and so shuts down.

EverlastingBugstopper:
I seem to have cracked the code thanks to some help from the friendly Tokio discord server :) - only one runtime now

src/commands/dev/mod.rs
```diff
@@ -43,12 +41,16 @@ pub fn dev(
...
-    thread::spawn(move || socket::listen(session_id));
-
-    // spawn tokio runtime on the main thread to handle incoming HTTP requests
     let mut runtime = TokioRuntime::new()?;
-    runtime.block_on(serve(server_config, preview_id))?;
+
+    let devtools_listener = socket::listen(session_id);
+    let server = serve(server_config, preview_id);
+
+    let runners = futures::future::join(devtools_listener, server);
```
steveklabnik: 🎊

src/commands/dev/mod.rs
```diff
@@ -43,12 +41,16 @@ pub fn dev(
...
+
+    runtime.block_on(async {
+        let (_, _) = runners.await;
```
steveklabnik: any reason to not 

    let _ =

instead?

steveklabnik: also, are you sure that ignoring all failures is what you want here?

src/commands/dev/socket.rs
```diff
@@ -0,0 +1,80 @@
...
+                }
+            }
+        })
+    };
+    pin_mut!(keep_alive_to_ws, print_ws_messages);
+    future::select(keep_alive_to_ws, print_ws_messages).await;
+    Ok(())
+}
+
+async fn keep_alive(tx: futures::channel::mpsc::UnboundedSender<Message>) {
```
steveklabnik: might be worth giving this a return type of `!`

src/commands/dev/socket.rs
```diff
@@ -0,0 +1,80 @@
...
+    let duration = Duration::from_millis(1000 * KEEP_ALIVE_INTERVAL);
+    let mut interval = time::interval(duration);
+
+    let mut id = 2;
```
steveklabnik: i am interested why this is 2. not because it's wrong, but it feels unusual! Might deserve a comment

src/commands/dev/mod.rs
EverlastingBugstopper: That's a good point - I'm going to add some `.unwrap()` calls to them. Not quite sure if there's a way for me to bubble up those failures through the async fn

src/commands/dev/socket.rs
EverlastingBugstopper: Interesting! I'm not familiar with this syntax and can't seem to find an example of a function returning `!` - would you mind elaborating?

src/commands/dev/socket.rs
EverlastingBugstopper: Fair point - the reason it's 2 is because we have already sent `id = 1` earlier to enable the runtime. Eventually I plan on making the ID handling automatic and in the chrome-devtools-rs library.

src/commands/dev/mod.rs
steveklabnik: pub fn block_on<F: Future>(&mut self, future: F) -> F::Output

should return that `Result` back out of `block_on`, I think.

src/commands/dev/socket.rs
steveklabnik: https://doc.rust-lang.org/stable/book/ch19-04-advanced-types.html#the-never-type-that-never-returns

steveklabnik: (`loop` has the type `!`, but since `!` coerces to any other type, it coerces to the return type of `()` here, which is why this code works the way it does)

EverlastingBugstopper: this is very cool

src/commands/dev/mod.rs
EverlastingBugstopper: very interesting

EverlastingBugstopper:
thanks steve!

PR #999, opened 2020-01-14
[dev] Live reload/code watching

EverlastingBugstopper:
So this works - 

```console
$ wrangler dev
👂  Listening on http://localhost:8787
💁  watching "./"
[2020-01-08 11:34:33] "GET example.com/ HTTP/1.1"
"hello, world! i have not edited a file"
🌀  Detected changes...
[2020-01-08 11:34:37] "GET example.com/ HTTP/1.1"
"hello, world! i have edited a file"
🌀  Detected changes...
[2020-01-08 11:34:55] "GET example.com/ HTTP/1.1"
unknown console event: "Script modified; context reset."
"hello, world! i have edited a file again"
```

The problem is that the console event `"Script modified; context reset"` only occurs after I issue another curl command, which makes me think that my thread logic is blocking somewhere that it shouldn't be. I'm not quite sure how that would be happening as I made sure to drop the mutex guards after i use them.

### DIFF
--- a/src/commands/dev/watch.rs
+++ b/src/commands/dev/watch.rs
@@ -1,0 +1,33 @@
+use std::mem;
+use std::sync::{mpsc, Arc, Mutex};
+
+use crate::commands;
+use crate::commands::dev::get_preview_id;
+use crate::commands::dev::server_config::ServerConfig;
+
+use crate::settings::global_user::GlobalUser;
+use crate::settings::toml::Target;
+
+pub fn watch_for_changes(
+    target: Target,
+    user: Option<GlobalUser>,
+    server_config: &ServerConfig,
+    preview_id: Arc<Mutex<String>>,
+    session_id: &str,
+    verbose: bool,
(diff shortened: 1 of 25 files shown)


src/commands/dev/watch.rs
```diff
@@ -0,0 +1,33 @@
...
+
+pub fn watch_for_changes(
+    target: Target,
+    user: Option<GlobalUser>,
+    server_config: &ServerConfig,
+    preview_id: Arc<Mutex<String>>,
+    session_id: &str,
+    verbose: bool,
+) -> Result<(), failure::Error> {
+    let (tx, rx) = mpsc::channel();
```
ashleymichal: what are tx and rx? we really gotta go for readability with code like this, and comments, even if it's just like "this is the boilerplate from the getting started for this package.

EverlastingBugstopper: this is standard notation for mpsc channels, but i'll change it to `sender and receiver` since it's more readable 😄

PR #1001, opened 2020-01-15
Added URL support for preview command

larkin-nz:
Hey team,

Thanks for all the great work you and everyone else has done so far!

I have added support a custom URL to be passed to the `preview` command.

When either a URL is passed using the flag `-u` or `--url` to the `preview` command the browser will open to the URL specified.

Example usage:

```bash
$ wrangler preview --url https://example.com/endpoint?id=random
```

In the case where `--headless` is also passed, the server will output the response of the request to the URL.

When this flag is not specified the application will continue to use the default `https://example.com`.

The protocol specified in the URL can be either HTTP or HTTPS.

In the case that neither protocol are used, or the specified URL is not a valid URL, the command will fallback to using the default as specified above.

I hope you can review and implement this ASAP as I know a number of developers would love this feature!

Thanks heaps


larkin-nz:
A side note relevant to this feature.

It seems that currently, the CloudflareWorkers preview server automatically appends a forward-slash to the final component of the `__ew_fiddle_preview` cookie (the domain). 

I guess this currently occurs as the server currently expects a domain rather than a domain + pathname + query.

For example, if the final component of the cookie value is `example.com/?query=10` the server will process it as `example.com/?query=10/`.

This doesn't break anything, but it would good if this trailing forward-slash was removed when not needed (i.e. when the final section of the URL is either a pathname or query).

A very minor change required, but obviously something which I cannot do myself haha.

Thanks again

larkin-nz:
Added tests now also!

src/commands/preview/mod.rs
```diff
@@ -43,9 +46,21 @@ pub fn preview(
...
+
+    let regex = Regex::new(r"((?:http|https)://)([^/\\]*)(.*)").unwrap();
+
+    if regex.is_match(url) {
+
+        let captures = regex.captures(url).unwrap();
+
+        protocol = captures.get(1).map_or("https://", |m| m.as_str());
+        domain = captures.get(2).map_or("example.com", |m| m.as_str());
+        path = captures.get(3).map_or("", |m| m.as_str());
```
EverlastingBugstopper: I think for this I'd recommend using the `Url` crate which will handle all of this parsing for you - https://docs.rs/url/2.1.1/url/struct.Url.html

You should be able to `let url = Url::parse(url);` and pass that to the preview function where you can then pull out `url.scheme()`, `url.domain()`, and `url.path()`

src/commands/preview/mod.rs
```diff
@@ -57,8 +72,8 @@ pub fn preview(
 
         if !headless {
             open_browser(&format!(
-            "https://cloudflareworkers.com/?wrangler_session_id={0}&wrangler_ws_port={1}&hide_editor#{2}:{3}{4}",
-            &session.to_string(), ws_port, script_id, https_str, preview_host,
+            "https://cloudflareworkers.com/?wrangler_session_id={0}&wrangler_ws_port={1}&hide_editor#{2}:{3}{4}{5}",
+            &session.to_string(), ws_port, script_id, protocol, domain, path
```
EverlastingBugstopper: The way the previewer works at cloudflareworkers.com is that it takes a protocol and a domain to spin up a session at a specific URL and then appends the path to that URL. This is likely why you're running into the issue with the trailing slash. Ideally, this URL would only include the protocol and the domain, not the path, and the path would be sent later in the `get` and `post` functions in `preview/mod.rs` in the `client.post(PREVIEW_ADDRESS)` section.

larkin-nz: I see this wasn't correctly implemented in the `get` and `post functions in `preview/mod.rs` as they currently only send `PREVIEW_ADDRESS` and no `path` or `query`. I'll amend this and push up later today! Thanks

src/commands/preview/mod.rs
larkin-nz: Good thinking, I'll work on this now!

larkin-nz: Resolved this by using the `Url` crate as mentioned.

The `preview` command now passes a `Url` through to the handler.

A new struct has been created called `RequestPayload` which accepts a `Url` alongside the `method` and `body`. 

This struct now handles this functionality alongside dynamically generating the cookie based on the `script_id`

src/commands/preview/mod.rs
larkin-nz: The new `RequestPayload` struct is now responsible for processing all params relating to the request. 

It contains a `service_url` which used by the `client_request` with the `cookie` to make a request and log to the terminal (replaces `PREVIEW_URL`).

e.g. `https://00000000000000000000000000000000.cloudflareworkers.com/something?query=true`

There is also a `browser_url` which is opened in the browser for preview.

e.g. `https://example.com/something?query=true`

It will dynamically generate the data including `cookie` when needed using the `script_id`.

The `cookie` now contains only the domain, and the path + query are appended correctly to the `service_url`.

e.g. `__ew_fiddle_preview=ff624eeb7647ef8ca8b967bd888ee50b9ae103cdd9ae474e8a12a35c4a4f70830example.com`

larkin-nz:
Great to hear about the `dev` feature, this will be very useful for sure!

I guess this will just be a form of proxy around the existing endpoint (https://00000000000000000000000000000000.cloudflareworkers.com) which sends the necessary cookie?

I was heavily involved in the development of [Cloudworker] (https://github.com/dollarshaveclub/cloudworker) (https://github.com/dollarshaveclub/cloudworker)` long back before `wrangler` came out and it would be great to see the workflow get much closer to this.

In some ways my own library [Restt-CLI](https://github.com/resttjs/restt-cli) (https://github.com/resttjs/restt-cli) was a very early precursor to `wrangler` well before [workers.dev](workers.dev) was even setup.

It's nice to be involved with improving workers where possible as I have since way back!

Anyways, I've patched what was mentioned above in the previous commit.

I have also added support now for the `--watch` flag to work correctly.

If the `--watch` flag is on then each request will be logged using the `client_request` function added in the `preview` command when running with `--headless`.

`client_request` is responsible for making a request using the reqwest `client` using the `method`, `body` and `request_payload`.

All changes are working correctly now as needed.

I hope you can review this and integrate it to the core as soon as possible.

Thanks again for all the great work!

larkin-nz:
@EverlastingBugstopper any idea when this will be reviewed and approved?

larkin-nz:
@EverlastingBugstopper - any reason this pull request is taking so long to be approved?

EverlastingBugstopper:
Hi @larkin-nz - I was on caretaker leave the past couple of days, taking some time to look at this now. Thanks for your patience.

src/commands/preview/request_payload.rs
```diff
@@ -0,0 +1,55 @@
...
+
+impl RequestPayload {
+    pub fn create(method: HTTPMethod, url: Url, body: Option<String>) -> RequestPayload {
+        let session = Uuid::new_v4().to_simple().to_string();
+        let https = if url.scheme() == "https" { 1 } else { 0 };
+
+        let protocol = if https == 1 { "https://" } else { "http://" }.to_string();
+        let domain = url.domain().unwrap().to_string();
+        let path = url.path().to_string();
+        let query = format!("?{}", url.query().unwrap_or(""));
```
EverlastingBugstopper: this line makes the default url send `https://example.com/?` which we don't want - I'd recommend changing this to 

```rust
let query = match url.query() {
  Some(query) => format!("?{}", query),
  None => "".to_string()
}
```

src/main.rs
```diff
@@ -503,6 +512,11 @@ fn run() -> Result<(), failure::Error> {
 
         let method = HTTPMethod::from_str(matches.value_of("method").unwrap_or("get"))?;
 
+        let default_url = "https://example.com";
+
+        let url = Url::parse(matches.value_of("url").unwrap_or(default_url))
+            .unwrap_or(Url::parse(default_url)?);
```
EverlastingBugstopper: We should change this `.unwrap_or` to be a `?` so that it fails if the url fails to parse. That way if the user passes something like `--url invalid-string`, they can fix their mistake without finding out that they input something incorrect before wrangler starts to rebuild

src/commands/preview/request_payload.rs
```diff
@@ -0,0 +1,55 @@
...
+    pub service_url: String,
+    pub body: Option<String>,
+}
+
+impl RequestPayload {
+    pub fn create(method: HTTPMethod, url: Url, body: Option<String>) -> RequestPayload {
+        let session = Uuid::new_v4().to_simple().to_string();
+        let https = if url.scheme() == "https" { 1 } else { 0 };
+
+        let protocol = if https == 1 { "https://" } else { "http://" }.to_string();
```
EverlastingBugstopper: nit: we should be able to just use `url.scheme()` where we use `protocol`, no?

larkin-nz: I'm going to keep this bit how it is as we use `protocol` and `https` differently.

`https` is used in the cookie to be sent to the workers domain

src/commands/preview/request_payload.rs
larkin-nz: Good point -- hadn't noticed this and have added this now!

src/main.rs
larkin-nz: I agree, have added validation here (including scheme validation as either `http` or `https`)

larkin-nz:
Hey @EverlastingBugstopper - I've added up those updates!

All is working as expected on Ubuntu and Windows locally, but obviously the Windows test is still failing as referenced in #1012 

I hope you're able to merge this request now!

Thanks

larkin-nz:
Excellent, thank you @EverlastingBugstopper! 

Hope you work out the issues soon. 

Thanks for maintaining the project!

larkin-nz:
Just added a quick patch to the logic.

Previously on startup, both the `client` and the user's web browser would perform a request to the `URL`.

This should not be the case, the request from `client` should only be made with `--headless`

EverlastingBugstopper:
Hey @larkin-nz - that's actually expected behavior. We still want the output of the request to show up in the terminal even if the browser is opened 😄. If you could revert to right before that change that would be 💯

larkin-nz:
This behaviour was already removed from by the first commit I made as this pull request because we don't want duplicate requests occuring. I think it would be better to pipe the output from the browser to the terminal using a socket if we do output it. Currently it will make two requests and this can be painful for testing things.

EverlastingBugstopper:
@ashleymichal do you have thoughts on removing the output when `wrangler preview` is not run with the `--headless` argument? I'm hesitant to change it because it changes the default behavior that people expect. I'm especially worried about folks on the Windows Linux Subsystem who haven't configured their browser to open and are used to running `wrangler preview` and the output being there in their terminal.

@larkin-nz I'm interested in how duplicate requests makes it painful to test things - I agree it's not the most efficient but if you're testing can you just ignore the output in the terminal if you're not using `--headless`?

ashleymichal:
i can see how duplicate output can make testing difficult, though i had assumed (perhaps naively) that folks running automated tests would opt in to the `--headless` flag.

i'd be inclined to at least include instructions for those in the Windows Linux subsystem camp to take action (i.e. set their browser,  or use the `--headless` flag; if there's a better way we should be encouraging users to use that rather than supporting iffy behavior. That being said,  it could break folks' CI or development workflows and to me that suggests "breaking change", which deserves the rollout procedure that goes with such changes, i.e. deprecation warnings etc.

TL;DR -  I'm not opposed to making this change, but I think we should be courteous in how we release it.

larkin-nz:
@ashleymichal and @EverlastingBugstopper -- I can understand your logic here.

I guess an example of where this is a problem is a case where I have been making a CRON style task scheduler within Workers and KV using Streams.

When there are duplicate requests then the service ends up having multiple streams glitching out on different sessions while writing to KV at the same time due to the slow speeds of KV writing and reading. This can lead to a task in the schedule to be ran twice which is problematic in some cases.

Obviously this is a fairly specific case and is complex, but it still highlights the point of concern. Duplicate requests can cause a number of issue with testing things such as Stripe payments (Stripe will decline duplicates).



I guess where my logic sits is that if the `--headless` flag is used then the request should be performed from the terminal. In the case where it is not it should be performed from the browser as that is the whole point of the preview page. 

Headless would be for testing a request, where as standard preview would be exactly that, a interactive preview.

Would it not be easier to just change the way the preview site works so we can pipe back the response from there via a WebSocket so there are no duplicates. Maybe we can already do this without needing to change the preview site? I am unsure.

Let me know what you think!

EverlastingBugstopper:
@larkin-nz if you have time to implement piping output to the terminal from the browser by all means go ahead! 

I don't think we can merge this as is though due to the backwards compatibility issues previously discussed.

We are also planning on releasing a beta version of `wrangler dev` soon™️ which may satisfy your testing needs in a more ergonomic manner.

larkin-nz:
@EverlastingBugstopper - I'll revert the changes to how it was so you can
merge this into the main branch.

I'll then work on a new new pull later next week when I get the chance.

Cheers

EverlastingBugstopper:
@larkin-nz - in the meantime, we've released the beta! If you want to try it out you can install with `npm i -g @cloudflare/wrangler@beta` and check out the [testing instructions](https://github.com/cloudflare/wrangler/releases/tag/v1.8.0-rc.0).

larkin-nz:
@EverlastingBugstopper - great work will check this out soon! About to jump on another international flight but I'll add up the changes once I'm settled down again in the next couple days ☺️

larkin-nz:
@EverlastingBugstopper - sorry, have had a family emergency and haven't been online. Will aim to push up this stuff early next week. Cheers

EverlastingBugstopper:
Hey no worries at all 😄

larkin-nz:
Hey @EverlastingBugstopper & @ashleymichal 

Sorry for the delays (my brother passed away).

I've reverted the logic to be correct for those requests.

I'll try to add some logic as discussed above with WebSockets at a later date.

For now, this is all working as expected so this request can be merged!

Keep up the good work

src/commands/preview/mod.rs
```diff
@@ -57,47 +58,37 @@ pub fn preview(
...
         if !headless {
             open_browser(&format!(
-            "https://cloudflareworkers.com/?wrangler_session_id={0}&wrangler_ws_port={1}&hide_editor#{2}:{3}{4}",
-            &session.to_string(), ws_port, script_id, https_str, preview_host,
-        ))?;
+                "https://cloudflareworkers.com/?wrangler_session_id={0}&wrangler_ws_port={1}&hide_editor#{2}:{3}",
+                session, ws_port, script_id, browser_url
+            ))?;
+
+            message::preview("Your Worker is a Workers Site, please preview it in browser window.");
```
EverlastingBugstopper, line 65: This logic isn't quite right

```
        let msg = if sites_preview {	
            "Your Worker is a Workers Site, please preview it in browser window.".to_string()	
        } else {	
            format!("Your Worker responded with: {}", worker_res)	
        };
```

should be printing this output only if it's a workers site

larkin-nz: Presumably it came from a different version of the master when I pulled from upstream. Will patch that now.

larkin-nz:
@EverlastingBugstopper 

All done now -- ready to be merged! 

Cheers

PR #1003, opened 2020-01-17
dont reinstall same binary versions

### LINKED ISSUE #768: Installing on every change wasm-pack / wranglerjs
### 🐛 Bug Report

When i run `wrangler preview --watch` it's installing on every change those packages. Is that normal? It takes lot of time on every change.

#### Environment

* operating system: NixOs
* output of `rustc -V`: Is not installed
* output of `node -v`: v12.5.0
* output of `wrangler -V`: wrangler 1.4.0

#### Steps to reproduce

Empty new Project and then preview it with watch

Faster loading without installing on every change that package

Taking time on every change to install that package

EverlastingBugstopper:
Fixes #768 

This relies on [this PR](https://github.com/rustwasm/binary-install/pull/7) in @rustwasm/binary-install


src/install/mod.rs
```diff
@@ -53,52 +54,53 @@ pub fn install_artifact(
...
+                let version = Version::parse(version);
+                if let Ok(version) = version {
+                    let install = Some((entry.path(), version));
+                    // if let Some((_, latest_version)) = result {
+                    //     if version > latest_version {
+                    //         result = install;
+                    //     }
+                    // } else {
+                    //     result = install;
+                    // }
```
EverlastingBugstopper: this section is giving me trouble because 

```console
$ cargo check 
    Checking wrangler v1.6.0 (/Users/averyharnish/Documents/work/wrangler)
error[E0382]: use of moved value
  --> src/install/mod.rs:83:37
   |
83 |                     if let Some((_, latest_version)) = result {
   |                                     ^^^^^^^^^^^^^^ value moved here, in previous iteration of loop
   |
   = note: move occurs because value has type `semver::version::Version`, which does not implement the `Copy` trait

error[E0382]: borrow of moved value: `version`
  --> src/install/mod.rs:84:28
   |
81 |                 if let Ok(version) = version {
   |                           ------- move occurs because `version` has type `semver::version::Version`, which does not implement the `Copy` trait
82 |                     let install = Some((entry.path(), version));
   |                                                       ------- value moved here
83 |                     if let Some((_, latest_version)) = result {
84 |                         if version > latest_version {
   |                            ^^^^^^^ value borrowed here after move

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0382`.
error: could not compile `wrangler`.

To learn more, run the command again with --verbose.
```

EverlastingBugstopper:
@steveklabnik if you have time could you take a look? really appreciate the straightforward API you've provided with the `semver` crate 👍

src/install/mod.rs
steveklabnik: try 

```
83 |                     if let Some((_, latest_version)) = &result {

```

EverlastingBugstopper: i think i got it - only one clone now

EverlastingBugstopper:
This relies on [this PR](https://github.com/rustwasm/binary-install/pull/7) in @rustwasm/binary-install

EverlastingBugstopper:
this is good to go once a new version of binary-install is released. going to open another issue for some more work i'd like to do but it depends on work in #1190 so it should be ok to go through with this as is

EverlastingBugstopper:
binary-install has been published so this is ready for review. really excited for this to land.. finally

PierBover:
Awesome! Thanks @EverlastingBugstopper I will try the binary once it's released.

Cargo.toml
```diff
@@ -66,3 +67,6 @@ predicates = "1.0.2"
 
 [features]
 vendored-openssl = ['openssl/vendored']
+
+[profile.release]
+debug = true
```
ashleymichal, line 72: what's  this mean?

EverlastingBugstopper: oh i was using this to debug, great great catch.

background: wranglerjs doesn't download a binary if you're in debug mode so i was building in release mode but still wanted some sorta debug annotation i think? honestly i added this so long ago i kinda forget why i added it

ashleymichal: okay i thought it smelled funny 😀

Issue #301, opened 2019-07-04
Does wrangler support a corporate proxy?


lawred2:
Hi,

I attempted 'wrangler preview' and got the following

`Error: Error(Hyper(Error(Connect, Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })), "https://cloudflareworkers.com/script")`

I am using linux behing a corp proxy and have set the usual environment variables..

Is it possible that wrangler is ignoring these variables?

Regards,
Dave

gabbifish:
Hi! Some questions to help us investigate:
1. Do you mind sharing which environment variables you've set? 
2. Are you able to send a GET request to https://cloudflareworkers.com?

michaellwest:
I started off with an issue installing wrangler. Was able to download the package manually and extra to here: `C:\Users\michael.west\.wrangler\out`.

Now I see an error, which I assume is proxy related, when using the generate command.

> PS C:\Work\cloudflare> wrangler generate my-worker
Error: https://crates.io/api/v1/crates/cargo-generate: error trying to connect: An existing connection was forcibly closed by the remote host. (os error 10054)

lawred2:
> Hi! Some questions to help us investigate these connectivity issues:
> 
> 1. Do you mind sharing which environment variables you've set?
> 2. Are you able to send a GET request to https://cloudflareworkers.com?
> 
> With this information I can look into this issue more in-depth.

Hi Gabbi,

Forgive my tardiness replying. Curse of distractions. Back to this now...

$http_proxy and $https_proxy are set to our corporate proxy. These are squid proxies with no authentication

See screengrab for terminal output

![image](https://user-images.githubusercontent.com/4548982/68077416-52b7ce00-fdbb-11e9-9d3b-a35eb13288c8.png)

As you can see we have no issue hitting https://cloudflareworkers.com

Also to further illustrate we have no issue hitting https://crates.io/api/v1/crates/cargo-generate either through the proxy

![image](https://user-images.githubusercontent.com/4548982/68077429-9b6f8700-fdbb-11e9-9980-7366ea550d37.png)

It's just wrangler. Doesn't seem to be proxy aware to me.

lawred2:
Hi Cloudflare - any progress on this?

gabbifish:
https://github.com/cloudflare/wrangler/pull/1009 will fix this problem--reqwest 0.10.1 introduced proxy env var support for its HTTP client, which is exactly what we needed.

[...]

PR #1009, opened 2020-01-22
Update cloudflare-rs, reqwest, http, uuid

ashleymichal:
had to update cloudflare-rs and it sort of...spiraled. but now we're on the latest reqwest so that's cool. it looks like a big change but it's mostly in the Cargo.lock.

this is critical for work on Config Variables, as all the relevant endpoints will be added to cloudflare-rs post-reqwest bump.


ashleymichal:
atm this is failing in CI on windows. it hangs on two specific tests there. everything works locally on the windows machine we have here in the office. idk what that's about.

gabbifish:
This PR, by bumping to reqwest 0.10.1, also solves https://github.com/cloudflare/wrangler/issues/301. Yay! This reqwest version is the first to support corporate proxy env vars out of the box.

PR #1018, opened 2020-01-30
feat: configurable binary host url

noherczeg:
In controlled environments usually the case is that binaries besides the NPM packages could not pass through or it's cumbersome to obtain them. This change allows developers to specify the base URL where the binaries are stored.


EverlastingBugstopper:
@noherczeg this seems fine to me, but do you think you could change the name of the environment variable to `WRANGLER_BINARY_HOST` and then we can merge this in 😄

noherczeg:
@EverlastingBugstopper Updated per request.

PR #1037, opened 2020-02-07
more github actinos

EverlastingBugstopper:
GitHub Actinos status update

# *Tests now finish in ~4 minutes* on cache hits

- [x] caching cross-platform tests on stable and nightly - every push
- [x] caching cross-platform debug builds on stable - every push
- [x] cross-platform release builds and tarballs created - tag pushes that start with `v`
- [x] cargofmt - every push
- [x] create wranglerjs tarball - every push or just tag pushes?
- [x] create github releases - on release

Open questions:

- Since this builds a release on tag push, should we create them as pre-releases so that they go live when we say they go live or is it fine the way it is.
- Do we need to be targeting MacOS 10.7?
- Do tests compile all of our code/Do we need to build on every push if we're running tests on every push or should we just do the builds for releases?


EverlastingBugstopper:
this will create a github release when you push a tag that starts with `v`

Here's a screenshot of my fork's [release page](https://github.com/EverlastingBugstopper/wrangler/releases):

image

we should double check to make sure that each release binary works as expected.

cc @ashleygwilliams for review

ashleygwilliams:
>Since this builds a release on tag push, should we create them as pre-releases so that they go live when we say they go live or is it fine the way it is.

i feel fine about how it is now, but i'm happy to be persuaded otherwise

>Do we need to be targeting MacOS 10.7?

i would focus on the most up-to-date stable macOS only

>Do tests compile all of our code/Do we need to build on every push if we're running tests on every push or should we just do the builds for releases?

i think we should just build for releases

>this will create a github release when you push a tag that starts with v

the last setup would create a release with any named tag- do we want to change that behavior?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,19 @@
+name: Linters
```
ashleygwilliams: do we want clippy in here?

.github/workflows/release.yml
```diff
@@ -0,0 +1,214 @@
...
+          - build: windows
+            os: windows-latest
+            rust: stable
+          - build: wranglerjs
+            os: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Cache Cargo registry
```
ashleygwilliams: this logic is duplicated - is there anyway to pull it out into it's own thing and share it across the actions (not a blocker, just curious)

.github/workflows/release.yml
```diff
@@ -0,0 +1,214 @@
...
+        if: matrix.build == 'linux'
+        run: |
+          rustup target add ${{ env.LINUX_TARGET }}
+          cargo build --release --target ${{ env.LINUX_TARGET }} --features vendored-openssl
+
+      - name: Build (MacOS)
+        if: matrix.build == 'macos'
+        run: cargo build --release
+        env:
+          MACOSX_DEPLOYMENT_TARGET: 10.7
```
ashleygwilliams: how'd we pick this number? was it what we used previously?

.github/workflows/lint.yml
EverlastingBugstopper: we currently do not check clippy in CI and have a number of warnings. this would fail right now so i think this is worth tackling separately.

.github/workflows/release.yml
EverlastingBugstopper: i'm not sure! i wasn't able to find anything similar to azure pipelines templates but happy to use if it exists (perhaps @signalnerve knows?)

.github/workflows/release.yml
EverlastingBugstopper: this is what was in azure and why i asked if we needed it

.github/workflows/lint.yml
ashleygwilliams: got it! do we have an issue?

.github/workflows/release.yml
ashleygwilliams: got it. i don't think we need it! apple is particularly shitty about what counts as latest stable... based off of what still gets security updates (https://support.apple.com/en-us/HT201222), i think we should aim for 10.14 right now, but this feels like a guess on my part, so i'm open to thoughts you have!

.github/workflows/release.yml
ashleygwilliams: we can refactor, let's get this in and open an issue to investigate? could be a great way for @signalnerve to contribute to wrangler <3

.github/workflows/lint.yml
EverlastingBugstopper: #373 sorta

.github/workflows/release.yml
signalnerve: ahh hey i'm sorry i missed this notification! sounds interesting, would be down to collab with @EverlastingBugstopper on this in the future if you're down 👍

Issue #1023, opened 2020-01-31
Add a link to the current release page on wrangler --version


EverlastingBugstopper:
When a user runs `wrangler --version`, it should output the version, and on a new line it should output https://github.com/cloudflare/wrangler/releases/latest if the user is on the latest version, and the specific tagged release https://github.com/cloudflare/wrangler/releases/tag/v1.6.0 if they're on an older version

EverlastingBugstopper:
doing #1190 instead

Issue #397, opened 2019-08-05
Print update message on wrangler usage when new version of wrangler is available.


gabbifish:
To encourage wrangler users to use the latest version of the CLI, we should print out a message along the lines of 
```
A new version of wrangler (1.x.0) is available! Make sure to update by following the update instructions: https://github.com/cloudflare/wrangler#updating-wrangler
```
whenever a new version of wrangler is available. This could be printed out before every wrangler invocation if a user's wrangler version is outdated.

This version information could be pulled from github, crates.io, or npm; figuring out how to manage this without accidentally DOSing one of these endpoints is useful. (I know github rate limits calls to the API used to access repository tags, so we should keep this in mind).

EverlastingBugstopper:
Good starting point https://github.com/rustwasm/wasm-pack/blob/a267075be87d0744b2436c820c60b24524cda6a5/src/main.rs#L81-L84

PR #1039, opened 2020-02-07
Add link to latest release to long version.

hjarrell:
Hello I added a simple link to the latest GitHub release to the long version string. Please let me know if you would prefer a different message.

This is for issue #1023.


EverlastingBugstopper:
Hey @hjarrell - I think I should have been more specific in the issue that I filed. 

What we're looking for here is version-specific output, so if the version they're on is not the latest (say, version 1.6.0), it should print out a link to that release instead of the latest one.

As a side note, issue https://github.com/cloudflare/wrangler/issues/397 seems semi-related to this as well.

hjarrell:
Thanks for the feedback! I'll look at that issue and update the PR with a version-specific output.

EverlastingBugstopper:
solved by #1190

PR #994, opened 2020-01-09
🤫 Secrets commands

victoriabernard92:
Secrets create, list and delete commands all added.
Note includes:
* Alew's fix for cloudflare-rs http update https://github.com/cloudflare/wrangler/pull/994/commits/90dbe12ce34c33abe6e9bbbc55c36ff8af226382
* Moved KV::interactive_delete to utils to be used across commands

TODOs:

- [ ] Verify error codes and messages (e.g. message for `10007`)
- [ ] [Nice to have] Check 200 v. 201 and give appropriate messaging for create
- [x] Remove debug print outs from format_error

Issue #912, opened 2019-11-26
command to delete a secret from a worker


ashleymichal:
wrangler should support a command that allows the user to remove a secret environment variable from the script of their choosing. should confirm deletion before carrying out the command. should allow user to pipe `yes` as their input.

### Syntax
```
$ wrangler secret:delete secret-name
=> Are you sure?
# or
$ yes | wrangler secret:create secret-name
```
this command should support the environment flag; when not present, the command should remove the secret in the context of the top level worker. when present, the command should remove the secret in the context of the worker specified in the env.

### Depends on
* workers config api upload change; if the upload script endpoint explicitly deletes secret bindings if they are not present in the metadata sent with the upload form, then this endpoint and command are unnecessary, and the user can simply remove the secret binding name from their wrangler.toml.
* add endpoint to workers configuration api to delete a secret binding from a worker
* add endpoint definition to cloudflare-rs for calling the above endpoint

Issue #909, opened 2019-11-26
command to list secrets for a worker


ashleymichal:
wrangler should support a command that allows the user to view a list of secret bindings configured on their worker. 

### Syntax
```
$ wrangler secret:list
```

this command should support the environment flag; when not present, the command should list secrets bound to the top level worker. when present, the command should list secrets bound to the worker specified in the env.

### Depends on
add endpoint definition to cloudflare-rs for calling the GET script bindings endpoint

victoriabernard92:
Maybe add a note on whether env scoping will be supported (I think it's overkill for list)

ashleymichal:
> Maybe add a note on whether env scoping will be supported (I think it's overkill for list)

actually env scoping will be necessary, since every env is its own script

PR #1045, opened 2020-02-11
Add Secret commands

### LINKED ISSUE #907: command to add a secret to a worker
wrangler should support a command that allows the user to securely add a secret environment variable to the script of their choosing. should be interactive by default to prevent sensitive data from leaking into terminal history. ideally also allows for values to be piped to it.

### Syntax
```
$ wrangler secret set secret-name
=> Enter your secret value
# or
$ echo $SECRET_VALUE | wrangler secret secret-name
```
this command should support the environment flag; when not present, the command should create a secret in the context of the top level worker. when present, the command should create a secret in the context of the worker specified in the env.

#### outstanding questions:

1. what to do when the worker does not exist yet
2. does this require extra encryption before sending?

### Depends on
* add endpoint to workers configuration api to add a secret binding to an already existing worker
* add endpoint definition to cloudflare-rs for calling the above endpoint

ashleymichal:
closes #907, #912, #909 . just a repackaging of #994


src/terminal/interactive.rs
```diff
@@ -0,0 +1,26 @@
+// For interactively handling  reading in a string
+pub fn get_string(prompt_string: &str) -> String {
+    println!("{}", prompt_string);
+    let input: String = read!("{}\n");
```
EverlastingBugstopper: nit: i wonder if you can `read!("{} {}", prompt_string);

src/terminal/interactive.rs
```diff
@@ -0,0 +1,26 @@
+// For interactively handling  reading in a string
+pub fn get_string(prompt_string: &str) -> String {
```
EverlastingBugstopper: nit: rename to `get_user_input`

src/terminal/interactive.rs
```diff
@@ -0,0 +1,26 @@
...
+const YES: &str = "y";
+const NO: &str = "n";
+// For interactively handling deletes (and discouraging accidental deletes).
+// Input like "yes", "Yes", "no", "No" will be accepted, thanks to the whitespace-stripping
+// and lowercasing logic below.
+pub fn delete(prompt_string: &str) -> Result<bool, failure::Error> {
+    println!("{} [y/n]", prompt_string);
+    let mut response: String = read!("{}\n");
+    response = response.split_whitespace().collect(); // remove whitespace
+    response.make_ascii_lowercase(); // ensure response is all lowercase
```
EverlastingBugstopper: is there a reason we use `make_ascii_lowercase` instead of `to_lowercase`?

src/commands/secret/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// secret_errors() provides more detailed explanations of Workers KV API error codes.
+// See https://api.cloudflare.com/#workers-secrets ? for details.
+fn secret_errors(error_code: u16) -> &'static str {
+    match error_code {
+        7003 | 7000 => {
+            "Your wrangler.toml is likely missing the field \"account_id\", which is required to write to Workers KV."
+        }
+        10001 => "The content you passed in is not an accepted string. Try entering just a string",
```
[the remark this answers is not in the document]
ashleymichal: i think this is not a thing that gets returned for these endpoints anyway.. will remove.

src/commands/secret/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+        10053 => "There is already another binding with a different type by this name. Check your wrangler.toml or dashboard for conflicting bindings",
```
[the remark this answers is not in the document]
ashleymichal: so this happens in the  case that there is a binding by the same name but of another kind, so perhaps a kv namespace or plain text binding, for example. this isn't in the case of a secret by the same name, because this is a set so it'll just overwrite.

src/terminal/interactive.rs
ashleymichal: it got mad, or i don't understand.

src/terminal/interactive.rs
ashleymichal: idk, @victoriabernard92 ?

victoriabernard92: This was copy from gabbi's work on interactive delete

ashleymichal: interesting. it is oddly specific...perhaps we should leave it?

ashleymichal: i'm going to revert it to using this, and cc @gabbifish for when she gets back to see if we can change it. if not, i'd like to add a comment so we don't re-litigate this in the future.

src/commands/secret/mod.rs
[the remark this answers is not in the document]
EverlastingBugstopper: ah. how about `"There is already a KV namespace or text binding with this name associated with your script"`or something like that/ "binding with a different type by this name" is hard to parse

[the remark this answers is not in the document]
ashleymichal: there is a world in which it is actually a text blob binding you are using for sites, or you went and called it `wasm` or something else that's taken by our build process. if we wanted to be really robust, we could query the list of bindings to show what's up, check their toml, look to see if they're stomping  on their own asset manifest or wasm binding... this feels like a good add-on issue though.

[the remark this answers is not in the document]
EverlastingBugstopper: 👍

PR #1058, opened 2020-02-13
Change file permissions

xtuc:
`npm/binary.js` doesn't need to be executable. This change removes the
flag.


EverlastingBugstopper:
hmm - what exactly did this do? did you run chmod on it? i see no changes in diff :octocat:

xtuc:
I ran `chmod -x npm/binary.js`

Issue #681, opened 2019-09-23
Dist script should be named worker.js


sheerun:
#### Environment

I'm using newest wrangler 1.4.0-rc.1

#### Steps to reproduce

The dist file and sourcemap are named `main.js` and `main.js.map`, but when using error reporting like sentry it turns out that server-side .js file is names worker.js. Because of this there's filename mismatch and sentry cannot link uploaded sourcemaps to error stacktrace and we get "Source code was not found" error.

#### What did you expect to see?

If generated file and sourcemap were names worker.js Sentry would would out of the box.

Here's how proper configuration should look like:

```
  devtool: 'cheap-module-source-map',
  output: {
    path: path.join(__dirname, 'dist'),
    publicPath: 'dist',
    filename: 'worker.js',
    sourceMapFilename: 'worker.js.map'
  },

```

xtuc:
that's interesting.

Currently, Cloudflare worker hardcodes the name of the script as `worker.js`, that's why your source map won't match.
We could easily rename `worker/script.js` to `worker.js`. Alternatively, we could use the name in the source map instead of hardcoding it to `worker.js`

sheerun:
In any case Sentry won't work until worker.js is used in sourcemap. I hope you are able to confirm this issue and fix it (you don't need to use my solution).

xtuc:
Implemented your fix in https://github.com/cloudflare/wrangler/pull/1063.

PR #1063, opened 2020-02-14
Configure sourcemap file

xtuc:
`webpack` will by default emit a sourcemap that maps to a `main.js`
file. However, the worker runtime will create the script with a
`worker.js` name. In some cases (like devtools) it doesn't matter
because it resolves the source infos based on the sourcemap `sources`
key. In Sentry for instance, that would fail.

Fixes https://github.com/cloudflare/wrangler/issues/681

### DIFF
--- a/wranglerjs/index.js
+++ b/wranglerjs/index.js
@@ -3,10 +3,16 @@ const { join } = require("path");
 const fs = require("fs");
 const WasmMainTemplatePlugin = require("webpack/lib/wasm/WasmMainTemplatePlugin");
 
+const WEBPACK_OUTPUT_FILENAME = "worker.js";
+const WEBPACK_OUTPUT_SOURCEMAPFILENAME = WEBPACK_OUTPUT_FILENAME + ".map";
+
 function error(msg) {
   console.error("Error: " + msg);
   process.exit(1);
   return new Error("error");
+}
+function warn(...msg) {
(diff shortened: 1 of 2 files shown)


wranglerjs/index.js
```diff
@@ -54,6 +54,18 @@ function filterByExtension(ext) {
...
 
+  // The worker runtime will set the name of the script to `worker.js`,
+  // regardless of what's specified in the sourcemap.
+  // We can tell webpack to name the generated worker by configuring the output.
+  // It's also safe to force that configuration because it mirrors what the
+  // runtime does.
+  // https://github.com/cloudflare/wrangler/issues/681
+  if (config.output === undefined) {
+    config.output = {};
+  }
```
EverlastingBugstopper: what do you think about adding some logic here to provide this info to users? i'm thinking a println with an emoji::INFO should do the trick: maybe something like "webpack's output filename is being renamed to `worker.js` because of requirements from the Workers runtime"

xtuc: I'm not so an emoji person but I will give it a try

xtuc:
Doing a few WIP commits on this branch, please ignore

wranglerjs/index.js
ashleygwilliams: @EverlastingBugstopper can you file an issue for this, i'm gonna merge this :)

PR #1070, opened 2020-02-14
Sends custom user agent with cloudflare-rs

### LINKED ISSUE #731: Use wrangler user agent for cloudflare-rs requests
Might need to make a request to cloudflare-rs to do this, or to modify the `reqwest` that we get back from cloudflare-rs before executing it.

EverlastingBugstopper:
Fixes #731


src/commands/publish/route.rs
```diff
@@ -29,7 +28,7 @@ pub fn publish_routes(
 }
 
 fn fetch_all(user: &GlobalUser, zone_identifier: &str) -> Result<Vec<Route>, failure::Error> {
-    let client = cf_v4_api_client(user, HttpApiClientConfig::default())?;
```
ashleygwilliams: `HttpApiClientConfig::default()` <- what does this eval to?

EverlastingBugstopper: https://github.com/cloudflare/cloudflare-rs/blob/master/cloudflare/src/framework/mod.rs#L69

src/commands/kv/mod.rs
```diff
@@ -19,15 +20,7 @@ pub mod namespace;
...
-    http::cf_v4_api_client(
-        user,
-        HttpApiClientConfig {
-            default_headers: http::headers(None),
-            // Use 5 minute timeout instead of default 30-second one.
-            // This is useful for bulk upload operations.
-            http_timeout: Duration::from_secs(5 * 60),
-        },
-    )
+    http::cf_v4_api_client(user, Some(Feature::KV))
```
ashleymichal: these are trying to do vastly different things and i don't think this is the right move.

EverlastingBugstopper: 👍 i've updated the `cf_api_client` to take a `CfApiClientConfig` struct which has a builder API and a default

EverlastingBugstopper: this allows you to either pass `CfApiClientConfig::default()` as the second parameter or you could do `CfApiClientConfig::builder().timeout(Duration::from_secs(5 * 60)).feature(Feature::Sites).build()`

I've left the legacy api without something similar because we dont need to overengineer something we're moving away from

src/http/v4.rs
```diff
@@ -0,0 +1,94 @@
...
+use crate::settings::global_user::GlobalUser;
+use crate::terminal::{emoji, message};
+
+pub struct CfApiClientConfig {
+    feature: Option<Feature>,
+    http_timeout: Duration,
+}
+
+impl CfApiClientConfig {
+    pub fn default() -> HttpApiClientConfig {
```
ashleymichal: this isn't the usual signature of default; Default is a Trait where the interface is a function default that returns the struct it is impl'd on; so here it could be kind of  confusing. i'm inclined to simplify into bare functions that return an HttpApiClientConfig which includes our default headers. Since HttpApiClientConfig already has a field for timeout, it feels weird to do this if what we're trying to standardize is the headers. i'd be happy writing

```rust
let http_config = HttpApiConfig {
  default_headers: wrangler_headers(Some(feature))
  ..Default::default()
}}
```
or mutating a default struct:

```rust
let mut http_config = wrangler_default_api_config();
add_feature(&http_config, FEATURE);
```
until we need something more clever. this pattern feels like overkill here atm.

EverlastingBugstopper: ooh - TIL that Rust has a spread operator like JavaScript. Definitely gonna use that instead

src/http/cf.rs
```diff
@@ -0,0 +1,89 @@
...
+    };
+
+    HttpApiClient::new(
+        Credentials::from(user.to_owned()),
+        config,
+        Environment::Production,
+    )
+}
+
+pub fn kv_auth_client(user: &GlobalUser) -> Result<HttpApiClient, failure::Error> {
```
EverlastingBugstopper, line 25: maybe we rename to `long_timeout_auth_client`?

src/commands/config/mod.rs
```diff
@@ -46,7 +45,7 @@ pub fn global_config(user: &GlobalUser, verify: bool) -> Result<(), failure::Err
 // validate_credentials() checks the /user/tokens/verify endpoint (for API token)
 // or /user endpoint (for global API key) to ensure provided credentials actually work.
 pub fn validate_credentials(user: &GlobalUser) -> Result<(), failure::Error> {
-    let client = http::cf_v4_api_client(user, HttpApiClientConfig::default())?;
+    let client = http::auth_client(user)?;
```
ashleymichal, line 48: the distinction for this client really is that it is optimized as a client for the cloudflare API, not that it includes auth info, so i'd be inclined to leave this name as it was.

src/http/cf.rs
ashleymichal: yeah prolly. or we handle config less centrally in another iteration of this refactor

PR #1049, opened 2020-02-12
Adds alpha warning to wrangler dev

EverlastingBugstopper:
it comes with a fancy auto-resizing box

```console
$ wrangler dev
      ┌────────────────────────────────────────────────────────────────┐
      │                                                                │
      │             wrangler dev is currently unstable and             │
      │            there are likely to be breaking changes!            │
      │                                                                │
      │                                                                │
      │            For this reason, we cannot yet recommend            │
      │           using wrangler dev for integration testing           │
      │                                                                │
      │                Please submit any feedback here:                │
      │       https://github.com/cloudflare/wrangler/issues/1047       │
      │                                                                │
      │                                                                │
      └────────────────────────────────────────────────────────────────┘
```

```console
$ wrangler dev
      ┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
      │                                                                                                                                   │
      │                          wrangler dev is currently unstable and there are likely to be breaking changes!                          │
      │                                                                                                                                   │
      │                        For this reason, we cannot yet recommend using wrangler dev for integration testing                        │
      │                        Please submit any feedback here: https://github.com/cloudflare/wrangler/issues/1047                        │
      │                                                                                                                                   │
      └───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

and on terminals that are too small, it falls back to regular prints

```console
$ wrangler dev
⚠️  wrangler dev is currently unstable and there are likely to be breaking changes!
⚠️  For this reason, we cannot yet recommend using wrangler dev for integration testing

💁  Please submit any feedback here: https://github.com/cloudflare/wrangler/issues/1047
```

PR #1071, opened 2020-02-14
fixes clippy warnings

EverlastingBugstopper:
also includes changes in previously approved pr #1072


src/terminal/message_box.rs
```diff
@@ -20,7 +20,7 @@ pub fn dev_alpha_warning() {
             print_centered(width, FEEDBACK_MSG);
             print_centered(width, "");
             print_bottom(width);
-            println!("");
+            println!();
```
xtuc: what's that supposed to do?

EverlastingBugstopper: this code was introduced in #1049 and this diff is just from clippy telling me not to send it an empty &str

PR #1073, opened 2020-02-14
1.8.0-rc.1

EverlastingBugstopper:
semi-changelog

things that changed since last release:

wrangler dev now has colors enabled
the `text` field is now the `vars` field
environment vars no longer inherit from the top-level config
if webpack outputs a source map it will automatically be renamed to match the names the workers runtime gives them

if we merge
`wrangler whoami` now lists the account ids and zone ids with the current authenticated user

**still need to run tests locally**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -488,9 +488,9 @@ dependencies = [
 
 [[package]]
 name = "csv-core"
-version = "0.1.9"
+version = "0.1.10"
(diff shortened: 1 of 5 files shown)


EverlastingBugstopper:
@ashleygwilliams this should be ready to go - once the release is done building we should be able to publish on npm. will want to update the release notes with a bit more polished version of what i have in the description here but i'm heading home for the day 😄

PR #1075, opened 2020-02-14
refactor: extract upload and deploy to lib modules

ashleymichal:
this is a pure refactor i've been meaning to do for awhile; it pulls a lot of the logic out of the publish command into two lib level modules: upload and deploy. the logic in these is used in a couple of places, and it's starting to make sense to keep it out of the commands modules. this will help with a few issues, including uploading a "draft" when a secret is created for a new project, and a lot of the work around consolidating oddly duplicated logic between preview, dev, and publish.


src/upload/form/mod.rs
```diff
@@ -19,6 +19,7 @@ use project_assets::ProjectAssets;
 use text_blob::TextBlob;
 use wasm_module::WasmModule;
 
+// these likely belong elsewhere, perhaps in a mod for project type specific behavior.
```
EverlastingBugstopper: nit: i think we should remove this comment and either file an issue or fix it up in this refactor, thoughts?

ashleymichal: i'm happy to do it in this refactor, i'll put it on my Friday list

ashleymichal: jk we need this in, can make an issue.

ashleymichal: https://github.com/cloudflare/wrangler/issues/1083

PR #1087, opened 2020-02-19
Upload "draft" worker if secret is created before initial worker script has been uploaded

### LINKED ISSUE #913: allow wrangler to upload a "draft" of a worker before uploading a secret, if the worker script doesn't already exist yet.
because secrets must be attached to a specific worker, it would be nice to be able to upload a "draft" of a worker, that is, push to the upload script endpoint without setting any routes or making the script available on a workers.dev subdomain. There is an open question about whether this should be a stand alone command, a flag on publish, or just an operation wrangler does automatically on the user's behalf if they attempt to create a secret for a script that has never been uploaded before.

gabbifish:
This PR closes #913.

It introduces fallback logic to upload a "draft" worker whenever a secrets API call is made for a specific script name, and a worker for the script name does not exist yet. This fallback logic currently applies to secret upload and deletion; we can also extend it to `wrangler secret list` if we'd like. I currently avoid doing that given that one is extremely unlikely to use `wrangler secret list` before calling `wrangler secret put/delete`, but then again, it could be worth preventing unexpected behavior for this CLI call as well.


ashleymichal:
> This fallback logic currently applies to secret upload **and deletion**;

if the script doesn't exist, then no secret could exist to delete; i think this logic only makes sense for `put`, and the error should be handled with a message::warn on `list` and `delete`.

src/commands/secret/mod.rs
```diff
@@ -35,25 +36,50 @@ fn validate_target(target: &Target) -> Result<(), failure::Error> {
...
         10055 => "You have exceeded the limit of 32 text bindings for this worker. Run `wrangler secret list` or go to your Cloudflare dashboard to clean up unused text/secret variables",
         _ => "",
     }
 }
 
+// upload_draft_worker will attempt to upload a "draft" version of a worker script if it does not
+// already exist in the API (API error code 10007 is returned). The function returns None if this draft
+// script was uploaded, or else returns Some (with a Result type so we can return a potential script upload error
+// up the call chain).
+pub fn upload_draft_worker(
```
ashleymichal: this really only makes sense to do on `put`; i'd say we should just inline the logic there.

src/commands/secret/mod.rs
```diff
@@ -108,12 +151,29 @@ pub fn delete_secret(name: &str, user: &GlobalUser, target: &Target) -> Result<(
...
         script_name: &target.name,
-        secret_name: name,
+        secret_name: name.clone(),
     });
 
     match response {
-        Ok(_) => message::success(&format!("You've deleted the secret {}.", name)),
-        Err(e) => failure::bail!(format_error(e)),
+        Ok(_) => message::success(&format!("Success! Deleted secret {}.", name)),
+        Err(e) => match upload_draft_worker(&e, user, target) {
```
ashleymichal: it doesn't make sense to create a  worker  in this scenario; we should use normal error messages here.

src/commands/secret/mod.rs
gabbifish: The code itself for upload_draft_worker is verbose enough that I do like the idea of still keeping a separate function around for it--it makes for nice decomposition imho!

src/commands/secret/mod.rs
```diff
@@ -35,25 +36,50 @@ fn validate_target(target: &Target) -> Result<(), failure::Error> {
...
+pub fn upload_draft_worker(
+    e: &ApiFailure,
+    user: &GlobalUser,
+    target: &Target,
+) -> Option<Result<(), failure::Error>> {
+    match e {
+        ApiFailure::Error(_, api_errors) => {
+            for error in &api_errors.errors {
+                if error.code == 10007 {
+                    message::working("This worker doesn't exist in the API yet. Uploading worker script so we can create new secret.");
```
ashleymichal: nit: (warning: not `cargo check`-ed)
```suggestion
                    message::working(&fmt("Worker {} doesn't exist yet. Creating a draft to attach secret to.", target.name));
```

src/commands/secret/mod.rs
ashleymichal: fair.

src/commands/secret/mod.rs
```diff
@@ -35,25 +36,50 @@ fn validate_target(target: &Target) -> Result<(), failure::Error> {
 fn secret_errors(error_code: u16) -> &'static str {
     match error_code {
         7003 | 7000 => {
-            "Your wrangler.toml is likely missing the field \"account_id\", which is required to write to Workers KV."
+            "Your wrangler.toml is likely missing the field \"account_id\", which is required to create a secret."
         }
```
ashleymichal: can we add a message for delete/list 10007 errors here?

gabbifish: We'll handle this in a separate ticket given that not found errors for both workers and bindings are both error code 10007.

PR #1097, opened 2020-02-20
1.8.0-rc.2

### LINKED ISSUE #1088: wrangler list secrets produces broken output when list of secrets is empty
If I attempt to run `wrangler secret list` for a worker which has no secrets, I get unexpected output:

```
$ wrangler secret list
Error:  Error: error decoding response body: invalid type: null, expected a sequence at line 2 column 16
```

This should ideally return no output (or maybe helpful text like `No secrets found for worker <worker-name>`).

Tagging @ashleymichal @victoriabernard92 :)

gabbifish:
Mini changelog:

Implemented #913 
Fixed #1088 in API
Minor readme fixes for `wrangler dev`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -512,9 +512,9 @@ dependencies = [
 
 [[package]]
 name = "curl-sys"
-version = "0.4.25"
+version = "0.4.26"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0c38ca47d60b86d0cc9d42caa90a0885669c2abc9791f871c81f58cdf39e979b"
+checksum = "0853fe2a575bb381b1f173610372c7722d9fa9bc4056512ed99fe6a644c388c6"
 dependencies = [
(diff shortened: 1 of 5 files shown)


EverlastingBugstopper:
just a heads up since i will be ooo when this is tagged - if the github release fails (i've had the action flake out on me before with an opaque error message) you will need to delete the tag, push the deleted tag, and then re-add and re-push the tag again :)

gabbifish:
@ashleygwilliams good catch, the readme changes don't seem to have been copied to the npm dir as well. @EverlastingBugstopper, since you handled this community PR, do you mind taking this on? Otherwise I can pick up the pieces!

EverlastingBugstopper:
for future reference: this is something we always do in our release process - not when changes are made to the README.

see item #11 [here](https://github.com/cloudflare/wrangler/blob/master/RELEASE_CHECKLIST.md)

Issue #975, opened 2019-12-18
all wrangler commands should accept `--verbose` flag


ashleymichal:
Even if it doesn't actually change the output, it is jarring to pass this flag and have wrangler complain. we should accept it everywhere and use it where applicable.

ericdeansanchez:
Hi 👋, could I start on this?

ashleymichal:
@ericdeansanchez i see no reason why not! it should be a pretty straight forward change in `main.rs`, and a good opportunity to learn clap.rs. feel free to ping here if you want any assistance or suggestions!

ashleymichal:
@ericdeansanchez there are a couple of commands (i don't remember which off the top of my head) that already accept a verbose flag; we should follow the same pattern those use but just kind of... ignore it where it is not already used.

ericdeansanchez:
Ok, great! And yeah I thought that we could add the `--verbose` flag as an `Arg` to all the `SubCommand`s. 

And just to clarify, you want `--verbose` to be ignored **silently** when passed and only applied where verbose output is supported?

PR #1110, opened 2020-02-27
add verbose argument acceptance to commands

bradyjoslin:
addresses #975


victoriabernard92:
I agree with @EverlastingBugstopper it's confusing to have a verbose flag added to a subcommand if it doesn't do anything yet. But it shouldn't yell at you if you do add it, just shouldn't show up on the help if it does nothing.

Is it possible for the subcommands that don't do anything to just accept the verbose arg, but not show up with `--help`?

EverlastingBugstopper:
hmm.. i don't think that is possible, but on second thought the goal here is to prevent the commands from failing and this achieves that. 

going to cc in @ashleymichal since she filed that issue and leave her to approve if she thinks it's good 🚀

ashleymichal:
i think it is more disruptive to complain when passed the verbose flag than to see it in the help and not see any difference in output. that being said, upon searching the clap docs i did find a couple of argsettings that might be helpful: Global and Hidden: https://docs.rs/clap/2.33.0/clap/enum.ArgSettings.html#variant.Global
https://docs.rs/clap/2.33.0/clap/enum.ArgSettings.html#variant.Hidden

bradyjoslin:
Great feedback @ashleymichal

A straightforward way of conditionally displaying the help text while always handling the verbose flag is to define two variables for the verbose argument, one with the hidden argument setting set to true.  Example implementation below, where verbose_arg would be added to any subcommands that handles the verbose flag, and the silent_verbose_arg applied where it isn't.

```Rust
    let verbose_arg = Arg::with_name("verbose")
        .long("verbose")
        .takes_value(false)
        .help("toggle verbose output");

    let silent_verbose_arg = verbose_arg.clone().hidden(true);
...
        .subcommand(
            SubCommand::with_name("init")
                .about(&*format!(
                    "{} Create a wrangler.toml for an existing project",
                    emoji::INBOX
                ))
                .arg(silent_verbose_arg.clone())
...
        .subcommand(
            SubCommand::with_name("preview")
                .about(&*format!(
                    "{} Preview your code temporarily on cloudflareworkers.com",
                    emoji::MICROSCOPE
                ))
                .arg(verbose_arg.clone())
...
```

As an alternative solution, I evaluated setting a global arg for verbose with the hidden flag enabled, but struggled finding a way to override the hidden argument setting for those subcommands that handle the flag without getting a runtime error for non-unique argument names.

If the two variable option is sufficient, happy to make the change.

EverlastingBugstopper:
Hey @bradyjoslin the two variable approach seems perfect

ashleymichal:
hey @bradyjoslin , would you  like to take a stab at the two-variable approach? or should  we  pick this PR up?

bradyjoslin:
@ashleymichal - Hey, thanks for the reminder on this.  Should be something I can tackle, I'll review where I left off in the next few days.

Hope you all are doing well - the recent additions to Wrangler have been awesome!

bradyjoslin:
Two-variable verbose handling changes done.  Let me know if you see anything that needs refinement.

EverlastingBugstopper:
Thanks @bradyjoslin!

PR #1114, opened 2020-03-02
[dev] Improves error message when there is no internet connection

### LINKED ISSUE #1093: 1.8.0-rc.1 needs internet to build locally (-dev)
Would be good to document that `wrangler dev` needs internet to build  :)

EverlastingBugstopper:
Fixes #1093 

**Before:**

```
Error: error sending request for url (https://api.cloudflare.com/client/v4/accounts/12s4t3i5n8k8y0f2ar7t5s1/workers/scripts/worker/preview): error trying to connect: dns error: failed to lookup address information: nodename nor servname provided, or not known
```

**After:**

```
Error: Could not upload your script, check your internet connection or https://www.cloudflarestatus.com/ for any incidents impacting the Cloudflare Workers API.
```

### DIFF
--- a/src/commands/dev/mod.rs
+++ b/src/commands/dev/mod.rs
@@ -196,7 +196,7 @@ pub fn get_preview_id(
     verbose: bool,
 ) -> Result<String, failure::Error> {
     let sites_preview = false;
-    let script_id = upload(&mut target, user.as_ref(), sites_preview, verbose)?;
+    let script_id = upload(&mut target, user.as_ref(), sites_preview, verbose).map_err(|_| failure::format_err!("Could not upload your script, check your internet connection or https://www.cloudflarestatus.com/ for any incidents impacting the Cloudflare Workers API."))?;
     Ok(format!(
         "{}{}{}{}",
         &script_id,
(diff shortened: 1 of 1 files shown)


src/commands/dev/mod.rs
```diff
@@ -196,7 +196,7 @@ pub fn get_preview_id(
     verbose: bool,
 ) -> Result<String, failure::Error> {
     let sites_preview = false;
-    let script_id = upload(&mut target, user.as_ref(), sites_preview, verbose)?;
+    let script_id = upload(&mut target, user.as_ref(), sites_preview, verbose).map_err(|_| failure::format_err!("Could not upload your script, check your internet connection or https://www.cloudflarestatus.com/ for any incidents impacting the Cloudflare Workers API."))?;
```
victoriabernard92:
```suggestion
    let script_id = upload(&mut target, user.as_ref(), sites_preview, verbose).map_err(|_| failure::format_err!("Could not upload your script. Check your internet connection or https://www.cloudflarestatus.com/ for rare incidents impacting the Cloudflare Workers API."))?;
```

victoriabernard92: When I user gets this their attention will immediately be directed to check cloudflarestatus.com which likely is not the problem it's much more likely their internet

ashleymichal:
nit: in theory you should be able to see the difference between a network call that never hit the server and one that came back with a gateway error.

PR #1115, opened 2020-03-02
Delay KV bulk delete

### LINKED ISSUE #783: Bug: Downtime on Site update
When we publish a `site` worker, often the worker and its accompanying asset manifest make it to the edge ahead of the new KV keys, and/or the "old" KV keys are removed before the worker is updated. This causes a window of downtime when a page is updated. We need to somehow delay removing "old" assets from KV until everything is live. There are a few ways to implement this, will update this issue with possible strategies soon.

ashleymichal:
## Overview
Fixes #783 

Previously, we would run a full KV namespace sync BEFORE uploading the worker script to the edge. This caused some unacceptable downtime for updated assets, as the Worker would continue  handling requests to paths with outdated KV keys. This PR does a few things in service of re-ordering these operations, but the bottom line is that now on Wrangler publish, when the project is a Workers Site, we:

1. validate the wrangler.toml
2. run `wrangler build`
3. query the site's KV namespace for existing keys and diff them against the local site `bucket` to build the list of files to upload, the list of keys to delete, and the asset manifest to be uploaded with the Worker script.
4. execute the bulk file upload
5. upload the worker with its new asset manifest
6. execute the bulk file delete

This has the added benefit of NOT deleting old values in the case where the Worker upload fails, since the first KV operation is additive.

## Things this is not
This PR introduces a little more verbosity into preview/upload and publish, however there is considerable overlap between those operations already, but I'd like to address those in a separate PR in order to expedite this fix.

I'm also interested in optimizing the preview command's behavior (refrain from deleting until a `ctrl-C` or something like that) but not here.

## 💅 
There's a few house-keeping changes here, though more remain which I'll include as inline comments:

* we now pass a set of KeyValuePairs to kv::bucket::upload, rather than a bucket.
* some house-keeping in the publish command, mostly renaming and some refactors
* ~remove unused/unsupported `bucket` attribute from KvNamespace struct~

### DIFF
--- a/src/commands/kv/bucket/mod.rs
+++ b/src/commands/kv/bucket/mod.rs
@@ -6,6 +6,7 @@ mod upload;
 
 pub use manifest::AssetManifest;
 pub use sync::sync;
+pub use upload::upload_files;
 
 use std::ffi::OsString;
 use std::fs;
--- a/src/commands/kv/bucket/mod.rs
+++ b/src/commands/kv/bucket/mod.rs
@@ -76,7 +76,7 @@ pub fn directory_keys_values(
         Ok(_file_type) => {
             // any other file types (files, symlinks)
             // TODO: return an error type here, like NotADirectoryError
-            Err(format_err!("site.bucket must be a directory"))
+            Err(format_err!("Check your wrangler.toml; the `bucket` attribute for [site] should point to a directory."))
         }
         Err(e) => Err(format_err!("{}", e)),
     }
--- a/src/commands/kv/bucket/mod.rs
+++ b/src/commands/kv/bucket/mod.rs
@@ -4,9 +4,6 @@ mod manifest;
 mod sync;
 mod upload;
 
-use data_encoding::HEXLOWER;
-use sha2::{Digest, Sha256};
-
 pub use manifest::AssetManifest;
 pub use sync::sync;
 
@@ -14,10 +11,13 @@ use std::ffi::OsString;
 use std::fs;
 use std::path::Path;
 
-use cloudflare::endpoints::workerskv::write_bulk::KeyValuePair;
-
+use data_encoding::HEXLOWER;
+use failure::format_err;
 use ignore::overrides::{Override, OverrideBuilder};
 use ignore::{Walk, WalkBuilder};
+use sha2::{Digest, Sha256};
+
+use cloudflare::endpoints::workerskv::write_bulk::KeyValuePair;
 
 use crate::settings::toml::Target;
 use crate::terminal::message;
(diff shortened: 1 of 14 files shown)


src/commands/kv/bucket/mod.rs
```diff
@@ -32,43 +33,53 @@ pub fn directory_keys_values(
...
-            asset_manifest.insert(url_safe_path, key);
+                    asset_manifest.insert(url_safe_path, key);
+                }
+            }
+            Ok((upload_vec, asset_manifest))
+        }
+        Ok(_file_type) => {
+            // any other file types (files, symlinks)
+            // TODO: return an error type here, like NotADirectoryError
+            Err(format_err!("site.bucket must be a directory"))
```
EverlastingBugstopper: Can we get a more actionable error message here that includes the value of what they have set as their bucket and direct them to their `wrangler.toml`? "site.bucket" is a bit opaque.

src/commands/publish.rs
```diff
@@ -57,73 +80,57 @@ fn warn_site_incompatible_route(deploy_config: &DeployConfig) {
...
+            "{} bucket directory \"{}\" does not exist",
+            emoji::WARN,
+            path.display()
+        )
+    } else if !path.is_dir() {
+        failure::bail!(
+            "{} bucket \"{}\" is not a directory",
+            emoji::WARN,
+            path.display()
+        )
```
EverlastingBugstopper: these failures are not related to if it is a root bucket, should probably rename the function 

(worth noting that it's also invalid to have bucket be `dist` for webpack types but we can leave that for another PR)

src/commands/publish.rs
```diff
@@ -57,73 +80,57 @@ fn warn_site_incompatible_route(deploy_config: &DeployConfig) {
...
-            }
-        }
+// We don't want folks setting their bucket to the top level directory,
+// which is where wrangler commands are always called from.
+pub fn assert_non_root_bucket(bucket: &PathBuf) -> Result<(), failure::Error> {
+    // TODO: this should really use a convenience function for "Wrangler Project Root"
+    let current_dir = env::current_dir()?;
+    if bucket.as_os_str() == current_dir {
+        failure::bail!(
+            "{} You need to specify a bucket directory in your wrangler.toml",
```
EverlastingBugstopper: this seems like the wrong error message

ashleymichal:
update: i'm not  necessarily gunning to get this into 1.8.1; i'm happy to wait on 1.9.0. one of the refactors here removes the `bucket` attribute from kv-namespaces, which from what i could tell was never supported, but i want to do a little extra testing to confirm that there isn't any support here so i don't break unsuspecting crafty users.

AngusMorton:
> update: i'm not necessarily gunning to get this into 1.8.1; i'm happy to wait on 1.9.0. one of the refactors here removes the `bucket` attribute from kv-namespaces, which from what i could tell was never supported, but i want to do a little extra testing to confirm that there isn't any support here so i don't break unsuspecting crafty users.

Is there an ETA for 1.9.0 or for this to be released (even as an alpha/RC)?

ashleymichal:
> Is there an ETA for 1.9.0 or for this to be released (even as an alpha/RC)?

We'll be putting this out on its own as a point release (1.8.2)

PR #1117, opened 2020-03-03
[dev] fix: properly route to new preview id after reload

### LINKED ISSUE #1082: [dev] Worker updates not seen in terminal until restarting dev
_Mentioned initially in https://github.com/cloudflare/wrangler/issues/1078#issuecomment-587695836._

@EverlastingBugstopper 

#### Environment

* operating system: MacOS (Catalina, 10.15.1)
* output of `rustc -V`: n/a
* output of `node -v`: v10.17.0
* output of `wrangler -V`: 1.8.0-rc.1

#### Steps to reproduce

With dev running, add a console log statement inside the worker. Save. The console log does not show up. Restart dev, the console log is there.

The new console log should show up in the terminal without dev needing to be restarted.

The new console log does not show up in the terminal.

### LINKED ISSUE #1107: [dev] On change, cloudflared tunnel breaks
@EverlastingBugstopper

> If you would like to try something that hasn't been tested yet, you could try to put a worker in front of something you're running on localhost.
>
> Let's say you have a react app that you're developing locally and it's running on localhost:4000 and you want to also develop a worker at the same time (maybe play around with the new HTMLRewriter API a bit). Unfortunately, you cannot call $ wrangler dev --host localhost:4000 because Cloudflare's preview service needs the back end to be a public URL.
>
> What you can do instead is spin up a TryCloudflare endpoint and pass that url to the --host argument of wrangler dev and then develop them in tandem.
>
> This is a completely unpaved path that has not been tested at all so your mileage may vary. If you get it to work, we'd love to hear about it!

I've tested this out and it _kinda_ works. The first issue I've run into is getting changes to update. The steps of what I'm seeing: 
1. save a change (say, a simple change in my worker site's HTML)
2. refresh the page 
3. the change is not there...
4. if I restart `wrangler dev` the change is there
5. if i _do not_ restart `wrangler dev` but instead refresh my browser every few seconds I eventually see a 500 for the HTML page with an error code along the lines of `could not find index.cc2f4dbd9f20219bea74e17400a95478e40941f0f0a4c17e2ec83eac27a2e0e8.html in your content namespace`.
6. Wait maybe 2 minutes or so (seriously!), refreshing every few seconds, and the 500 is gone and the update is live (without having restarted `wrangler dev`).

So it seems like there's some type of conflict between the way `wrangler dev` is updating on changes and the way the tunnel is serving them.

EverlastingBugstopper:
Fixes #1082 and might fix #1107

We keep `preview_id` in an `Arc<Mutex` so that when a reload is detected on another thread, the dev server can properly route new requests to the new preview_id. Previously, the server would only attempt to acquire the lock every couple of minutes. After moving the lock acquisition to `service_fn` from `make_service_fn`, the lock is acquired on every request, making sure that each request is going to the right version of the script.

### DIFF
--- a/src/commands/dev/mod.rs
+++ b/src/commands/dev/mod.rs
@@ -97,13 +97,13 @@ async fn serve(
     // create a closure that hyper will use later to handle HTTP requests
     let make_service = make_service_fn(move |_| {
         let client = client.to_owned();
-        let preview_id = preview_id.lock().unwrap().to_owned();
         let server_config = server_config.to_owned();
+        let preview_id = preview_id.to_owned();
         async move {
             Ok::<_, failure::Error>(service_fn(move |req| {
                 let client = client.to_owned();
-                let preview_id = preview_id.to_owned();
                 let server_config = server_config.to_owned();
+                let preview_id = preview_id.lock().unwrap().to_owned();
                 let version = req.version();
                 let (parts, body) = req.into_parts();
                 let req_method = parts.method.to_string();


tomByrer:
> Previously, the server would only attempt to acquire the lock every couple of minutes

That would explain why small changes sometimes would not take.

EverlastingBugstopper:
Yup! We're going to try to get this fix into a point release today.

PR #1125, opened 2020-03-05
reinstate kv buckets

ashleymichal:
# Overview
in #1115 i removed this feature to clean up the logic around site deploys, however regression  testing revealed that there's a slight possibility that some users might be using this functionality in production and we don't want to introduce that kind of breaking change. This PR re-instates the feature, and smooths it out a little, but also introduces a deprecation warning, as this feature is not well thought out and was not intended to be released in the first place. we  will monitor for responses  to the warning to decide next steps.

# Also happened
to get the "feature" header in to see who uses this crazy side effect, i refactored `upload::script` to take a client as an argument. i actually think this will be beneficial for future unit testing efforts.

# Still TODO
- [x] add a deprecation warning when  the edge case is hit
- [x] add a "feature" header for instances of kv namespace uploads / publishes in service of this edge case.
- [x] use `message::billboard` for deprecation warning
- [x] create and link issue for deprecation feedback

### DIFF
--- a/src/commands/publish.rs
+++ b/src/commands/publish.rs
@@ -1,6 +1,8 @@
 use std::env;
 use std::path::{Path, PathBuf};
 
+use console::style;
+
 use crate::commands;
 use crate::commands::kv;
 use crate::commands::kv::bucket::{sync, upload_files};
@@ -27,6 +29,7 @@ pub fn publish(
         let path = &site_config.bucket.clone();
         validate_bucket_location(path)?;
         warn_site_incompatible_route(&deploy_config);
+
         let site_namespace = add_site_namespace(user, target, false)?;
 
         let (to_upload, to_delete, asset_manifest) =
@@ -38,7 +41,6 @@ pub fn publish(
         }
         upload_files(target, user, &site_namespace.id, to_upload)?;
 
-        sync_other_buckets(target, user, verbose)?;
         let upload_client = http::auth_client(Some("site"), user);
 
         // Next, upload and deploy the worker with the updated asset_manifest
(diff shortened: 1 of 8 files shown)


src/terminal/message.rs
```diff
@@ -44,3 +44,8 @@ pub fn help(msg: &str) {
     let msg = format!("{} {}", emoji::SLEUTH, msg);
     message(&msg);
 }
+
+pub fn deprecation_warning(msg: &str) {
+    let msg = format!("\n\t{} {}", emoji::WARN, msg);
+    message(&msg);
+}
```
EverlastingBugstopper: ooh we should make this use #1128 big neon box

src/commands/publish.rs
```diff
@@ -51,7 +55,17 @@ pub fn publish(
             delete_bulk(target, user, &site_namespace.id, to_delete)?;
         }
     } else {
-        upload::script(&user, &target, None)?;
+        let uses_kv_bucket = sync_non_site_buckets(target, user, verbose)?;
+
+        let feature = if uses_kv_bucket {
+            message::deprecation_warning("As of 1.9.0, you will no longer be able to specify a `bucket` for a kv namespace in your `wrangler.toml`. If your application depends on this feature, please file an issue at https://github.com/cloudflare/wrangler/issues/new with your use case.https://github.com/cloudflare/wrangler/issues/new?labels=user+report&template=feature_request.md");
```
EverlastingBugstopper: this could probably use some whitespace and maybe a bit of coloring. in #1128 i use style(url).blue().bold() for urls and style(code).yellow().bold() for code blocks (like `wrangler.toml`)

```suggestion
            message::deprecation_warning("As of 1.9.0, you will no longer be able to specify a `bucket` for a kv namespace in your `wrangler.toml`.\nIf your application depends on this feature, please file an issue with your use case here:\nhttps://github.com/cloudflare/wrangler/issues/new?labels=user+report&template=feature_request.md");
```

src/terminal/message.rs
ashleymichal: yeah i knew that was incoming; comes down to which PR gets merged first 😜

EverlastingBugstopper: maybe we can make it look different - red border and double border >:)

ashleymichal: sure; which pr gets merged first tho will determine who gets to do it 😛

src/commands/publish.rs
```diff
@@ -51,7 +57,21 @@ pub fn publish(
             delete_bulk(target, user, &site_namespace.id, to_delete)?;
         }
     } else {
-        upload::script(&user, &target, None)?;
+        let uses_kv_bucket = sync_non_site_buckets(target, user, verbose)?;
+
+        let feature = if uses_kv_bucket {
+            let wrangler_toml = style("`wrangler.toml`").yellow().bold();
+            let issue_link = style("https://github.com/cloudflare/wrangler/issues/new?labels=user+report&template=feature_request.md").blue().bold();
```
ashleymichal: actually after reading @EverlastingBugstopper 's pr, i want to make this a specific issue, probably called "deprecate kv_namespaces.bucket" or something.

src/terminal/message.rs
ashleymichal: did this.

PR #1131, opened 2020-03-10
Prototype: Wrangler Tail

gabbifish:
This PR serves as the feature branch for wrangler tailing.

So far, it does the most fundamental tasks:
- start up a simple server for receiving POSTed logs and printing them out to stdout
- start a cloudflared process (assumes you already have cloudflared installed on your host) to Workers runtime can post to it
- make the appropriate tailing API calls and regularly send heartbeats (once per 60s) to keep tailing alive.

This PR will still need some fixes before I'm satisfied:
- whenever we issue a SIGINT to stop `wrangler tail`, do: (a) graceful shutdown of cloudflared (b) make a `DELETE tail` API call to toggle logging tunnel state to "off"
- Some miscellaneous code maintenance todos
- https://github.com/cloudflare/cloudflare-rs/pull/112 needs to be merged and a new cloudflare-rs release cut.


Cargo.toml
```diff
@@ -17,7 +17,7 @@ binary-install = "0.0.3-alpha"
 chrome-devtools-rs = { version = "0.0.0-alpha.0", features = ["color"] }
 chrono = "0.4.9"
 clap = "2.32.0"
-cloudflare = "0.6.4"
+cloudflare = { path = "../cloudflare-rs/cloudflare" }
```
EverlastingBugstopper: ik this is wip but one thing i like to do when developing libraries in parallel is to comment out the `path` loading for a local dependency and replace it with a `git = https://github.com/cloudflare/cloudflare-rs` and a `rev = 7897482390` when i push it up so it's clear what code it's dependent on

src/commands/tail/mod.rs
```diff
@@ -184,8 +211,8 @@ pub fn command(args: &[&str], binary_path: &PathBuf) -> Command {
     };
 
     c.args(args);
-    // Let user read cloudflared process logs iff RUST_LOG=debug.
-    if !log_enabled!(Debug) {
+    // Let user read cloudflared process logs iff RUST_LOG=info.
```
EverlastingBugstopper: this should probably be enabled with `--verbose`

EverlastingBugstopper:
getting `Error: ⚠️  Code 10046: workers.api.error.duplicate_tail` if i ctrl+c and restart

EverlastingBugstopper:
got logs to work! piping to jq doesn't really work because of all the escapes - just looks like this: 
image

awesome work!

ashleymichal:
> piping to jq doesn't really work because of all the escapes

should be fixed now.

src/commands/tail/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+use tokio::process::Command;
+use tokio::runtime::Runtime as TokioRuntime;
+
+pub fn start_tail(target: &Target, user: &GlobalUser) -> Result<(), failure::Error> {
+    let mut runtime = TokioRuntime::new()?;
+    // Note that we use eprintln!() throughout this file; this is because we want any
+    // helpful output to not be mixed with actual log JSON output, so we use this macro
+    // to print messages to stderr instead of stdout (where log output is printed).
+    eprintln!(
+        "Setting up log streaming from Worker \"{}\" to Wrangler. This may take a few seconds...",
```
EverlastingBugstopper: nit: should add emoji or perhaps a loading spinner, can come in separate pr

src/commands/tail/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+        target.name
+    );
+
+    // TODO: get rid of this gross clone. Maybe just default to static lifetime.
```
EverlastingBugstopper: i dont think this todo is necessary, clones are pretty unavoidable w/async i think. other languages clone by default dont think it's worth beating ourselves up about them

src/commands/tail/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+    runtime.block_on(run_cloudflared_start_server(target.clone(), user.clone()))
+}
+
+// TODO: handle sigint to tear down tunnel, http server
+async fn run_cloudflared_start_server(
```
EverlastingBugstopper: nit: this name is a bit confusing, might just want to rename to `run` honestly

src/commands/tail/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+
+    match res {
+        Ok(_) => Ok(()),
+        Err(e) => failure::bail!(e),
+    }
+}
+
+async fn start_log_collection_http_server() -> Result<(), failure::Error> {
+    // Start HTTP echo server that prints whatever is posted to it.
+    let addr = ([127, 0, 0, 1], 8080).into();
```
EverlastingBugstopper: TODO: make this configurable

src/commands/tail/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+
+    let service = make_service_fn(|_| async { Ok::<_, hyper::Error>(service_fn(print_logs)) });
+
+    let server = Server::bind(&addr).serve(service);
+
+    server.await?;
+
+    Ok(())
```
EverlastingBugstopper: might be cleaner?
```suggestion
    server.await
```

src/commands/tail/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+    // let binary_path = install::install(tool_name, "cloudflare")?.binary(tool_name)?;
+
+    // TODO: allow user to pass in their own ports in case ports 8080 (used by cloudflared)
+    // and 8081 (used by cloudflared metrics) are both already being used.
+    let args = ["tunnel", "--metrics", "localhost:8081"];
+
+    let mut command = command(&args, &tool_name);
+    let command_name = format!("{:?}", command);
+
+    let status = command
```
EverlastingBugstopper: i'm pretty sure this call returns something that you can actually _call_ status on, don't think `status` is the right variable name here

src/commands/tail/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+        .kill_on_drop(true)
+        .spawn()
+        .expect(&format!("{} failed to spawn", command_name))
+        .await?;
+
+    if !status.success() {
+        failure::bail!(
+            "tried running command:\n{}\nexited with {}",
+            command_name.replace("\"", ""),
```
EverlastingBugstopper: is there a `command_name.display()` function?

src/commands/tail/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+        .request(&CreateTail {
+            account_identifier: &target.account_id,
+            script_name: &target.name,
+            params: CreateTailParams { url },
+        })
+        .await;
+
+    match response {
+        Ok(success) => {
+            eprintln!("Now prepared to stream logs.");
```
EverlastingBugstopper: nit: add emoji

src/commands/tail/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+        }
+        Err(e) => failure::bail!(http::format_error(e, None)),
+    }
+}
+
+async fn print_logs(req: Request<Body>) -> Result<Response<Body>, hyper::Error> {
+    match (req.method(), req.uri().path()) {
+        (&Method::POST, "/") => {
+            let whole_body = hyper::body::to_bytes(req.into_body()).await?;
+            println!("{}", str::from_utf8(&whole_body).unwrap());
```
EverlastingBugstopper: nit: replace unwrap with expect

src/commands/tail/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+
+            Ok(Response::new(Body::from("Success")))
+        }
+        _ => {
+            let mut not_found = Response::default();
+            *not_found.status_mut() = StatusCode::NOT_FOUND;
+            Ok(not_found)
+        }
```
EverlastingBugstopper: when is this case executed? what went wrong? how can the user recover?

src/commands/tail/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+    // regex for extracting url from cloudflared metrics port.
+    let url_regex = Regex::new("userHostname=\"(https://[a-z.-]+)\"").unwrap();
+
+    let mut attempt = 0;
+
+    // This exponential backoff retry loop retries retrieving the cloudflared endpoint url
+    // from the cloudflared /metrics endpoint until it gets the URL or has tried retrieving the URL
+    // over 5 times.
+    while attempt < 5 {
+        if let Ok(resp) = reqwest::get("http://localhost:8081/metrics").await {
```
EverlastingBugstopper: this will need to be not hardcoded at some point

src/commands/tail/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+            let body = resp.text().await?;
+
+            for cap in url_regex.captures_iter(&body) {
+                return Ok(cap[1].to_string());
+            }
+        }
+
+        attempt = attempt + 1;
```
EverlastingBugstopper: style nit
```suggestion
        attempt += 1;
```

src/commands/tail/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+        thread::sleep(Duration::from_millis(attempt * attempt * 100));
```
EverlastingBugstopper: possibly worth adding a comment about why this sleep is necessary and how the value is constructed. seems like it gets exponentially longer as attempts increase? possible performance bottleneck

src/commands/tail/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+async fn send_heartbeat(
+    target: &Target,
+    client: &async_api::Client,
+    tail_id: &str,
+) -> Result<(), failure::Error> {
+    let response = client
+        .request(&SendTailHeartbeat {
+            account_identifier: &target.account_id,
+            script_name: &target.name,
+            tail_id: tail_id,
```
EverlastingBugstopper: nit:
```suggestion
            tail_id,
```

src/install/mod.rs
```diff
@@ -84,6 +84,7 @@ fn tool_needs_update(tool_name: &str, path: &Path) -> Result<bool, failure::Erro
         Some(v) => v,
     };
     let latest_tool_version = get_latest_version(tool_name)?;
+    // todo(gabbi): If latest_tool_version is error, try getting latest version via a github api call.
```
EverlastingBugstopper:
```suggestion
    // TODO: If latest_tool_version is error, try getting latest version via a github api call.
```

src/install/mod.rs
```diff
@@ -159,6 +160,8 @@ fn prebuilt_url(tool_name: &str, owner: &str, version: &str) -> Option<String> {
 }
 
 fn get_latest_version(tool_name: &str) -> Result<String, failure::Error> {
+    // TODO(gabbi): return the latest version pulled from github api, not via Krate.
```
EverlastingBugstopper:
```suggestion
    // TODO: return the latest version pulled from github api, not via Krate.
```

src/main.rs
```diff
@@ -505,6 +505,7 @@ fn run() -> Result<(), failure::Error> {
             "{} Retrieve your user info and test your auth config",
             emoji::SLEUTH
         )))
+        .subcommand(SubCommand::with_name("tail").about("Aggregate logs from production worker"))
```
EverlastingBugstopper: will need to add `.arg(env)` here I believe

src/commands/tail/mod.rs
ashleymichal: TIL 😄

src/commands/tail/mod.rs
gabbifish: BLAST FROM THE PAST ;)! This is a common technique called exponential backoff. It is standard for API calls (whenever a resource request fails and can be re-requested over the network, the time to re-request should exponentially increase to avoid creating packet receive queues which can cause network congestion/bottlenecks: https://en.wikipedia.org/wiki/Exponential_backoff)

ashleymichal: i believe it is covered in a comment earlier, it just requires you already know what exponential backoff is to tie the comment to the sleep call.

src/main.rs
ashleymichal: i think it already works...

src/commands/tail/mod.rs
EverlastingBugstopper: TIL! Never knew about exponential backoff :) (hi gabbi)

src/main.rs
EverlastingBugstopper:
```console
$ wrangler tail --env dev
error: Found argument '--env' which wasn't expected, or isn't valid in this context

USAGE:
    wrangler tail

For more information try --help
```

it might be handled in main.rs but you also have to add the argument explicitly to the long clap chain here

src/commands/tail/mod.rs
ashleymichal: https://github.com/cloudflare/wrangler/issues/1175

src/commands/tail/mod.rs
ashleymichal: https://github.com/cloudflare/wrangler/issues/1161

src/commands/tail/mod.rs
ashleymichal: i will be doing a bit of refactoring so renames are inevitable.

src/commands/tail/mod.rs
ashleymichal: https://github.com/cloudflare/wrangler/issues/1176

src/commands/tail/mod.rs
ashleymichal: this is likely to change in a subsequent refactor.

src/commands/tail/mod.rs
ashleymichal: what emoji?

src/commands/tail/mod.rs
ashleymichal: https://github.com/cloudflare/wrangler/issues/1177

src/commands/tail/mod.rs
ashleymichal: https://github.com/cloudflare/wrangler/issues/1178

src/main.rs
ashleymichal: https://github.com/cloudflare/wrangler/issues/1179

src/commands/tail/mod.rs
ashleymichal: possibly; this is also a thing that may change.

PR #1144, opened 2020-03-17
removes unused logs folder

EverlastingBugstopper:


ashleygwilliams:
do we need to add  that file to .gitignore?

EverlastingBugstopper:
🤷‍♂ i'm not sure how it got there tbh, maybe?

EverlastingBugstopper:
it doesn't seem to reappear in my directory at all so dont think it needs to go in gitignore

PR #1145, opened 2020-03-17
Asset Manifest QOL improvements

EverlastingBugstopper:
- Shrink size of asset manifest by reducing hash length from 64 to 10
- Removes `--verbose` flag in favor of a spinner that tells you what file it's adding to the asset manifest
- Doesn't compute the asset manifest hashes twice (it used to)

spinner_gif

for once my commits are actually (fairly) logically grouped if you want to review one commit at a time

### DIFF
--- a/src/commands/kv/bucket/mod.rs
+++ b/src/commands/kv/bucket/mod.rs
@@ -16,6 +16,7 @@ use data_encoding::HEXLOWER;
 use failure::format_err;
 use ignore::overrides::{Override, OverrideBuilder};
 use ignore::{Walk, WalkBuilder};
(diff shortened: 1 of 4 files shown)


src/commands/kv/bucket/mod.rs
```diff
@@ -82,31 +83,6 @@ pub fn directory_keys_values(
     }
 }
 
-// Returns only the hashed keys for a directory's files.
-fn directory_keys_only(target: &Target, directory: &Path) -> Result<Vec<String>, failure::Error> {
```
ashleymichal: i thought i had already eliminated this; thank you!

PR #1149, opened 2020-03-18
tidy tail file, unescape json log output

ashleymichal:
Lots of 💅 but the heart of this is changing a `{:?}` to a `{}` in the `print_logs` fn so we output unescaped JSON.

* s/todo/TODO/g
* put tail function declarations in order of calling
* s/re/url_regex/g

### DIFF
--- a/src/commands/tail/mod.rs
+++ b/src/commands/tail/mod.rs
@@ -32,10 +32,11 @@ pub fn start_tail(target: &Target, user: &GlobalUser) -> Result<(), failure::Err
         target.name
     );
 
-    // todo: get rid of this gross clone. Maybe just default to static lifetime.
+    // TODO: get rid of this gross clone. Maybe just default to static lifetime.
     runtime.block_on(run_cloudflared_start_server(target.clone(), user.clone()))
 }
 
+// TODO: handle sigint to tear down tunnel, http server
 async fn run_cloudflared_start_server(
     target: Target,
     user: GlobalUser,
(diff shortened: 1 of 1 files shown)


src/commands/tail/mod.rs
```diff
@@ -52,18 +53,31 @@ async fn run_cloudflared_start_server(
     }
 }
 
+async fn start_log_collection_http_server() -> Result<(), failure::Error> {
+    // Start HTTP echo server that prints whatever is posted to it.
+    let addr = ([127, 0, 0, 1], 8080).into();
```
EverlastingBugstopper: `// TODO: Make this configurable`

src/commands/tail/mod.rs
```diff
@@ -52,18 +53,31 @@ async fn run_cloudflared_start_server(
     }
 }
 
+async fn start_log_collection_http_server() -> Result<(), failure::Error> {
```
EverlastingBugstopper:
```suggestion
async fn start_log_collection_http_server() -> ! {
```

if i understand correctly this will never return (unless we start handling sigint/sigkill)

src/commands/tail/mod.rs
```diff
@@ -152,9 +137,25 @@ async fn enable_tailing_start_heartbeat(
     }
 }
 
+async fn print_logs(req: Request<Body>) -> Result<Response<Body>, hyper::Error> {
+    match (req.method(), req.uri().path()) {
+        (&Method::POST, "/") => {
+            let whole_body = hyper::body::to_bytes(req.into_body()).await?;
+            println!("{}", str::from_utf8(&whole_body).unwrap());
```
EverlastingBugstopper: should probably replace the unwrap with a ? if at all possible, at least an expect

src/commands/tail/mod.rs
```diff
@@ -152,9 +137,25 @@ async fn enable_tailing_start_heartbeat(
...
+
+            Ok(Response::new(Body::from("Success")))
```
EverlastingBugstopper: Is this response body used anywhere? What's the reasoning behind not using `Response::default`?

ashleymichal: tbh i didn't change this code from what Gabbi wrote, only re-org'd it.

PR #1169, opened 2020-03-27
allow kv-namespaces AND kv_namespaces

### LINKED ISSUE #1158: make _ and - equivalent in wrangler.toml
It's confusing to have a mix of `_` and `-` in `wrangler.toml`. We should allow either or since we've already introduced inconsistencies (see `kv-namespaces` vs `workers_dev`)

EverlastingBugstopper:
serde is so cool

fixes #1158

will allow both kv-namespaces and kv_namespaces. in the future we should always just use `_` for everything and not use `-`. will submit PR to docs to change examples to be merged after this is released


EverlastingBugstopper:
@ashleymichal maybe you have ideas as to how to test this? From what I understand we just create structs and dont test strings directly

ashleymichal:
@EverlastingBugstopper you should be able to do a unit test in the manifest file itself using the FromStr trait found here: https://github.com/cloudflare/wrangler/blob/master/src/settings/toml/manifest.rs#L333
using raw string literals.

PR #1184, opened 2020-03-31
fix: use checkout@v1

EverlastingBugstopper:


EverlastingBugstopper:
Hey @imbsky - looks like we have to revert the bump to v2 since tests are broken on other branches. Not sure why, especially since tests passed on your initial PR. We are investigating with GitHub and will update again when it doesn't break our tests anymore but will have to go through with revert to get a point release out.

PR #1189, opened 2020-04-03
Alewis/refactor tail

ashleymichal:
This PR refactors the prototype into smaller modules and structs. The structs in this PR still need to be tested.

## Still TODO
- [ ] test structs

I'll list the changes in this PR here by commit message:

## Refactors

**rename tail command to tail::start**
does what it says on the tin

**move async tasks to structs, tail module**
pulls all the logic from `commands/tail/mod.rs` into the root folder `tail` and separates them according the the tasks they are responsible for: `Host` for the local http server; `Tunnel` for the cloudflared child process; and `Session` for establishing and maintaining the tail in the API.

**move tokio runtime into tail mod, add tunnel new**
further abstracts the tokio implementation details from the command; now the command just calls `Tail::run()` and passes it target and user.


src/tail/session.rs
```diff
@@ -0,0 +1,118 @@
...
+
+impl Session {
+    pub async fn run(
+        target: Target,
+        user: GlobalUser,
+        mut rx: Receiver<()>,
+    ) -> Result<(), failure::Error> {
+        let client = http::cf_v4_api_client_async(&user, HttpApiClientConfig::default())?;
+
+        let url = get_tunnel_url().await?;
```
ashleymichal, line 31: TODO: i'd like tunnel to be in charge of its own url, and pass it in to this handler.

src/tail/session.rs
```diff
@@ -0,0 +1,118 @@
...
+            })
+            .await;
+
+        match response {
+            Ok(success) => {
+                eprintln!("Now prepared to stream logs.");
+
+                let tail_id = success.result.id;
+
+                // Loop indefinitely to send "heartbeat" to API and keep log streaming alive.
```
ashleymichal, line 47: likely everything in this function before this loop could be done in an initializer; it doesn't quite matter but idk.

src/tail/session.rs
```diff
@@ -0,0 +1,118 @@
...
+                }
+            }
+            Err(e) => failure::bail!(http::format_error(e, None)),
+        }
+    }
+}
+
+async fn get_tunnel_url() -> Result<String, failure::Error> {
+    // regex for extracting url from cloudflared metrics port.
+    let url_regex = Regex::new("userHostname=\"(https://[a-z.-]+)\"").unwrap();
```
ashleymichal, line 78: TODO: what i'd really like to do is parse the output of the metrics endpoint as a stream (it's just prometheus metrics so there is a way to parse them) rather than using this regex but this is fine for RC.

Bonus points if we don't have to query this metrics endpoint at all but instead could get this as part of the initialization process but stdout is too wordy/subject to change; will try to work with tunnel team to figure out a good way to go about this.

src/tail/session.rs
```diff
@@ -0,0 +1,118 @@
...
+        .request(&SendTailHeartbeat {
+            account_identifier: &target.account_id,
+            script_name: &target.name,
+            tail_id,
+        })
+        .await;
+
+    match response {
+        Ok(_) => Ok(()),
+        Err(e) => failure::bail!(http::format_error(e, None)),
```
ashleymichal, line 116: TODO: write proper error formatter for tail api request errors

src/tail/tunnel.rs
```diff
@@ -0,0 +1,73 @@
...
+        let child = command
+            .spawn()
+            .expect(&format!("{} failed to spawn", command_name));
+
+        Ok(Tunnel { child })
+    }
+
+    pub async fn run(self, rx: Receiver<()>) -> Result<(), failure::Error> {
+        rx.await?;
+        self.shutdown().await
```
ashleymichal, line 38: since this is a child process, it  doesn't need us to do anything to run, but it does need to listen for ctrl-c in order to do a swift shut down.

src/tail/tunnel.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    pub async fn shutdown(mut self) -> Result<(), failure::Error> {
+        // eprintln!("killing cloudflared");
+        if let Err(e) = self.child.kill() {
```
ashleymichal, line 43: TODO: consider a small delay on this kill; since cloudflared is a child process it receives the signal to shut down independently of its parent process (wrangler) but its "graceful" shutdown is relatively long and we do this double tap because it is,  as of now, relatively harmless; we still want to wait on it to exit though hence  the wait_with_output after.

src/tail/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+            let tunnel_process = Tunnel::new()?;
+            let (tunnel_tx, tunnel_rx) = tokio::sync::oneshot::channel();
+            let tunnel_handle = tokio::spawn(tunnel_process.run(tunnel_rx));
+
+            let (session_tx, session_rx) = tokio::sync::oneshot::channel();
+            let session_handle = tokio::spawn(Session::run(target, user, session_rx));
+
+            let txs = vec![log_tx, tunnel_tx, session_tx];
+            let sigint_handle = tokio::spawn(handle_sigint(txs));
+
```
ashleymichal, line 36: TODO: receiving a ctrl-c before the message "Ready to stream logs" hangs. need to handle that problem.

src/tail/host.rs
```diff
@@ -0,0 +1,38 @@
...
+            println!(
+                "{}",
+                std::str::from_utf8(&whole_body).expect("failed to deserialize tail log body")
+            );
+
+            Ok(Response::new(Body::from("Success")))
+        }
+        _ => {
+            let mut not_found = Response::default();
+            *not_found.status_mut() = StatusCode::NOT_FOUND;
```
EverlastingBugstopper, line 34: nit: still think this should be a [400 Bad Request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) instead of a not found

ashleymichal: agreed. https://github.com/cloudflare/wrangler/issues/1196

PR #1190, opened 2020-04-03
feat: message if wrangler needs updating

### LINKED ISSUE #397: Print update message on wrangler usage when new version of wrangler is available.

EverlastingBugstopper:
Fixes #397

every 24 hours, wrangler will ping crates.io and ask for the latest version. if latest version is greater than currently installed version, a message will be printed stating that the version needs updating. it saves the result from crates.io at `$WRANGLER_HOME/version.toml` which includes the latest version and the time it was last checked.


victoriabernard92:
Since there aren't tests yet could we add an undocumented/temp option to turn this off incase it breaks anyone? It could live in ` $WRANGLER_HOME/config/default.toml` as 
```
api_token="0xbd.."
warn_update="false"
```
We could remove this later once we know users aren't irritated by it

Just an idea. LGTM either way

src/util/version.rs
```diff
@@ -0,0 +1,127 @@
...
+pub fn background_check_for_updates() -> mpsc::Receiver<Version> {
+    let (sender, receiver) = mpsc::channel();
+
+    let _detached_thread = thread::spawn(move || match check_wrangler_versions() {
+        Ok(wrangler_versions) => {
+            if wrangler_versions.current != wrangler_versions.latest {
+                let _ = sender.send(wrangler_versions.latest);
+            }
+        }
+        Err(e) => log::debug!("could not determine if update is needed:\n{}", e),
```
victoriabernard92, line 123:
```suggestion
        Err(e) => log::debug!("could not connect to crates.io to check up to date with latest version:\n{}", e),
```

victoriabernard92: Always a good idea to include a best next step for the user to take even if it's ignoring the message 
e.g.
```
ignore this message if the output of wrangler -V matches that at https://crates.io/api/v1/crates/wrangler
```

EverlastingBugstopper: this log is not provided to the user unless they set `RUST_LOG=debug`

EverlastingBugstopper:
so the only thing that needs to be done with this is making it only message once every 24 hours. right now, if your version of wrangler needs updating, it will print a message telling you to update after _every single command_. this shouldn't happen. Let's only bug folks about it once a day.

src/version/mod.rs
```diff
@@ -0,0 +1,5 @@
+mod block;
```
ashleymichal, line 1: this module got moved into the wranglerjs module in a previous commit; we should be able to remove this and just have  `src/version.rs`

src/version/version.rs
```diff
@@ -0,0 +1,155 @@
...
+    };
+
+    Ok(WranglerVersion {
+        current,
+        latest,
+        checked,
+    })
+}
+
+/// Reads version out of version file, is `None` if file does not exist
```
ashleymichal:
```suggestion
/// Reads version out of version file, is `None` if file does not exist or is corrupted
```

src/version/wrangler_version.rs
```diff
@@ -0,0 +1,155 @@
...
+#[derive(Deserialize, Debug)]
+struct CrateInformation {
+    max_version: String,
+}
+
+pub fn background_check_for_updates() -> mpsc::Receiver<Version> {
+    let (sender, receiver) = mpsc::channel();
+
+    let _detached_thread = thread::spawn(move || match check_wrangler_versions() {
+        Ok(wrangler_versions) => {
```
ashleymichal, line 144: nit: why plural?

src/version/wrangler_version.rs
```diff
@@ -0,0 +1,155 @@
...
+            let time_since_last_checked =
+                current_time.duration_since(last_checked_version.last_checked)?;
+
+            if time_since_last_checked.as_secs() < ONE_DAY {
+                checked = true;
+            }
+            Version::parse(&last_checked_version.latest_version)?
+        }
+        // If version.toml doesn't exist, fetch latest version
+        None => get_latest_version(&current.to_string(), &version_file, current_time)?,
```
ashleymichal, line 76: this is  the only place we  call `get_latest_version`;  is that  right?

src/version/mod.rs
```diff
@@ -0,0 +1,157 @@
+use std::fs;
```
ashleymichal, line 1: last nit: if this is only `version/mod.rs` it can just be `version.rs`

src/version/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+use std::path::PathBuf;
+use std::str::FromStr;
+use std::sync::mpsc;
+use std::thread;
+use std::time::SystemTime;
+
+use crate::settings::global_user::get_wrangler_home_dir;
```
ashleymichal, line 8: nit: this probably does not  belong  to global_user,  but idk where to  put it just now.

Issue #1177, opened 2020-03-31
better error handling / messaging


ashleymichal:
There's a lot of places where tails can go wrong. We need to handle each of them gracefully and with clear instructions to users on how to correct the issues.

- [x] handle case where worker has not yet been published (should be an api error code)

- [x] handle missing cloudflared: https://github.com/cloudflare/wrangler/issues/1207#issuecomment-615338332

ashleymichal:
closed by #1247

PR #1195, opened 2020-04-06
Alewis/handle sigint

ashleymichal:
# Overview

Previously, if you `ctrl-c`'d out of `wrangler tail`, wrangler would leave an orphaned child process for `cloudflared` spinning on a "graceful shutdown". This PR introduces changes that catch `ctrl-c` and manually send a second `kill` signal to the child process, as well as gracefully stopping the localhost server and discontinuing heartbeat messages to the Workers API. There's also a little clean up task in here that starts the log server before joining all the async tasks, which removes the need to "sleep" in the Tunnel task before spinning up cloudflared.

## Still TODO based on usage testing:
- [x] handle ctrl-c before  `try_join`

I'll list the changes in this PR here by commit message:

## Changes

**add host new method to initialize server before tunnel**
previously the tunnel function would sleep for a short time to ensure that the http server had started up before it tried to spin up a tunnel pointing to it. in this pr, i move the port binding step to an initializer function on `Host` and call it before running it as a task, ensuring that the server already exists before Tunnel tries to do its thing.

**handle ctrl_c nicely using channels**
this is the meat of the PR, and a lot happens here.
* rather than passing all futures directly to `tokio::try_join` to run concurrently, we instead `spawn` each task and pass their JoinHandles in to `try_join`. This is based on the following instruction from the [tokio documentation](https://docs.rs/tokio/0.2.15/tokio/macro.try_join.html#runtime-characteristics):
> If parallelism is required, spawn each async expression using tokio::spawn and pass the join handle to `try_join!`.
* introduce a fourth task, `handle_sigint`, whose sole purpose is to listen for `ctrl-c` and send messages to each of the other tasks so that they might shut down properly.
* introduce three `oneshot` channels and pass the transmitters to `handle_sigint`, and their receivers to each task; each task handles this in its own special way:
  * `Host` uses the receiver in its callback to `with_graceful_shutdown`, which is a convenience method provided by hyper but I assume it is to complete in flight requests etc before exiting.
  * `Tunnel` uses the receiver to "double-tap" the kill call to the cloudflared process and wait on its output before exiting. If the double-tap turns out to be premature, we can always put a delay on it.
  * `Session` now uses a `Delay` rather than `thread::sleep` to space out heartbeat calls, and all of its execution code is wrapped in a call to `rx.try_recv`, which will cause the task to exit upon receipt of the signal from our sigint handler.

### DIFF
--- a/src/tail/host.rs
+++ b/src/tail/host.rs
@@ -2,6 +2,7 @@ use hyper::server::conn::AddrIncoming;
 use hyper::server::Builder;
 use hyper::service::{make_service_fn, service_fn};
 use hyper::{Body, Method, Request, Response, Server, StatusCode};
+use tokio::sync::oneshot::Receiver;
 
 pub struct Host {
     server: Builder<AddrIncoming>,
@@ -17,10 +18,16 @@ impl Host {
         Ok(Host { server })
     }
 
-    pub async fn run(self) -> Result<(), failure::Error> {
+    pub async fn run(self, rx: Receiver<()>) -> Result<(), failure::Error> {
         let service = make_service_fn(|_| async { Ok::<_, hyper::Error>(service_fn(print_logs)) });
 
-        self.server.serve(service).await?;
+        let server = self.server.serve(service);
+
+        let graceful = server.with_graceful_shutdown(async {
+            rx.await.ok();
+        });
+
+        graceful.await?;
 
         Ok(())
     }
--- a/src/tail/host.rs
+++ b/src/tail/host.rs
@@ -6,22 +6,25 @@ use tokio::sync::oneshot::Receiver;
 
 pub struct Host {
     server: Builder<AddrIncoming>,
+    rx: Receiver<()>,
 }
 
 impl Host {
-    pub fn new() -> Result<Host, failure::Error> {
+    pub fn new(rx: Receiver<()>) -> Host {
         // Start HTTP echo server that prints whatever is posted to it.
         let addr = ([127, 0, 0, 1], 8080).into();
 
         let server = Server::bind(&addr);
 
-        Ok(Host { server })
+        Host { server, rx }
     }
 
-    pub async fn run(self, rx: Receiver<()>) -> Result<(), failure::Error> {
+    pub async fn run(self) -> Result<(), failure::Error> {
         let service = make_service_fn(|_| async { Ok::<_, hyper::Error>(service_fn(print_logs)) });
 
         let server = self.server.serve(service);
+
+        let rx = self.rx;
 
         let graceful = server.with_graceful_shutdown(async {
             rx.await.ok();
--- a/src/tail/host.rs
+++ b/src/tail/host.rs
@@ -1,18 +1,26 @@
+use hyper::server::conn::AddrIncoming;
+use hyper::server::Builder;
 use hyper::service::{make_service_fn, service_fn};
 use hyper::{Body, Method, Request, Response, Server, StatusCode};
 
-pub struct Host;
+pub struct Host {
+    server: Builder<AddrIncoming>,
+}
 
 impl Host {
-    pub async fn run() -> Result<(), failure::Error> {
+    pub fn new() -> Result<Host, failure::Error> {
         // Start HTTP echo server that prints whatever is posted to it.
         let addr = ([127, 0, 0, 1], 8080).into();
 
+        let server = Server::bind(&addr);
+
+        Ok(Host { server })
+    }
+
+    pub async fn run(self) -> Result<(), failure::Error> {
         let service = make_service_fn(|_| async { Ok::<_, hyper::Error>(service_fn(print_logs)) });
 
-        let server = Server::bind(&addr).serve(service);
-
-        server.await?;
+        self.server.serve(service).await?;
 
         Ok(())
     }
--- a/src/tail/log_server.rs
+++ b/src/tail/log_server.rs
@@ -6,19 +6,19 @@ use tokio::sync::oneshot::Receiver;
 
 pub struct LogServer {
     server: Builder<AddrIncoming>,
-    rx: Receiver<()>,
+    shutdown_rx: Receiver<()>,
 }
 
 /// LogServer is just a basic HTTP server running locally; it listens for POST requests on the root
 /// path and simply prints the JSON body of each request as its own line to STDOUT.
 impl LogServer {
-    pub fn new(rx: Receiver<()>) -> LogServer {
+    pub fn new(shutdown_rx: Receiver<()>) -> LogServer {
         // Start HTTP echo server that prints whatever is posted to it.
         let addr = ([127, 0, 0, 1], 8080).into();
 
         let server = Server::bind(&addr);
 
-        LogServer { server, rx }
+        LogServer { server, shutdown_rx }
     }
 
     pub async fn run(self) -> Result<(), failure::Error> {
@@ -28,7 +28,7 @@ impl LogServer {
 
         // The shutdown receiver listens for a one shot message from our sigint handler as a signal
         // to gracefully shut down the hyper server.
-        let shutdown_rx = self.rx;
+        let shutdown_rx = self.shutdown_rx;
 
         let graceful = server.with_graceful_shutdown(async {
             shutdown_rx.await.ok();
@@ -44,8 +44,6 @@ async fn print_logs(req: Request<Body>) -> Result<Response<Body>, hyper::Error> 
     match (req.method(), req.uri().path()) {
         (&Method::POST, "/") => {
             let whole_body = hyper::body::to_bytes(req.into_body()).await?;
-            // TODO: at some point I would like to send this body as a message to a central writer
-            // rather than just using println!, but this works just fine for now.
             println!(
                 "{}",
                 std::str::from_utf8(&whole_body).expect("failed to deserialize tail log body")
--- a/src/tail/log_server.rs
+++ b/src/tail/log_server.rs
@@ -4,19 +4,21 @@ use hyper::service::{make_service_fn, service_fn};
 use hyper::{Body, Method, Request, Response, Server, StatusCode};
 use tokio::sync::oneshot::Receiver;
 
-pub struct Host {
+pub struct LogServer {
     server: Builder<AddrIncoming>,
     rx: Receiver<()>,
 }
 
-impl Host {
-    pub fn new(rx: Receiver<()>) -> Host {
+/// LogServer is just a basic HTTP server running locally; it listens for POST requests on the root
+/// path and simply prints the JSON body of each request as its own line to STDOUT.
+impl LogServer {
+    pub fn new(rx: Receiver<()>) -> LogServer {
         // Start HTTP echo server that prints whatever is posted to it.
         let addr = ([127, 0, 0, 1], 8080).into();
 
         let server = Server::bind(&addr);
 
-        Host { server, rx }
+        LogServer { server, rx }
     }
 
     pub async fn run(self) -> Result<(), failure::Error> {
@@ -24,10 +26,12 @@ impl Host {
 
         let server = self.server.serve(service);
 
-        let rx = self.rx;
+        // The shutdown receiver listens for a one shot message from our sigint handler as a signal
+        // to gracefully shut down the hyper server.
+        let shutdown_rx = self.rx;
 
         let graceful = server.with_graceful_shutdown(async {
-            rx.await.ok();
+            shutdown_rx.await.ok();
         });
 
         graceful.await?;
@@ -40,6 +44,8 @@ async fn print_logs(req: Request<Body>) -> Result<Response<Body>, hyper::Error> 
     match (req.method(), req.uri().path()) {
         (&Method::POST, "/") => {
             let whole_body = hyper::body::to_bytes(req.into_body()).await?;
+            // TODO: at some point I would like to send this body as a message to a central writer
+            // rather than just using println!, but this works just fine for now.
             println!(
                 "{}",
                 std::str::from_utf8(&whole_body).expect("failed to deserialize tail log body")
--- a/src/tail/log_server.rs
+++ b/src/tail/log_server.rs
@@ -18,7 +18,10 @@ impl LogServer {
 
         let server = Server::bind(&addr);
 
-        LogServer { server, shutdown_rx }
+        LogServer {
+            server,
+            shutdown_rx,
+        }
     }
 
     pub async fn run(self) -> Result<(), failure::Error> {
--- a/src/tail/mod.rs
+++ b/src/tail/mod.rs
@@ -1,5 +1,5 @@
-/// The Tail feature for Wrangler is the Workers ecosystem answer to the need for live
-/// log collection in production. When a user runs `wrangler tail`, several things happen:
+/// `wrangler tail` allows Workers users to collect logs from their deployed Workers.
+/// When a user runs `wrangler tail`, several things happen:
 ///     1. A simple HTTP server (LogServer) is started and begins listening for requests on localhost:8080
 ///     2. An [Argo Tunnel](https://developers.cloudflare.com/argo-tunnel/) instance (Tunnel) is started
 ///        using [cloudflared](https://developers.cloudflare.com/argo-tunnel/downloads/), exposing the
@@ -9,7 +9,7 @@
 ///     4. The Workers API binds the URL to a [Trace Worker], and directs all `console` and
 ///        exception logging to the Trace Worker, which POSTs each batch of logs as a JSON
 ///        payload to the provided Tunnel URL.
-///     5. Upon receipt. the LogServer prints the payload of each POST request to STDOUT.
+///     5. Upon receipt, the LogServer prints the payload of each POST request to STDOUT.
 mod log_server;
 mod session;
 mod tunnel;
@@ -64,13 +64,14 @@ impl Tail {
     }
 }
 
-/// handle_sigint simply waits on a ctrl_c from the system and sends messages to each registered
+/// handle_sigint waits on a ctrl_c from the system and sends messages to each registered
 /// transmitter when it is received.
 async fn listen_for_sigint(txs: Vec<oneshot::Sender<()>>) -> Result<(), failure::Error> {
     tokio::signal::ctrl_c().await?;
     for tx in txs {
-        // if send errors, it is because the receiver has been de-allocated, and that's okay
-        // because we were just telling it to shut down anyway.
+        // if `tx.send()` returns an error, it is because the receiver has gone out of scope,
+        // likely due to the task returning early for some reason, in which case we don't need
+        // to tell that task to shut down because it already has.
         tx.send(()).ok();
     }
 
--- a/src/tail/mod.rs
+++ b/src/tail/mod.rs
@@ -1,8 +1,19 @@
-mod host;
+/// The Tail feature for Wrangler is the Workers ecosystem answer to the need for live
+/// log collection in production. When a user runs `wrangler tail`, several things happen:
+///     1. A simple HTTP server (LogServer) is started and begins listening for requests on localhost:8080
+///     2. An [Argo Tunnel](TODO) instance (Tunnel) is started using [cloudflared](TODO), exposing the
+///        LogServer to the internet on a randomly generated URL.
+///     3. Wrangler initiates a tail Session by making a request to the Workers API /tail endpoint,
+///        providing the Tunnel URL as an argument.
+///     4. The Workers API binds the URL to a [Trace Worker], and directs all `console` and
+///        exception logging to the Trace Worker, which POSTs each batch of logs as a JSON
+///        payload to the provided Tunnel URL.
+///     5. Upon receipt. the LogServer prints the payload of each POST request to STDOUT.
+mod log_server;
 mod session;
 mod tunnel;
 
-use host::Host;
+use log_server::LogServer;
 use session::Session;
 use tunnel::Tunnel;
 
@@ -20,27 +31,29 @@ impl Tail {
         let mut runtime = TokioRuntime::new()?;
 
         runtime.block_on(async {
-            // Create three one-shot channels for handling ctrl-c
+            // Create three [one-shot](https://docs.rs/tokio/0.2.16/tokio/sync#oneshot-channel)
+            // channels for handling ctrl-c. Each channel has two parts:
+            // tx: Transmitter
+            // rx: Receiver
             let (log_tx, log_rx) = oneshot::channel();
-            let (session_tx, session_rx) = tokio::sync::oneshot::channel();
-            let (tunnel_tx, tunnel_rx) = tokio::sync::oneshot::channel();
+            let (session_tx, session_rx) = oneshot::channel();
+            let (tunnel_tx, tunnel_rx) = oneshot::channel();
 
             // Pass the three transmitters to a newly spawned sigint handler
             let txs = vec![log_tx, tunnel_tx, session_tx];
-            let sigint_handle = tokio::spawn(handle_sigint(txs));
+            let listener = tokio::spawn(listen_for_sigint(txs));
 
             // Spin up a local http server to receive logs
-            let log_host = Host::new(log_rx);
-            let host_handle = tokio::spawn(log_host.run());
+            let log_server = tokio::spawn(LogServer::new(log_rx).run());
 
             // Spin up a new cloudflared tunnel to connect trace worker to local server
             let tunnel_process = Tunnel::new()?;
-            let tunnel_handle = tokio::spawn(tunnel_process.run(tunnel_rx));
+            let tunnel = tokio::spawn(tunnel_process.run(tunnel_rx));
 
             // Register the tail with the Workers API and send periodic heartbeats
-            let session_handle = tokio::spawn(Session::run(target, user, session_rx));
+            let session = tokio::spawn(Session::run(target, user, session_rx));
 
-            let res = tokio::try_join!(sigint_handle, host_handle, session_handle, tunnel_handle,);
+            let res = tokio::try_join!(listener, log_server, session, tunnel);
 
             match res {
                 Ok(_) => Ok(()),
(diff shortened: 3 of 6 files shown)


src/tail/mod.rs
```diff
@@ -19,12 +20,27 @@ impl Tail {
         let mut runtime = TokioRuntime::new()?;
 
         runtime.block_on(async {
+            // Create three one-shot channels for handling ctrl-c
+            let (log_tx, log_rx) = oneshot::channel();
+            let (session_tx, session_rx) = tokio::sync::oneshot::channel();
+            let (tunnel_tx, tunnel_rx) = tokio::sync::oneshot::channel();
```
EverlastingBugstopper, line 26:
```suggestion
            let (session_tx, session_rx) = oneshot::channel();
            let (tunnel_tx, tunnel_rx) = oneshot::channel();
```

src/tail/mod.rs
```diff
@@ -19,12 +20,27 @@ impl Tail {
...
+
+            // Pass the three transmitters to a newly spawned sigint handler
+            let txs = vec![log_tx, tunnel_tx, session_tx];
```
EverlastingBugstopper, line 29:
```suggestion
            let transmitters = vec![log_transmitter, tunnel_transmitter, session_transmitter];
```

src/tail/mod.rs
```diff
@@ -19,12 +20,27 @@ impl Tail {
...
+            let sigint_handle = tokio::spawn(handle_sigint(txs));
```
EverlastingBugstopper, line 30:
```suggestion
            let sigint_handler = tokio::spawn(handle_sigint(txs));
```

src/tail/host.rs
```diff
@@ -1,18 +1,36 @@
+use hyper::server::conn::AddrIncoming;
+use hyper::server::Builder;
 use hyper::service::{make_service_fn, service_fn};
 use hyper::{Body, Method, Request, Response, Server, StatusCode};
+use tokio::sync::oneshot::Receiver;
 
-pub struct Host;
+pub struct Host {
```
EverlastingBugstopper, line 7: not sure what `Host` is supposed to be, worth adding a comment or make it a more descriptive struct name

src/tail/mod.rs
```diff
@@ -33,3 +49,14 @@ impl Tail {
         })
     }
 }
+
+async fn handle_sigint(txs: Vec<oneshot::Sender<()>>) -> Result<(), failure::Error> {
+    tokio::signal::ctrl_c().await?;
+    for tx in txs {
+        if let Err(e) = tx.send(()) {
+            eprintln!("failed to transmit to channel {:?}", e);
```
EverlastingBugstopper, line 57: need a better error message here or panic since it's an unrecoverable bug

src/tail/session.rs
```diff
@@ -36,41 +44,66 @@ impl Session {
...
     }
 }
 
-async fn get_tunnel_url() -> Result<String, failure::Error> {
+async fn get_tunnel_url(rx: &mut Receiver<()>) -> Result<String, failure::Error> {
     // regex for extracting url from cloudflared metrics port.
     let url_regex = Regex::new("userHostname=\"(https://[a-z.-]+)\"").unwrap();
 
     let mut attempt = 0;
+    let mut delay = time::delay_for(Duration::from_millis(attempt * attempt * 1000));
```
EverlastingBugstopper, line 80: worth adding a comment about exponential backoff

src/tail/session.rs
```diff
@@ -36,41 +44,66 @@ impl Session {
...
     while attempt < 5 {
-        if let Ok(resp) = reqwest::get("http://localhost:8081/metrics").await {
-            let body = resp.text().await?;
+        match rx.try_recv() {
+            Err(TryRecvError::Empty) => {
+                if delay.is_elapsed() {
+                    if let Ok(resp) = reqwest::get("http://localhost:8081/metrics").await {
+                        let body = resp.text().await?;
+
+                        for cap in url_regex.captures_iter(&body) {
```
EverlastingBugstopper, line 92:
```suggestion
                        for match in url_regex.captures_iter(&body) {
```

src/tail/session.rs
EverlastingBugstopper: also, this will be 0 always since you just set attempt to 0.

src/tail/session.rs
```diff
@@ -36,41 +44,66 @@ impl Session {
...
+                            let url = cap[1].to_string();
+                            return Ok(url);
+                        }
+                    }
 
-            for cap in url_regex.captures_iter(&body) {
-                return Ok(cap[1].to_string());
+                    let duration = attempt * attempt;
+                    attempt = attempt + 1;
```
EverlastingBugstopper, line 99:
```suggestion
                    attempt += 1;
```

src/tail/host.rs
ashleymichal: yeah it may be better to call this `LogServer`? it's the "localhost" you point your tunnel to that then receives the log requests from the trace worker. i almost think it is worth putting a really long comment description in `tail/mod.rs` and at the top of each component for this to outline the architecture.

src/tail/session.rs
ashleymichal: the first one will, yes, but since rust doesn't have do while i have to set it as something, and on the first time i don't want a delay. in the while loop i then update to set it again each time the request fails.

src/tail/mod.rs
ashleymichal: technically these are all `JoinHandle`s but i also sort of hate calling things by their mechanical names...

src/tail/mod.rs
ashleymichal: i'm going to link this in my larger error handling issue, because i need to look at the various types of errors i could receive here.

src/tail/host.rs
EverlastingBugstopper: definitely worth the comment at least

src/tail/mod.rs
ashleymichal: https://github.com/cloudflare/wrangler/issues/1177

src/tail/log_server.rs
```diff
@@ -0,0 +1,62 @@
...
+use hyper::{Body, Method, Request, Response, Server, StatusCode};
+use tokio::sync::oneshot::Receiver;
+
+pub struct LogServer {
+    server: Builder<AddrIncoming>,
+    rx: Receiver<()>,
+}
+
+/// LogServer is just a basic HTTP server running locally; it listens for POST requests on the root
+/// path and simply prints the JSON body of each request as its own line to STDOUT.
```
ashleymichal, line 13: there's nothing preventing some other process from making requests to this server or the tunnel; this is unlikely to become an issue but may be worth calling out.

src/tail/log_server.rs
```diff
@@ -0,0 +1,62 @@
...
+
+        Ok(())
+    }
+}
+
+async fn print_logs(req: Request<Body>) -> Result<Response<Body>, hyper::Error> {
+    match (req.method(), req.uri().path()) {
+        (&Method::POST, "/") => {
+            let whole_body = hyper::body::to_bytes(req.into_body()).await?;
+            // TODO: at some point I would like to send this body as a message to a central writer
```
ashleymichal, line 47: i actually don't like that i put this TODO inline, i'd rather turn it into an issue...

src/tail/mod.rs
```diff
@@ -1,13 +1,25 @@
-mod host;
+/// The Tail feature for Wrangler is the Workers ecosystem answer to the need for live
+/// log collection in production. When a user runs `wrangler tail`, several things happen:
+///     1. A simple HTTP server (LogServer) is started and begins listening for requests on localhost:8080
+///     2. An [Argo Tunnel](TODO) instance (Tunnel) is started using [cloudflared](TODO), exposing the
```
ashleymichal, line 4: need to add links to argo tunnel documentation.

src/tail/mod.rs
```diff
@@ -19,12 +31,29 @@ impl Tail {
         let mut runtime = TokioRuntime::new()?;
 
         runtime.block_on(async {
+            // Create three [one-shot](https://docs.rs/tokio/0.2.16/tokio/sync#oneshot-channel)
+            // channels for handling ctrl-c. Each channel has two parts:
+            // tx: Transmitter
+            // rx: Receiver
```
ashleymichal, line 37: i may turn this inside out and create a "subscribe" method on the sigint handler but this works fine for now.

src/tail/session.rs
```diff
@@ -13,13 +15,23 @@ use crate::http;
...
 use crate::settings::toml::Target;
 
+const KEEP_ALIVE_INTERVAL: u64 = 60;
+
 pub struct Session;
 
+/// Session is responsible for interacting with the Workers API to establish and maintain the tail
+/// connection; once the tail session is established, it uses the returned tail_id to send periodic
+/// "keepalive" heartbeats to the service. If the service does not receive a heartbeat for ~10
+/// minutes, it will kill the tail session by removing the trace worker.
```
ashleymichal, line 25: > it will kill the tail session by removing the trace worker.

currently, wrangler has no way of knowing if this happens.

src/tail/session.rs
```diff
@@ -36,41 +48,96 @@ impl Session {
...
-                    // through other means.
                 }
             }
             Err(e) => failure::bail!(http::format_error(e, None)),
         }
     }
 }
 
-async fn get_tunnel_url() -> Result<String, failure::Error> {
+async fn get_tunnel_url(rx: &mut Receiver<()>) -> Result<String, failure::Error> {
```
ashleymichal, line 79: ultimately i would really like the Tunnel to be responsible for its own URL

src/tail/session.rs
```diff
@@ -36,41 +48,96 @@ impl Session {
...
     // regex for extracting url from cloudflared metrics port.
     let url_regex = Regex::new("userHostname=\"(https://[a-z.-]+)\"").unwrap();
```
ashleymichal, line 81: TODO: parse metrics output line by line using prometheus rules, get this based on the field rather than this regex.

src/tail/session.rs
```diff
@@ -36,41 +48,96 @@ impl Session {
...
+                attempt: 0,
+                max_attempts,
+            }
+        }
 
-            for cap in url_regex.captures_iter(&body) {
-                return Ok(cap[1].to_string());
+        // our retry delay implements an [exponential backoff](TODO), which simply waits twice as
+        // long between each attempt.
+        fn reset(self) -> RetryDelay {
```
ashleymichal, line 102: this acts a little bit functional because mutating the struct causes more problems in the loop than overwriting it.

src/tail/session.rs
```diff
@@ -36,41 +48,96 @@ impl Session {
...
+            RetryDelay {
+                delay: delay_for(Duration::from_millis(0)),
+                attempt: 0,
+                max_attempts,
+            }
+        }
 
-            for cap in url_regex.captures_iter(&body) {
-                return Ok(cap[1].to_string());
+        // our retry delay implements an [exponential backoff](TODO), which simply waits twice as
```
ashleymichal, line 100: TODO: link to resources on exponential backoff. probably just wikipedia.z

src/tail/tunnel.rs
```diff
@@ -34,28 +32,26 @@ impl Tunnel {
...
 
-        Ok(Tunnel {
-            child,
-            command_name,
-        })
+        Ok(Tunnel { child })
+    }
+
+    pub async fn run(self, rx: Receiver<()>) -> Result<(), failure::Error> {
+        rx.await?;
```
ashleymichal, line 42: rename this to `shutdown_rx`

src/tail/tunnel.rs
```diff
@@ -34,28 +32,26 @@ impl Tunnel {
...
+        self.shutdown().await
     }
 
-    pub async fn run(self) -> Result<(), failure::Error> {
-        let status = self.child.await?;
+    pub async fn shutdown(mut self) -> Result<(), failure::Error> {
+        // eprintln!("killing cloudflared");
```
ashleymichal, line 47: remove comment

src/tail/tunnel.rs
```diff
@@ -34,28 +32,26 @@ impl Tunnel {
...
+        if let Err(e) = self.child.kill() {
+            failure::bail!("failed to kill cloudflared: {}", e)
```
ashleymichal, line 49: this shouldn't happen but if it does we want to know. cloudflared will eventually exit, but we could also instruct the user to just `kill <child.id>`

src/tail/session.rs
```diff
@@ -13,13 +15,23 @@ use crate::http;
...
+/// connection; once the tail session is established, it uses the returned tail_id to send periodic
+/// "keepalive" heartbeats to the service. If the service does not receive a heartbeat for ~10
+/// minutes, it will kill the tail session by removing the trace worker.
 impl Session {
-    pub async fn run(target: &Target, user: &GlobalUser) -> Result<(), failure::Error> {
-        let client = http::cf_v4_api_client_async(user, HttpApiClientConfig::default())?;
+    pub async fn run(
+        target: Target,
+        user: GlobalUser,
+        mut rx: Receiver<()>,
```
ashleymichal, line 30: rename this `shutdown_rx`

src/tail/mod.rs
```diff
@@ -1,13 +1,25 @@
-mod host;
+/// The Tail feature for Wrangler is the Workers ecosystem answer to the need for live
+/// log collection in production. When a user runs `wrangler tail`, several things happen:
```
EverlastingBugstopper, line 2:
```suggestion
/// `wrangler tail` allows Workers users to collect logs from their deployed Workers
/// When a user runs `wrangler tail`, several things happen:
```

src/tail/mod.rs
```diff
@@ -1,13 +1,25 @@
...
+///     1. A simple HTTP server (LogServer) is started and begins listening for requests on localhost:8080
+///     2. An [Argo Tunnel](TODO) instance (Tunnel) is started using [cloudflared](TODO), exposing the
+///        LogServer to the internet on a randomly generated URL.
+///     3. Wrangler initiates a tail Session by making a request to the Workers API /tail endpoint,
+///        providing the Tunnel URL as an argument.
+///     4. The Workers API binds the URL to a [Trace Worker], and directs all `console` and
+///        exception logging to the Trace Worker, which POSTs each batch of logs as a JSON
+///        payload to the provided Tunnel URL.
+///     5. Upon receipt. the LogServer prints the payload of each POST request to STDOUT.
```
EverlastingBugstopper, line 11:
```suggestion
///     5. Upon receipt the LogServer prints the payload of each POST request to STDOUT.
```

src/tail/mod.rs
```diff
@@ -19,12 +31,29 @@ impl Tail {
...
+            // Create three [one-shot](https://docs.rs/tokio/0.2.16/tokio/sync#oneshot-channel)
+            // channels for handling ctrl-c. Each channel has two parts:
+            // tx: Transmitter
+            // rx: Receiver
+            let (log_tx, log_rx) = oneshot::channel();
+            let (session_tx, session_rx) = oneshot::channel();
+            let (tunnel_tx, tunnel_rx) = oneshot::channel();
+
+            // Pass the three transmitters to a newly spawned sigint handler
+            let txs = vec![log_tx, tunnel_tx, session_tx];
```
EverlastingBugstopper, line 43: TEXAS

src/tail/mod.rs
```diff
@@ -19,12 +31,29 @@ impl Tail {
...
             let tunnel_process = Tunnel::new()?;
-            let res = tokio::try_join!(
-                Host::run(),
-                tunnel_process.run(),
-                Session::run(&target, &user)
-            );
+            let tunnel = tokio::spawn(tunnel_process.run(tunnel_rx));
+
+            // Register the tail with the Workers API and send periodic heartbeats
+            let session = tokio::spawn(Session::run(target, user, session_rx));
```
EverlastingBugstopper, line 54:
```suggestion
            let keep_alive = tokio::spawn(Session::run(target, user, session_rx));
```

maybe keep_alive isn't the right name here but `session` feels... vague?

src/tail/mod.rs
```diff
@@ -33,3 +62,16 @@ impl Tail {
         })
     }
 }
+
+/// handle_sigint simply waits on a ctrl_c from the system and sends messages to each registered
```
EverlastingBugstopper, line 66:
```suggestion
/// handle_sigint waits on a ctrl_c from the system and sends messages to each registered
```

src/tail/mod.rs
```diff
@@ -33,3 +62,16 @@ impl Tail {
...
+/// transmitter when it is received.
+async fn listen_for_sigint(txs: Vec<oneshot::Sender<()>>) -> Result<(), failure::Error> {
+    tokio::signal::ctrl_c().await?;
+    for tx in txs {
+        // if send errors, it is because the receiver has been de-allocated, and that's okay
```
EverlastingBugstopper, line 71: this comment doesn't _quite_ make sense to me. i think i get it but might want some clearer wording

src/tail/session.rs
ashleymichal: https://github.com/cloudflare/wrangler/issues/1200

src/tail/session.rs
```diff
@@ -36,41 +48,96 @@ impl Session {
...
+                                    send_heartbeat(&target, &client, &tail_id).await;
+                                if heartbeat_result.is_err() {
+                                    return heartbeat_result;
+                                }
+                                delay = delay_for(duration);
+                            }
+                        }
+                        _ => {
+                            return Ok(());
+                        }
```
EverlastingBugstopper, line 70: i'm interested in this - the flow isn't super obvious to me, why is the heartbeat sent on this error and why do we stop if nothing is received?

src/tail/session.rs
EverlastingBugstopper, line 100:
```suggestion
        // our retry delay implements an [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff), which simply waits twice as
```

src/tail/session.rs
```diff
@@ -36,41 +48,96 @@ impl Session {
...
+                delay: delay_for(Duration::from_millis(0)),
+                attempt: 0,
+                max_attempts,
+            }
+        }
 
-            for cap in url_regex.captures_iter(&body) {
-                return Ok(cap[1].to_string());
+        // our retry delay implements an [exponential backoff](TODO), which simply waits twice as
+        // long between each attempt.
```
EverlastingBugstopper, line 101: what does this comment mean?

src/tail/session.rs
```diff
@@ -36,41 +48,96 @@ impl Session {
...
+        }
+    }
+
+    let mut delay = RetryDelay::new(5);
+
+    while !delay.expired() {
+        match rx.try_recv() {
+            Err(TryRecvError::Empty) => {
+                if delay.is_elapsed() {
+                    if let Ok(resp) = reqwest::get("http://localhost:8081/metrics").await {
```
EverlastingBugstopper, line 125: should make issue to make this configurable and also handle the error case (what happens if localhost:8081/metrics doesn't respond?

src/tail/tunnel.rs
```diff
@@ -1,29 +1,27 @@
...
 
 pub struct Tunnel {
     child: Child,
-    command_name: String,
 }
 
+/// Tunnel wraps a child process that runs cloudflared and forwards requests from the Trace Worker
+/// in the runtime to our local LogServer instance. We wrap it in a struct primarily to hold the
+/// state of the child process so that upon receipt of a SIGINT message we can more swiftly kill it
+/// and wait on its output; otherwise we end up leaving it behind when wrangler exits and this
```
EverlastingBugstopper, line 18:
```suggestion
/// and wait on its output; otherwise we leave an orphaned process when wrangler exits and this
```

src/tail/tunnel.rs
```diff
@@ -1,29 +1,27 @@
...
+/// causes problems if it still exists the next time we start up a tail.
```
EverlastingBugstopper, line 19: these problems should still be handled, what happens if the user spams ctrl+c and graceful shutdown doesn't happen? should make new issue

src/tail/tunnel.rs
EverlastingBugstopper, line 47:
```suggestion
```

src/tail/log_server.rs
```diff
@@ -0,0 +1,62 @@
...
+pub struct LogServer {
+    server: Builder<AddrIncoming>,
+    rx: Receiver<()>,
+}
+
+/// LogServer is just a basic HTTP server running locally; it listens for POST requests on the root
+/// path and simply prints the JSON body of each request as its own line to STDOUT.
+impl LogServer {
+    pub fn new(rx: Receiver<()>) -> LogServer {
+        // Start HTTP echo server that prints whatever is posted to it.
```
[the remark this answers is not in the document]
ashleymichal: yeah i thought about this myself; i'm not sure how to restrict it so i think that's sort of a "needs design" issue, but definitely an issue. https://github.com/cloudflare/wrangler/issues/1201

src/tail/log_server.rs
```diff
@@ -0,0 +1,62 @@
...
+
+async fn print_logs(req: Request<Body>) -> Result<Response<Body>, hyper::Error> {
+    match (req.method(), req.uri().path()) {
+        (&Method::POST, "/") => {
+            let whole_body = hyper::body::to_bytes(req.into_body()).await?;
+            // TODO: at some point I would like to send this body as a message to a central writer
+            // rather than just using println!, but this works just fine for now.
+            println!(
+                "{}",
+                std::str::from_utf8(&whole_body).expect("failed to deserialize tail log body")
```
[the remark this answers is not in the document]
ashleymichal: i would guess that it will go to stderr. i'm fine with this doing a full panic right now, since this isn't even parsing JSON it's just expecting utf8 streaming input and if we're getting something else there's a problem.

src/tail/session.rs
ashleymichal: is just giving an explanation for why the duration is set the way it is; added a link to wikipedia in case someone is unfamiliar with exponential backoff.

src/tail/session.rs
ashleymichal: there is already an issue for configurable ports. if there is no response it will keep retrying; really if any outcome happens that isn't the response being okay and containing the url, it will keep retrying until it runs out of retries.

src/tail/session.rs
EverlastingBugstopper: "long between each attempt"

src/tail/tunnel.rs
ashleymichal: so fun fact, what we do here is actually the same as sending another ctrl+c... there's really no time for the user to do it more than once.

src/tail/tunnel.rs
ashleymichal: have this locally will push asap.

src/tail/log_server.rs
dhaynespls: I don't think there would be. Let's assume a malicious actor has root access to a victims machine and deploys a binary that sends requests to this server, what can they do? All the server does is print out any POST requests to stdout, so I guess they could spam your terminal but otherwise it doesn't look like they could pivot to other functionality.

src/tail/session.rs
ashleymichal: scroll over? "which simply waits twice as long between each attempt"

src/tail/log_server.rs
```diff
@@ -0,0 +1,62 @@
...
+            );
+
+            Ok(Response::new(Body::from("Success")))
+        }
+        _ => {
+            let mut not_found = Response::default();
+            *not_found.status_mut() = StatusCode::NOT_FOUND;
+            Ok(not_found)
```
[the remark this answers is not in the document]
ashleymichal: this actually does not work here.

src/tail/mod.rs
ashleymichal: adding this.

src/tail/mod.rs
ashleymichal: idk what else to call it. it's the tail session with the api.

src/tail/log_server.rs
EverlastingBugstopper: guess if they had root access you've got bigger problems, thanks David :)

src/tail/mod.rs
ashleymichal: if the receiver goes out of scope, it's likely due to the task returning early for some reason, in which case we don't need to tell it to shut down because it already has. will use this verbiage in the comment.

src/tail/session.rs
ashleymichal: i'll add a comment

src/tail/session.rs
ashleymichal: added

src/tail/session.rs
```diff
@@ -89,7 +89,7 @@ async fn get_tunnel_url(shutdown_rx: &mut Receiver<()>) -> Result<String, failur
     // This retry loop retries retrieving the cloudflared endpoint url from the cloudflared /metrics
     // until it gets the URL or has tried retrieving the URL over 5 times.
     impl RetryDelay {
-        fn new(max_attempts: u64) -> Self {
+        fn new(max_attempts: u64) -> RetryDelay {
```
EverlastingBugstopper, line 98: was always interested in `Self` vs `${StructName}` - I think I actually like `Self` better so you don't have to update a billion instances but won't block on

src/tail/session.rs
```diff
@@ -55,6 +55,12 @@ impl Session {
 
                 loop {
                     match shutdown_rx.try_recv() {
+                        // this variant of the [TryRecvError](https://docs.rs/tokio/0.2.16/tokio/sync/oneshot/error/enum.TryRecvError.html)
+                        // occurs when the receiver listens for a message and the channel is empty;
+                        // in this case, it means that we have not received a shutdown command and
+                        // can continue with our task. The other variant would indicate that the
+                        // sender has been dropped, in which case we want to follow shut down as if
+                        // we received the signal.
```
EverlastingBugstopper, line 63: thank you - this makes a lot more sense now

PR #1202, opened 2020-04-09
feat: accept ipv4/6 for default dev

### LINKED ISSUE #1198: `wrangler dev` seems to be only listening on ipv6?
#### Environment

* operating system: macOS 10.15.3
* output of `rustc -V`: not using Rust, installed Wrangler from NPM
* output of `node -v`: v10.19.0
* output of `wrangler -V`: 👷 ✨  wrangler 1.8.4

#### Steps to reproduce

In one tab:
```
wrangler dev
```

In another:
```
▶ curl --ipv4 http://localhost:8787
curl: (7) Failed to connect to localhost port 8787: Connection refused
```

The worker

Connection refused message

This one totally baffled me for a while until I realized I had my script configured to hit `127.0.0.1` and `wrangler dev` was only listening on IPv6. Is this intentional? I can work around it if it is.

EDIT: Looks like this can be worked around with `--ip 127.0.0.1`.

EverlastingBugstopper:
Fixes #1198 

This is not a breaking change, just makes `wrangler dev` listen on both IPv4 and IPv6 

`curl http://localhost:8787` always worked
`curl --ipv4 http://localhost:8787` didn't work before and now does

_**juicy deets**_

From the [Wikipedia article on localhost](https://en.wikipedia.org/wiki/Localhost)
> The name localhost normally resolves to the IPv4 loopback address 127.0.0.1, and to the IPv6 loopback address ::1

Unfortunately for us, `localhost` was only resolving to the IPv6 loopback address `::1` and not the IPv4 address `127.0.0.1`. Switching the input from `localhost` to `[::]` makes our hyper server listen on both IPv4 and IPv6.

### DIFF
--- a/src/commands/dev/server_config/listening_address.rs
+++ b/src/commands/dev/server_config/listening_address.rs
@@ -19,7 +19,7 @@ impl ListeningAddress {
     }
 
     fn as_str(&self) -> String {
-        self.address.to_string().replace("[::1]", "localhost")
+        self.address.to_string().replace("[::]", "localhost")
     }
 }
 
--- a/src/commands/dev/server_config/mod.rs
+++ b/src/commands/dev/server_config/mod.rs
@@ -17,7 +17,7 @@ impl ServerConfig {
         port: Option<&str>,
     ) -> Result<Self, failure::Error> {
         let port = port.unwrap_or("8787");
-        let ip = ip.unwrap_or("localhost");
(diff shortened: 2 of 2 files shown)


EverlastingBugstopper:
unfortunately listening on `[::]` means that other computers on the same network can send requests to your worker if they know your private ip address so I dont think we can merge this as is

PR #1203, opened 2020-04-10
Alewis/handle api errors

ashleymichal:
there's a lot i want to clean up about this, like the way channels are kind of created and chucked all over the place in the main runtime procedure, but this at least gets us to where wrangler will exit on an API error. next we'll add nicer error messages.

### DIFF
--- a/src/tail/mod.rs
+++ b/src/tail/mod.rs
@@ -17,12 +17,11 @@ mod tunnel;
 
 use log_server::LogServer;
 use session::Session;
-use shutdown::{Signal, ShutdownHandler };
+use shutdown::ShutdownHandler;
 use tunnel::Tunnel;
 
 use tokio;
 use tokio::runtime::Runtime as TokioRuntime;
-use tokio::sync::oneshot;
 
 use crate::settings::global_user::GlobalUser;
 use crate::settings::toml::Target;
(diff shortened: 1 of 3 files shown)


src/tail/mod.rs
```diff
@@ -52,28 +53,19 @@ impl Tail {
...
-            let session = tokio::spawn(Session::run(target, user, session_rx));
+            let session = tokio::spawn(Session::run(target, user, session_rx, tx));
 
-            let res = tokio::try_join!(listener, log_server, session, tunnel);
+            let res = tokio::try_join!(
+                async { listener.await? },
+                async { log_server.await? },
+                async { session.await? },
+                async { tunnel.await? }
+            );
```
EverlastingBugstopper, line 63: wwwtttttffffff

src/tail/mod.rs
```diff
@@ -52,28 +53,19 @@ impl Tail {
...
 
             match res {
                 Ok(_) => Ok(()),
-                Err(e) => failure::bail!(e),
+                Err(e) => Err(e),
```
EverlastingBugstopper, line 67: this is interesting

ashleymichal: i'm not 100% certain this was necessary. there was a lot of shotgunning on this one.

src/tail/mod.rs
ashleymichal: yeah so this is effectively the thing i was talking about in the tokio-users discord where i was told that actually those join handles created by `tokio::spawn` wrap the future in another result/future, so this pulls them out. it's bonkers and imho try_join documentation should be clearer about this, or the macro should do this for you, but perhaps there is some utility to this approach. either way it didn't make sense for me to remove the parallelism provided by spawning new tasks, so this is what happened. probably deserves a comment though.

PR #1212, opened 2020-04-16
Add support for printing api token email on `whoami`

### LINKED ISSUE #863: `wrangler whoami` cannot print out email for users using API tokens
Because the /user Cloudflare v4 API endpoint doesn't support API tokens yet, we can't retrieve a user's email using their API token. This means `wrangler whoami` won't be able to provide the expected email output for API token users (an error message informing users of this is currently printed out).

dhaynespls:
- query the same api endpoint as we do with the global token, but now with the api token
- return errors, print helpful statement if the api token doesn't have the right permission

Closes #863

### DIFF
--- a/src/commands/whoami/mod.rs
+++ b/src/commands/whoami/mod.rs
@@ -1,19 +1,32 @@
 use crate::http;
 use crate::settings::global_user::GlobalUser;
-use crate::terminal::emoji;
+use crate::terminal::{emoji, message};
 
 use cloudflare::endpoints::account::{self, Account};
+use cloudflare::endpoints::user::GetUserDetails;
 use cloudflare::framework::apiclient::ApiClient;
+use cloudflare::framework::response::ApiFailure;
 
 use prettytable::{Cell, Row, Table};
 
 pub fn whoami(user: &GlobalUser) -> Result<(), failure::Error> {
-    // If using email + API key for auth, simply prints out email from config file.
+    // Attempt to print email for both GlobalKeyAuth and TokenAuth users
     let auth: String = match user {
         GlobalUser::GlobalKeyAuth { email, .. } => {
             format!("a Global API Key, associated with the email '{}'", email,)
         }
-        GlobalUser::TokenAuth { .. } => "an API Token".to_string(),
+        GlobalUser::TokenAuth { .. } => {
+            let token_auth_email: String = fetch_api_token_email(user)?;
+
+            if token_auth_email == "" {
+                "an API Token".to_string()
+            } else {
+                format!(
+                    "an API Token, associated with the email '{}'",
+                    token_auth_email,
+                )
+            }
+        }
     };
 
     println!("\n{} You are logged in with {}.\n", emoji::WAVING, auth,);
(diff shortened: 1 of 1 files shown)


src/commands/whoami/mod.rs
```diff
@@ -23,6 +36,26 @@ pub fn whoami(user: &GlobalUser) -> Result<(), failure::Error> {
...
+        Ok(res) => Ok(res.result.email),
+        Err(e) => match e {
+            ApiFailure::Error(_, api_errors) => {
+                let error = &api_errors.errors[0];
+                if error.code == 9109 {
+                    message::info("Your token is missing the 'User Details: Read' permission.\n\nPlease generate and auth with a new token that has these perms to be able to identify this token.\n");
+                    Ok("".to_string())
+                } else {
+                    Ok("".to_string())
+                }
```
EverlastingBugstopper, line 52:
```suggestion
                if error.code == 9109 {
                    message::info("Your token is missing the 'User Details: Read' permission.\n\nPlease generate and auth with a new token that has these perms to be able to identify this token.\n");
                }
                Ok(None)
```

src/commands/whoami/mod.rs
```diff
@@ -23,6 +36,26 @@ pub fn whoami(user: &GlobalUser) -> Result<(), failure::Error> {
     Ok(())
 }
 
+fn fetch_api_token_email(user: &GlobalUser) -> Result<String, failure::Error> {
```
EverlastingBugstopper, line 39:
```suggestion
fn fetch_api_token_email(user: &GlobalUser) -> Result<Option<String>, failure::Error> {
```

src/commands/whoami/mod.rs
```diff
@@ -23,6 +36,26 @@ pub fn whoami(user: &GlobalUser) -> Result<(), failure::Error> {
...
+    let client = http::cf_v4_client(user)?;
+    let response = client.request(&GetUserDetails {});
+    match response {
+        Ok(res) => Ok(res.result.email),
```
EverlastingBugstopper, line 43:
```suggestion
        Ok(res) => Ok(Some(res.result.email)),
```

src/commands/whoami/mod.rs
```diff
@@ -39,7 +72,7 @@ fn format_accounts(user: &GlobalUser, accounts: Vec<Account>) -> Table {
 
     if let GlobalUser::TokenAuth { .. } = user {
         if accounts.is_empty() {
-            println!("Your token is missing the 'Account Settings: Read' permission.\n\nPlease generate and auth with a new token that has these perms to be able to list your accounts.\n");
+            message::info("Your token is missing the 'Account Settings: Read' permission.\n\nPlease generate and auth with a new token that has these perms to be able to list your accounts.\n");
```
EverlastingBugstopper, line 75:
```suggestion
            message::billboard("Your token is missing the 'Account Settings: Read' permission.\n\nPlease generate and auth with a new token that has these perms to be able to list your accounts.");
```

src/commands/whoami/mod.rs
```diff
@@ -1,19 +1,32 @@
...
 pub fn whoami(user: &GlobalUser) -> Result<(), failure::Error> {
-    // If using email + API key for auth, simply prints out email from config file.
+    // Attempt to print email for both GlobalKeyAuth and TokenAuth users
     let auth: String = match user {
         GlobalUser::GlobalKeyAuth { email, .. } => {
             format!("a Global API Key, associated with the email '{}'", email,)
         }
-        GlobalUser::TokenAuth { .. } => "an API Token".to_string(),
+        GlobalUser::TokenAuth { .. } => {
+            let token_auth_email: String = fetch_api_token_email(user)?;
```
EverlastingBugstopper, line 19:
```suggestion
            let token_auth_email = fetch_api_token_email(user)?;
```

src/commands/whoami/mod.rs
```diff
@@ -1,19 +1,32 @@
...
+
+            if token_auth_email == "" {
+                "an API Token".to_string()
+            } else {
+                format!(
+                    "an API Token, associated with the email '{}'",
+                    token_auth_email,
+                )
+            }
```
EverlastingBugstopper, line 28:
```suggestion
            if let Some(token_auth_email) = token_auth_email {
                format!(
                    "an API Token, associated with the email '{}'",
                    token_auth_email,
                )
            } else {
                "an API Token".to_string()
            }
```

EverlastingBugstopper: it's more idiomatic to use `Option::None` instead of assigning an empty string for an empty result

src/commands/whoami/mod.rs
EverlastingBugstopper: this will print the info in a nice big attention-grabbing box

PR #1213, opened 2020-04-17
reduce hover

ashleymichal:
there was an ugly here that caused Session to enter an infinite loop; rather than just continuously checking if the delay is elapsed now we await it. because async.

### DIFF
--- a/src/tail/session.rs
+++ b/src/tail/session.rs
@@ -62,13 +62,12 @@ impl Session {
                 let mut delay = delay_for(duration);
 
                 loop {
-                    if delay.is_elapsed() {
-                        let heartbeat_result = send_heartbeat(&target, &client, &tail_id).await;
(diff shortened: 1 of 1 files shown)


src/tail/session.rs
```diff
@@ -62,13 +62,12 @@ impl Session {
                 let mut delay = delay_for(duration);
 
                 loop {
-                    if delay.is_elapsed() {
-                        let heartbeat_result = send_heartbeat(&target, &client, &tail_id).await;
-                        if heartbeat_result.is_err() {
-                            return heartbeat_result;
-                        }
-                        delay = delay_for(duration);
+                    delay.await;
```
ashleymichal, line 65: the difference here is simply the  number of times  we enter this loop; before it was  just constant and we kept running the  `.is_elapsed()` check; now we  enter the loop once per  delay/heartbeat.

PR #1220, opened 2020-04-21
refactor and document wrangler dev

EverlastingBugstopper:
This PR shouldn't change any behavior but it does add a decent chunk of comments, remove dependence on the `futures` crate, and paves the way for easily adding edge connectivity.

### DIFF
--- a/src/commands/dev/gcs/headers.rs
+++ b/src/commands/dev/gcs/headers.rs
@@ -1,5 +1,15 @@
+/// The preview service runs on cloudflareworkers.com and then sends parts
+/// of the incoming request to the `request` in the Workers Runtime
+///
+/// The way it does this is by prefixing the _real_ headers sent to
+/// and returned by the worker with this header prefix
 const HEADER_PREFIX: &str = "cf-ew-raw-";
-
+/// Any request headers sent to `wrangler dev` must be prefixed
+/// before sending it to the preview service
+/// and any response headers sent from the preview service
+/// that don't have the prefix must be removed
+/// and response headers that do have the prefix
+/// must have the prefix stripped
 use std::str::FromStr;
 
 use hyper::header::{HeaderMap, HeaderName};
@@ -7,15 +17,21 @@ use hyper::http::request::Parts as RequestParts;
 use hyper::http::response::Parts as ResponseParts;
 use hyper::http::status::StatusCode;
 
+/// modify an incoming request before sending it to the preview service
 pub fn structure_request(parts: &mut RequestParts) {
     prepend_request_headers_prefix(parts)
 }
 
+/// modify a response from the preview service before returning it to the user
 pub fn destructure_response(parts: &mut ResponseParts) -> Result<(), failure::Error> {
     set_response_status(parts)?;
     strip_response_headers_prefix(parts)
 }
 
+/// every header sent to `wrangler dev` must be prefixed
+/// before sending it along to the preview service
+/// so it is sent along to the Workers runtime and not
+/// consumed directly by the preview service
 fn prepend_request_headers_prefix(parts: &mut RequestParts) {
     let mut headers: HeaderMap = HeaderMap::new();
 
@@ -29,6 +45,12 @@ fn prepend_request_headers_prefix(parts: &mut RequestParts) {
     parts.headers = headers;
 }
 
+/// every header returned by the Worker is prefixed
+/// and headers without a prefix are specific to the preview service
+/// here we parse those headers and construct a new header map
+///
+/// discard headers without that prefix
+/// strip the prefix from real Workers headers
 fn strip_response_headers_prefix(parts: &mut ResponseParts) -> Result<(), failure::Error> {
     let mut headers = HeaderMap::new();
 
@@ -45,13 +67,16 @@ fn strip_response_headers_prefix(parts: &mut ResponseParts) -> Result<(), failur
     Ok(())
 }
 
+/// parse the response status from headers sent by the preview service
+/// and apply the parsed result to mutable ResponseParts
 fn set_response_status(parts: &mut ResponseParts) -> Result<(), failure::Error> {
     let status = parts
         .headers
         .get("cf-ew-status")
         .expect("Could not determine status code of response");
-    // status will be "404 not found" or "200 ok"
-    // we need to split that string to create hyper's status code
+
+    // `status` above will be "404 not found" or "200 ok"
+    // we need to parse that string to create hyper's status code
     let status_vec: Vec<&str> = status.to_str()?.split(' ').collect();
     parts.status = StatusCode::from_str(status_vec[0])?;
     Ok(())
--- a/src/commands/dev/gcs/mod.rs
+++ b/src/commands/dev/gcs/mod.rs
@@ -1,0 +1,83 @@
+mod headers;
+mod server;
+mod setup;
+mod watch;
+
+use server::serve;
+use setup::{get_preview_id, get_session_id};
+use watch::watch_for_changes;
+
+use crate::commands::dev::{socket, ServerConfig};
+use crate::settings::global_user::GlobalUser;
+use crate::settings::toml::Target;
+
+use std::sync::{Arc, Mutex};
+use std::thread;
+
+use tokio::runtime::Runtime as TokioRuntime;
+
+/// spin up a local server that routes requests to the preview service
+/// that has a Cloudflare Workers runtime without access to zone-specific features
+pub fn dev(
+    target: Target,
+    user: Option<GlobalUser>,
+    server_config: ServerConfig,
+    verbose: bool,
+) -> Result<(), failure::Error> {
+    // setup the session
+    let session_id = get_session_id()?;
+
+    // upload the initial script
+    let preview_id = get_preview_id(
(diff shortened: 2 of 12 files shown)


src/commands/dev/gcs/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+    runtime.block_on(async {
+        let devtools_listener = tokio::spawn(socket::listen(session_id.to_string()));
+        let server = tokio::spawn(serve(server_config, Arc::clone(&preview_id)));
+        let res = tokio::try_join!(async { devtools_listener.await? }, async { server.await? });
+
+        match res {
+            Ok(_) => Ok(()),
+            Err(e) => Err(e),
+        }
+    })
```
EverlastingBugstopper, line 82: this is sorta a new way of doing things, copied the way @ashleymichal implemented for wrangler tail

src/commands/dev/gcs/setup.rs
```diff
@@ -0,0 +1,38 @@
...
+    user: Option<GlobalUser>,
+    server_config: &ServerConfig,
+    session_id: &str,
+    verbose: bool,
+) -> Result<String, failure::Error> {
+    // setting sites_preview to `true` would print a message to the terminal
+    // directing the user to open the browser to view the output
+    // this message makes sense for `wrangler preview` but not `wrangler dev`
+    let sites_preview = false;
+    let script_id = upload(&mut target, user.as_ref(), sites_preview, verbose).map_err(|_| failure::format_err!("Could not upload your script. Check your internet connection or https://www.cloudflarestatus.com/ for rare incidents impacting the Cloudflare Workers API."))?;
```
EverlastingBugstopper, line 30: i've gotta fix this, we're clobbering a bunch of unrelated error messages here

src/commands/dev/socket.rs
```diff
@@ -42,26 +52,30 @@ pub async fn listen(session_id: &str) -> Result<(), failure::Error> {
...
-        interval.tick().await;
+        delay.await;
         let keep_alive_message = protocol::runtime::SendMethod::GetIsolateId(id.into());
         let keep_alive_message = serde_json::to_string(&keep_alive_message)
             .expect("Could not convert keep alive message to JSON");
         let keep_alive_message = Message::Text(keep_alive_message);
-        tx.unbounded_send(keep_alive_message).unwrap();
+        tx.send(keep_alive_message).unwrap();
         id += 1;
+        delay = delay_for(duration);
```
EverlastingBugstopper, line 79: this also mirrors how @ashleymichal implemented heartbeats in `wrangler tail` - functionally equivalent but moved for consistency.

src/commands/dev/socket.rs
```diff
@@ -42,26 +52,30 @@ pub async fn listen(session_id: &str) -> Result<(), failure::Error> {
...
     };
-    pin_mut!(keep_alive_to_ws, print_ws_messages);
-    future::select(keep_alive_to_ws, print_ws_messages).await;
-    Ok(())
+
+    // run the heartbeat and message printer in parallel
+    tokio::select! {
+        _ = keep_alive_to_ws => { Ok(()) }
+        _ = print_ws_messages => { Ok(()) }
+    }
```
EverlastingBugstopper, line 60: this is cleaner than using future::select and pin_mut

src/commands/dev/socket.rs
```diff
@@ -22,15 +25,22 @@ pub async fn listen(session_id: &str) -> Result<(), failure::Error> {
...
     let enable_runtime = protocol::runtime::SendMethod::Enable(1.into());
     let enable_runtime = serde_json::to_string(&enable_runtime)?;
     let enable_runtime = Message::Text(enable_runtime);
     write.send(enable_runtime).await?;
 
-    let (keep_alive_tx, keep_alive_rx) = futures::channel::mpsc::unbounded();
+    // if left unattended, the preview service will kill the socket
+    // that emits console messages
+    // send a keep alive message every so often in the background
+    let (keep_alive_tx, keep_alive_rx) = mpsc::unbounded_channel();
```
EverlastingBugstopper, line 39: use a tokio mpsc channel instead of the one from futures

src/commands/dev/gcs/headers.rs
```diff
@@ -7,15 +7,19 @@ use hyper::http::request::Parts as RequestParts;
 use hyper::http::response::Parts as ResponseParts;
```
ashleymichal, line 7: this page could  maybe  benefit from a top level comment explaining why headers need to be  handled a special way. that may be clearer than comments  on individual fns

src/commands/dev/gcs/server.rs
```diff
@@ -0,0 +1,133 @@
...
+    let client = HyperClient::builder().build::<_, Body>(https);
+
+    let listening_address = server_config.listening_address.clone();
+
+    // create a closure that hyper will use later to handle HTTP requests
+    let make_service = make_service_fn(move |_| {
+        let client = client.to_owned();
+        let server_config = server_config.to_owned();
+        let preview_id = preview_id.to_owned();
+        async move {
```
ashleymichal, line 36: mebbe not here, but i'd suggest pulling these nested anonymous functions into named functions for clarity. not 100% necessary tho.

src/commands/dev/gcs/server.rs
```diff
@@ -0,0 +1,133 @@
...
+    server_config: ServerConfig,
+    preview_id: Arc<Mutex<String>>,
+) -> Result<(), failure::Error> {
+    // set up https client to connect to the preview service
+    let https = HttpsConnector::new();
+    let client = HyperClient::builder().build::<_, Body>(https);
+
+    let listening_address = server_config.listening_address.clone();
+
+    // create a closure that hyper will use later to handle HTTP requests
```
ashleymichal, line 31: but what does the closure do?

src/commands/dev/gcs/setup.rs
```diff
@@ -0,0 +1,38 @@
...
+/// `wrangler dev` session
+pub(super) fn get_session_id() -> Result<String, failure::Error> {
+    Ok(Uuid::new_v4().to_simple().to_string())
+}
+
+/// upload the script to the Workers API, and combine its response
+/// with the session id to get the preview ID
+///
+/// this is used when sending requests to the Workers Runtime
+/// so it executes the correct Worker
```
ashleymichal, line 18: this is a great explainer comment 👍

src/commands/dev/gcs/watch.rs
```diff
@@ -23,7 +23,12 @@ pub fn watch_for_changes(
         let target = target.clone();
         commands::build(&target)?;
 
+        // acquire the lock
         let mut preview_id = preview_id.lock().unwrap();
+
+        // while holding the lock, assign a new preview id
+        // this allows the server to route subsequent requests
+        // to the proper script
```
ashleymichal, line 31: i think somewhere else in the code you explain what "the lock" is but it's probably worth explaining why there's a lock here, too; another instance of "why > what"

src/commands/dev/mod.rs
```diff
@@ -43,170 +20,24 @@ pub fn dev(
...
-    let sites_preview = false;
-    let script_id = upload(&mut target, user.as_ref(), sites_preview, verbose).map_err(|_| failure::format_err!("Could not upload your script. Check your internet connection or https://www.cloudflarestatus.com/ for rare incidents impacting the Cloudflare Workers API."))?;
-    Ok(format!(
-        "{}{}{}{}",
-        &script_id,
-        session_id,
-        server_config.host.is_https() as u8,
-        server_config.host
-    ))
+fn print_alpha_warning_message() {
```
ashleymichal, line 37: not sure if it matters as much here, but we may  consider  thinking more closely about  when we print to stdout vs stderr; this is for sure a  thing in tail logs but could use further  consideration across the codebase.

src/commands/dev/socket.rs
ashleymichal: note that the utility of `tokio::select` is to run until one of its blocks returns and then abandon the other. not certain that's the precise intention here, perhaps  `try_join` would be more in  line  with intent?

src/commands/dev/socket.rs
```diff
@@ -22,15 +25,22 @@ pub async fn listen(session_id: &str) -> Result<(), failure::Error> {
...
     tokio::spawn(keep_alive(keep_alive_tx));
     let keep_alive_to_ws = keep_alive_rx.map(Ok).forward(write);
```
ashleymichal, line 41: this  is interesting

src/commands/dev/socket.rs
EverlastingBugstopper: nah - select is what we want here, printing ws messages will be the first one to go because keep_alive never returns. want them to just run indefinitely

src/commands/dev/gcs/server.rs
EverlastingBugstopper: I actually spent like an hour trying to figure out how to do this but couldn't because of the `move` statements. Happy for some help on this one

src/commands/dev/socket.rs
ashleymichal: hrm. k.

Issue #1200, opened 2020-04-08
refactor: get url by parsing metrics instead of by regex


ashleymichal:
TODO: parse metrics output line by line using prometheus rules, get this based on the field rather than this regex.

EverlastingBugstopper:
Link is broken

Issue #1161, opened 2020-03-24
add progress spinner to tunnel setup message


ashleymichal:
nit: should add emoji or perhaps a loading spinner

_Originally posted by @EverlastingBugstopper in https://github.com/cloudflare/wrangler/pull/1131_

PR #1223, opened 2020-04-23
Better port selection

### LINKED ISSUE #1178: parameterize host name for tunnel/log server
this will need to be not hardcoded at some point

_Originally posted by @EverlastingBugstopper in https://github.com/cloudflare/wrangler/diffs_

ashleymichal:
* provide flags for both tunnel and metrics server port configuration
* if a port is explicitly set, fail if it is unavailable
* if no port is set, try a default, else request one from the OS

closes #1178

### DIFF
--- a/src/commands/tail/mod.rs
+++ b/src/commands/tail/mod.rs
@@ -1,15 +1,53 @@
+use std::net::{SocketAddr, TcpListener};
+
 use crate::settings::global_user::GlobalUser;
 use crate::settings::toml::Target;
 use crate::tail::Tail;
 
-pub fn start(target: &Target, user: &GlobalUser) -> Result<(), failure::Error> {
-    // Note that we use eprintln!() throughout this file; this is because we want any
+const DEFAULT_TUNNEL_PORT: u16 = 8080;
+const DEFAULT_METRICS_PORT: u16 = 8081;
+
+pub fn start(
+    target: &Target,
+    user: &GlobalUser,
+    tunnel_port: Option<u16>,
+    metrics_port: Option<u16>,
+) -> Result<(), failure::Error> {
+    let tunnel_port = find_open_port(tunnel_port, DEFAULT_TUNNEL_PORT)?;
+    let metrics_port = find_open_port(metrics_port, DEFAULT_METRICS_PORT)?;
+
+    // Note that we use eprintln!() throughout this module; this is because we want any
     // helpful output to not be mixed with actual log JSON output, so we use this macro
     // to print messages to stderr instead of stdout (where log output is printed).
     eprintln!(
-        "Setting up log streaming from Worker script \"{}\" to Wrangler. This may take a few seconds...",
-        target.name
+        "Setting up log streaming from Worker script \"{}\". Using ports {} and {}. This may take a few seconds...",
+        target.name,
+        tunnel_port,
+        metrics_port,
     );
 
-    Tail::run(target.clone(), user.clone())
+    Tail::run(target.clone(), user.clone(), tunnel_port, metrics_port)
+}
+
+/// Find open port takes two arguments: an Optional requested port, and a default port.
+fn find_open_port(requested: Option<u16>, default: u16) -> Result<u16, failure::Error> {
+    if let Some(port) = requested {
+        let addr = format!("127.0.0.1:{}", port);
+        match TcpListener::bind(addr) {
+            Ok(socket) => Ok(socket.local_addr()?.port()),
+            Err(_) => failure::bail!("the specified port {} is unavailable", port),
+        }
+    } else {
+        // try to use the default port; else get an ephemeral port from the OS
+        let addrs = [
+            SocketAddr::from(([127, 0, 0, 1], default)),
+            // Binding to port 0 effectively asks the OS to provide the next available
+            // ephemeral port: https://www.grc.com/port_0.htm
+            SocketAddr::from(([127, 0, 0, 1], 0)),
+        ];
+
+        let socket = TcpListener::bind(&addrs[..])?;
+
+        Ok(socket.local_addr()?.port())
+    }
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -524,6 +524,19 @@ fn run() -> Result<(), failure::Error> {
                         .long("env")
                         .takes_value(true)
                 )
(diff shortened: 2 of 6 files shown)


src/commands/tail/mod.rs
```diff
@@ -1,15 +1,51 @@
...
+
+/// Find open port takes two arguments: an Optional requested port, and a default port.
+fn find_open_port(requested: Option<u16>, default: u16) -> Result<u16, failure::Error> {
+    if let Some(port) = requested {
+        let addr = format!("127.0.0.1:{}", port);
+        match TcpListener::bind(addr) {
+            Ok(socket) => Ok(socket.local_addr()?.port()),
+            Err(_) => failure::bail!("the specified port {} is unavailable", port),
+        }
+    } else {
```
ashleymichal, line 40: if the user did not specify a port, just provision one for  them. try to use a default but don't block on it.

src/tail/session.rs
```diff
@@ -78,7 +85,7 @@ impl Session {
     }
 }
 
-async fn get_tunnel_url() -> Result<String, failure::Error> {
+async fn get_tunnel_url(metrics_port: u16) -> Result<String, failure::Error> {
```
ashleymichal, line 88: this is the last straw. my next refactor will move this tunnel url fetching into the tunnel struct.

src/commands/tail/mod.rs
```diff
@@ -1,15 +1,53 @@
...
+const DEFAULT_METRICS_PORT: u16 = 8081;
+
+pub fn start(
+    target: &Target,
+    user: &GlobalUser,
+    tunnel_port: Option<u16>,
+    metrics_port: Option<u16>,
+) -> Result<(), failure::Error> {
+    let t_port = find_open_port(tunnel_port, DEFAULT_TUNNEL_PORT)?;
+    let m_port = find_open_port(metrics_port, DEFAULT_METRICS_PORT)?;
```
EverlastingBugstopper, line 17:
```suggestion
    let tunnel_port = find_open_port(tunnel_port, DEFAULT_TUNNEL_PORT)?;
    let metrics_port = find_open_port(metrics_port, DEFAULT_METRICS_PORT)?;
```

src/commands/tail/mod.rs
```diff
@@ -1,15 +1,53 @@
...
+
+    // Note that we use eprintln!() throughout this module; this is because we want any
     // helpful output to not be mixed with actual log JSON output, so we use this macro
     // to print messages to stderr instead of stdout (where log output is printed).
     eprintln!(
-        "Setting up log streaming from Worker script \"{}\" to Wrangler. This may take a few seconds...",
-        target.name
+        "Setting up log streaming from Worker script \"{}\". Using ports {} and {}. This may take a few seconds...",
```
EverlastingBugstopper, line 23: may want to change this copy to "up to a minute" or something? it's usually more than like 2-3 seconds

src/commands/tail/mod.rs
EverlastingBugstopper: beautiful

src/main.rs
```diff
@@ -524,6 +524,19 @@ fn run() -> Result<(), failure::Error> {
                         .long("env")
                         .takes_value(true)
                 )
+                .arg(
+                    Arg::with_name("tunnel_port")
+                        .help("port to accept tail log requests")
+                        .short("p")
+                        .long("port")
```
EverlastingBugstopper, line 531: why does this one get `port` and the other one gets `metrics`?

i think i'd make them `tunnel_port` and `metrics_port` without a `.short` for either of them

src/main.rs
```diff
@@ -897,10 +910,16 @@ fn run() -> Result<(), failure::Error> {
...
-
         let user = settings::global_user::GlobalUser::new()?;
 
-        commands::tail::start(&target, &user)?;
+        let tunnel_port: Option<u16> = matches
+            .value_of("tunnel_port")
+            .map(|p| p.parse().expect("--port expects a number"));
+        let metrics_port: Option<u16> = matches
+            .value_of("metrics_port")
+            .map(|p| p.parse().expect("--metrics expects a number"));
```
EverlastingBugstopper, line 920: can't wait until clap 3 comes out and we can make serde do all of this stuff up front

src/tail/session.rs
```diff
@@ -28,18 +28,25 @@ impl Session {
...
+    async fn start(
+        target: Target,
+        user: GlobalUser,
+        tx: Sender<()>,
+        metrics_port: u16,
+    ) -> Result<(), failure::Error> {
         let client = http::cf_v4_api_client_async(&user, HttpApiClientConfig::default())?;
 
-        let url = get_tunnel_url().await?;
+        // TODO: make Tunnel struct responsible for getting its own port!
```
EverlastingBugstopper, line 48: want to make an issue or are you getting in this pr?

src/tail/session.rs
EverlastingBugstopper: lol

src/commands/tail/mod.rs
ashleymichal: i actually think i want to remove the copy and add the spinner #1161

src/main.rs
ashleymichal: honestly i did it because i expect that users are way less likely to set the metrics port and also because i want to deprecate that eventually in favor of a less hairy method for extracting the port

src/main.rs
ashleymichal: agreed.

src/tail/session.rs
ashleymichal: not happening in this PR, i'm going to do it as part of #1200

src/commands/tail/mod.rs
EverlastingBugstopper: beautiful

src/main.rs
EverlastingBugstopper: i feel like if they're setting one they'll set both - i'm not super hung up on this one but it feels strange for one to be `--port` when they are both ports

ashleymichal: again, wanna deprecate metrics. i could give it `--metrics-port` instead of just metrics but i'm pretty set on the primary being just `--port`

PR #1228, opened 2020-04-28
wip preview namespaces

EverlastingBugstopper:
need to figure out sites - its not being modified to use the preview namespace yet

### DIFF
--- a/src/commands/kv/namespace/list.rs
+++ b/src/commands/kv/namespace/list.rs
@@ -13,19 +13,24 @@ use crate::settings::toml::Target;
 const MAX_NAMESPACES_PER_PAGE: u32 = 100;
 const PAGE_NUMBER: u32 = 1;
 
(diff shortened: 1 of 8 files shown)


src/commands/kv/namespace/list.rs
```diff
@@ -13,19 +13,24 @@ use crate::settings::toml::Target;
 const MAX_NAMESPACES_PER_PAGE: u32 = 100;
 const PAGE_NUMBER: u32 = 1;
 
-pub fn list(target: &Target, user: &GlobalUser) -> Result<(), failure::Error> {
+pub fn print_list(target: &Target, user: &GlobalUser) -> Result<(), failure::Error> {
```
ashleymichal, line 16: i'd say just put  this logic in the caller (main.rs in this  case) rather than making it its own fn on this module. pulling the print out of the  original list fn is the right move though

PR #1231, opened 2020-04-29
add tests for find_open_port

ashleymichal:
(commented  inline but worth repeating here:)

These tests are extremely stateful; since what we are testing is how this function behaves
when requested ports are unavailable, and since our tests run concurrently, each test uses
unique ports to avoid collisions. There are two possible solutions to this problem:

1. require that these tests be run serially, and find a way to clean up bound ports
2. use dependency injection and write a substitute for the `TcpListener::bind` fn.

### DIFF
--- a/src/commands/tail/mod.rs
+++ b/src/commands/tail/mod.rs
@@ -51,3 +51,81 @@ fn find_open_port(requested: Option<u16>, default: u16) -> Result<u16, failure::
         Ok(socket.local_addr()?.port())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    // These tests are extremely stateful; since what we are testing is how this function behaves
+    // when requested ports are unavailable, and since our tests run concurrently, each test uses
+    // unique ports to avoid collisions. There are two possible solutions to this problem:
+    // 1. require that these tests be run serially, and find a way to clean up bound ports
+    // 2. use dependency injection and write a substitute for the TcpListener::bind fn.
+    use super::*;
+
+    #[test]
+    fn test_find_open_port_no_requested_default_available() {
+        let requested = None;
+        let default = 3000;
+        let port = find_open_port(requested, default).unwrap();
+
+        // returns default
+        assert_eq!(port, default);
+    }
+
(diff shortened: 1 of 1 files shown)


src/commands/tail/mod.rs
```diff
@@ -51,3 +51,79 @@ fn find_open_port(requested: Option<u16>, default: u16) -> Result<u16, failure::
         Ok(socket.local_addr()?.port())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    // These tests are extremely stateful; since what we are testing is how this function behaves
+    // when requested ports are unavailable, and since our tests run concurrently, each test uses
+    // unique ports to avoid collisions. There are two possible solutions to this problem:
+    // 1. require that these tests be run serially, and find a way to clean up bound ports
```
EverlastingBugstopper, line 60: cleaning up bound ports should happen automatically since i believe TcpListener::bind implements Drop

src/commands/tail/mod.rs
```diff
@@ -51,3 +51,79 @@ fn find_open_port(requested: Option<u16>, default: u16) -> Result<u16, failure::
...
+    // 2. use dependency injection and write a substitute for the TcpListener::bind fn.
+    use super::*;
+
+    #[test]
+    fn test_find_open_port_no_requested_default_available() {
+        let default = 3000;
+        let port = find_open_port(None, default).unwrap();
```
EverlastingBugstopper, line 67: can we name `None` above like we do for default, im not sure what the parameter is before hunting it down in the function definition

src/commands/tail/mod.rs
ashleymichal: it does; what i'm not certain of is how i can get these tests in particular to run serially and sandboxed from all other tests in the suite. nothing else currently binds ports, but y'know.

this is also just not a good unit testing practice.

src/commands/tail/mod.rs
ashleymichal: sure; it's also called out in the test fn name.

PR #1233, opened 2020-04-29
Refactor kv namespaces

EverlastingBugstopper:
This refactor starts to pull sites into the fold of kv namespaces

### DIFF
--- a/src/commands/kv/bucket/mod.rs
+++ b/src/commands/kv/bucket/mod.rs
@@ -21,7 +21,7 @@ use twox_hash::XxHash64;
 
 use cloudflare::endpoints::workerskv::write_bulk::KeyValuePair;
 
-use crate::settings::toml::Target;
+use crate::settings::toml::Site;
 
 pub const KEY_MAX_SIZE: usize = 512;
 // Oddly enough, metadata.len() returns a u64, not usize.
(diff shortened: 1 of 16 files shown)


src/commands/kv/bucket/mod.rs
```diff
@@ -21,23 +21,23 @@ use twox_hash::XxHash64;
...
+use crate::settings::toml::Site;
 
 pub const KEY_MAX_SIZE: usize = 512;
 // Oddly enough, metadata.len() returns a u64, not usize.
 pub const VALUE_MAX_SIZE: u64 = 10 * 1024 * 1024;
 
 // Returns the hashed key and value pair for all files in a directory.
 pub fn directory_keys_values(
-    target: &Target,
+    site: Option<Site>,
```
ashleymichal, line 32: running down this, we use this to find out whether to include or exclude files. i'd be inclined to pass those two parameters instead of an `Option<Site>`

src/settings/toml/target.rs
```diff
@@ -33,6 +80,10 @@ impl Target {
         self.kv_namespaces = Some(updated_namespaces);
     }
 
+    pub fn remove_all_kv_namespaces(&mut self) {
```
ashleymichal, line 83: this refactor is a great improvement overall, but this method sort of gives me pause. i don't have a better answer though so seems good for now.

EverlastingBugstopper: it's used for unauthenticated preview and nowhere else - agree it's a little weird

src/commands/kv/bucket/mod.rs
EverlastingBugstopper: if i try to do this, then `kv::bucket::sync` has an unwieldy number of parameters. going to leave this as is for another refactor

ashleymichal: it can be destructured in kv::bucket::sync before being passed  to this fn.

PR #1239, opened 2020-05-01
suggest `wrangler init` if wrangler.toml does not exist

### LINKED ISSUE #827: if no wrangler.toml found on build/preview/publish, give an error message to run wrangler init

ashleymichal:
closes #827


src/settings/toml/manifest.rs
```diff
@@ -46,6 +46,10 @@ pub struct Manifest {
 
 impl Manifest {
     pub fn new(config_path: &Path) -> Result<Self, failure::Error> {
+        failure::ensure!(
+            config_path.exists(),
+            "wrangler.toml not found; run `wrangler init` to create one."
```
EverlastingBugstopper, line 51: nit: add link to docs about how to configure

src/settings/toml/tests/mod.rs
```diff
@@ -147,5 +147,5 @@ fn base_fixture_path() -> PathBuf {
 }
 
 fn toml_fixture_path(fixture: &str) -> PathBuf {
-    base_fixture_path().join(fixture)
+    base_fixture_path().join(fixture).with_extension("toml")
```
EverlastingBugstopper, line 150: why is this change necessary?

ashleymichal: some magical thing in `config` crate was doing this for us when we went to use `merge`, but now that we first check `config_path.exists()` before we get to that point, the extension needs to be explicitly included in the path.

PR #1244, opened 2020-05-04
dont use fixture lock for testing

### LINKED ISSUE #1243: investigate failing ci
macos nightly and stable sometimes time out and it has not been reproduced locally and it's a massive time sink.

going to see if i can reproduce locally with [local github actions runner](https://github.com/nektos/act)

EverlastingBugstopper:
probably fixes #1243

changes: 

1) dont send `RUST_LOG=wrangler,info` since it doesn't provide any extra output anyway
2) cache entire `target` directory
3) since we create a unique temp directory for every fixture we do not need to run tests that use fixtures in parallel. this means that we won't have those intermittent CI errors, and it also means that our tests run slightly faster. _**IFF** my theory holds true and these builds run on each subsequent run_

what led me to believe this was the issue was that the tests appeared to be deadlocking, even when i limited cargo test to one thread (to run them consecutively). this must have meant that we were trying to acquire a lock twice. since we don't really use mutexes elsewhere in the codebase, it must have been the mutex used to run tests with fixtures consecutively. removing it _seems_ to work!

### DIFF
--- a/tests/build.rs
+++ b/tests/build.rs
@@ -1,6 +1,3 @@
-#[macro_use]
-extern crate lazy_static;
-
 pub mod fixture;
 
 use std::fs;
@@ -369,7 +366,6 @@ fn it_builds_with_webpack_name_output_warn() {
 }
 
 fn build_creates_assets(fixture: &Fixture, script_names: Vec<&str>) -> (String, String) {
-    let _lock = fixture.lock();
     let mut build = Command::cargo_bin(env!("CARGO_PKG_NAME")).unwrap();
     build.current_dir(fixture.get_path());
     build.arg("build");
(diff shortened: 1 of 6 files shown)


EverlastingBugstopper:
of course as soon as i say that it stops working again 🙄 back to drawing board

tests/build.rs
```diff
@@ -389,20 +389,23 @@ fn build_creates_assets(fixture: &Fixture, script_names: Vec<&str>) -> (String,
 
 fn build_fails_with(fixture: &Fixture, expected_message: &str) {
     let _lock = fixture.lock();
+    println!("lock {:?}", _lock);
```
ashleymichal, line 392: should these still be here?

EverlastingBugstopper: no

tests/preview.rs
```diff
@@ -248,17 +244,17 @@ fn preview_succeeds_with(fixture: &Fixture, env: Option<&str>, expected: &str) {
 }
 
 fn preview_succeeds(fixture: &Fixture) {
-    let _lock = fixture.lock();
     env::remove_var("CF_ACCOUNT_ID");
+    env::remove_var("CF_ZONE_ID");
```
ashleymichal, line 248: it'd  be nice to get the   mock env  integrated into the manifest. soon.

PR #1247, opened 2020-05-05
[tail] improve error messaging

### LINKED ISSUE #1177: better error handling / messaging

EverlastingBugstopper:
Fixes #1177

### DIFF
--- a/src/tail/mod.rs
+++ b/src/tail/mod.rs
@@ -20,11 +20,14 @@ use session::Session;
 use shutdown::ShutdownHandler;
 use tunnel::Tunnel;
 
+use console::style;
 use tokio;
 use tokio::runtime::Runtime as TokioRuntime;
+use which::which;
 
 use crate::settings::global_user::GlobalUser;
 use crate::settings::toml::Target;
+use crate::terminal::emoji;
 
 pub struct Tail;
 
@@ -35,6 +38,9 @@ impl Tail {
         tunnel_port: u16,
         metrics_port: u16,
     ) -> Result<(), failure::Error> {
+        is_cloudflared_installed()?;
+        print_startup_message(&target.name, tunnel_port, metrics_port);
+
         let mut runtime = TokioRuntime::new()?;
 
         runtime.block_on(async {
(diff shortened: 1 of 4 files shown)


src/tail/mod.rs
```diff
@@ -35,6 +38,9 @@ impl Tail {
         tunnel_port: u16,
         metrics_port: u16,
     ) -> Result<(), failure::Error> {
+        is_cloudflared_installed()?;
+        print_startup_message(&target.name, tunnel_port, metrics_port);
```
ashleymichal, line 42: why move this here? if anything perhaps we should modify the message, but i'd prefer to minimize the amount of printing being done in these fns.

src/tail/mod.rs
```diff
@@ -74,3 +80,28 @@ impl Tail {
...
     }
 }
+
+fn is_cloudflared_installed() -> Result<(), failure::Error> {
+    // this can be removed once we automatically install cloudflared
+    if which("cloudflared").is_err() {
+        let install_url = style("https://developers.cloudflare.com/argo-tunnel/downloads/")
+            .blue()
+            .bold();
+        failure::bail!("You must install cloudflared to use wrangler tail.\n\nInstallation instructions can be found here:\n{}", install_url);
```
ashleymichal, line 90: oh i like this. v. nice.

src/tail/mod.rs
EverlastingBugstopper: i moved it because i wanted the cloudflared error message to appear before we printed the "starting tunnel" message, and i moved it to a function because it's messy and doesn't need to be at the top level here

ashleymichal: hrm. i think it would be good to modify the message instead to say "starting tail session" and leave it. pulling to fn is good though. long term if we need more granular status updates we can use messaging to trigger them.

ashleymichal: withdrawn. multitasking today and i thought this was tunnel.rs.

PR #1257, opened 2020-05-07
use unique cache for fixture tests

EverlastingBugstopper:
ok this is the only other thing that uses lazy_static which has been known to cause deadlocks.


EverlastingBugstopper:
if anybody is out there reading this, please help

Issue #1219, opened 2020-04-21
Use `webpack --watch` for live reload


EverlastingBugstopper:
Instead of using `notify` to tell `webpack` to build a script from scratch, we should instead watch the output of `webpack`, and shell out to `webpack --watch`. Then, when a new build is complete, wrangler should reupload the Worker. This can already be achieved by an end user using the `javascript` type and watching with `webpack --watch`

EverlastingBugstopper:
i'm pretty sure this already happens, we just weren't taking advantage of it. closed by #1269 (i think)

PR #1269, opened 2020-05-11
removes redundant builds when watching

EverlastingBugstopper:
turns out wranglerjs already implements #1219 (correct me if im wrong @xtuc - we call `compiler.watch` when we pass `--watch=1` to `wranglerjs`)

the problem was that we would start a wranglerjs process with `--watch=1` at the beginning, but then we'd _also_ run a regular blocking wrangler js build after code was changed. this should speed up live reload by at least 2x...? probably? haven't actually benchmarked but this change should make a noticeable difference.

cc @PierBover this is relevant to your interests 😄


xtuc:
Sorry I can't remember. DIdn't you implemented https://github.com/cloudflare/wrangler/blob/master/wranglerjs/index.js#L158-L159?

EverlastingBugstopper:
nah - must have been @xortive. not sure where to even find the docs on using webpack as a library

PierBover:
Thanks for the mention!

PR #1276, opened 2020-05-12
[dev] reconnect socket after bad protocol message

### LINKED ISSUE #1241: [dev] Rust panics after inactivity "Connection reset without closing handshake"
### 🐛 Bug Report

I'm creating this bug as requested by @EverlastingBugstopper here: https://github.com/cloudflare/wrangler/issues/1047#issuecomment-623505738

After the computer goes to sleep or after the worker hasn't been used for some time (a couple of minutes) Rust panics and shows this error message:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Protocol("Connection reset without closing handshake")', src/commands/dev/socket.rs:36:27
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

#### Environment

* operating system: macOS 10.13.6
* output of `node -v`: `v12.16.1`
* output of `wrangler -V`: ` ✨  wrangler 1.9.0-rc.0`

#### Steps to reproduce

Open a worker and put the computer to sleep or wait a couple of minutes without using the worker.

EverlastingBugstopper:
fixes #1241

after a while, the preview service disconnects the websocket, even if you're being a good citizen and sending keep alive heartbeat messages every now and then. for shame! after merging this if your dev session gets disconnected from the websocket we'll issue a reconnect. :) :) :)


src/commands/dev/socket.rs
```diff
@@ -43,12 +43,28 @@ pub async fn listen(session_id: String) -> Result<(), failure::Error> {
...
+                            failure::format_err!("this event could not be parsed:\n{}", e)
+                        });
+                    if let Ok(protocol::Runtime::Event(event)) = parsed_message {
+                        println!("{}", event);
+                    }
+                }
+                Err(error) => match error {
+                    tokio_tungstenite::tungstenite::Error::Protocol(msg) => {
+                        if msg == "Connection reset without closing handshake" {
+                            println!("please restart the connection");
```
EverlastingBugstopper, line 61: not sure what to do here... if it's reached this point, there's no going back because the connection has been reset (and i believe it's being reset by cloudflare, not us)

i want to start the `listen(session_id)` function over from the beginning at this point and maybe print a little "reconnecting" message with a spinner or something but i have _no_ idea how.

src/commands/dev/socket.rs
```diff
@@ -1,63 +1,96 @@
...
-        })
-    };
 
-    // run the heartbeat and message printer in parallel
-    tokio::select! {
-        _ = keep_alive_to_ws => { Ok(()) }
-        _ = print_ws_messages => { Ok(()) }
+// the reason this needs to return a `BoxFuture` is so that we can call it recursively
+// if something goes wrong with the websocket connection
+pub fn listen(session_id: String) -> BoxFuture<'static, Result<(), failure::Error>> {
```
EverlastingBugstopper, line 27: `BoxFuture` is fun and interesting and lets me run this fn recursively (i want to restart the function every time something errors)

src/commands/dev/socket.rs
```diff
@@ -1,63 +1,96 @@
...
+
+        // if left unattended, the preview service will kill the socket
+        // that emits console messages
+        // send a keep alive message every so often in the background
+        let (keep_alive_tx, keep_alive_rx) = mpsc::unbounded_channel();
+        tokio::spawn(keep_alive(keep_alive_tx));
+        let keep_alive_to_ws = keep_alive_rx
+            .map(Ok)
+            .forward(write)
+            .map_err(|e| failure::format_err!("{:?}", e));
```
EverlastingBugstopper, line 56: mapping error to `failure` so everything we pass to `try_join` has the same return type bc that's a thing apparently

src/commands/dev/socket.rs
```diff
@@ -1,63 +1,96 @@
...
+        }
+    }
+    .boxed()
+}
+
+async fn print_ws_messages(
+    read: Arc<Mutex<SplitStream<WebSocketStream<Stream<TcpStream, TlsStream<TcpStream>>>>>>,
+) -> Result<(), failure::Error> {
+    let mut read = read.lock().unwrap();
+    while let Some(message) = read.next().await {
```
EverlastingBugstopper, line 77: `read.next()` requires mutability, which is why i have passed an `Arc<Mutex` so that we can grab a mutable reference

EverlastingBugstopper:
getting a pretty opaque compiler error that i'm not sure how to resolve.. 

```
error: future cannot be sent between threads safely
   --> src/commands/dev/socket.rs:59:23
    |
59  |         let printer = tokio::spawn(print_ws_messages(Arc::clone(&passable_read)))
    |                       ^^^^^^^^^^^^ future returned by `print_ws_messages` is not `Send`
    | 
   ::: /Users/averyharnish/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-0.2.20/src/task/spawn.rs:127:21
    |
127 |         T: Future + Send + 'static,
    |                     ---- required by this bound in `tokio::task::spawn::spawn`
    |
    = help: within `impl core::future::future::Future`, the trait `std::marker::Send` is not implemented for `std::sync::MutexGuard<'_, futures_util::stream::stream::split::SplitStream<tokio_tungstenite::WebSocketStream<tokio_tungstenite::stream::Stream<tokio::net::tcp::stream::TcpStream, tokio_tls::TlsStream<tokio::net::tcp::stream::TcpStream>>>>>`
note: future is not `Send` as this value is used across an await
   --> src/commands/dev/socket.rs:77:31
    |
76  |     let mut read = read.lock().unwrap();
    |         -------- has type `std::sync::MutexGuard<'_, futures_util::stream::stream::split::SplitStream<tokio_tungstenite::WebSocketStream<tokio_tungstenite::stream::Stream<tokio::net::tcp::stream::TcpStream, tokio_tls::TlsStream<tokio::net::tcp::stream::TcpStream>>>>>` which is not `Send`
77  |     while let Some(message) = read.next().await {
    |                               ^^^^^^^^^^^^^^^^^ await occurs here, with `mut read` maybe used later
...
94  | }
    | - `mut read` is later dropped here
```

i want an async fn that i can `spawn` and then `try_join` 

and if an error occurs in either of the futures i've spawned (keep_alive messages and printing incoming socket messages) i want to kill the futures i've spawned and start my `listen` fn over from the beginning

src/commands/dev/socket.rs
```diff
@@ -1,63 +1,96 @@
...
-        read.for_each(|message| async {
-            let message = message.unwrap().into_text().unwrap();
-            log::info!("{}", message);
-            let message: Result<protocol::Runtime, failure::Error> = serde_json::from_str(&message)
-                .map_err(|e| failure::format_err!("this event could not be parsed:\n{}", e));
-            if let Ok(protocol::Runtime::Event(event)) = message {
-                println!("{}", event);
-            }
-        })
-    };
```
EverlastingBugstopper, line 54: this is what i had before, but it doesn't quite work for me since it can't handle errors properly, i can't just pass it to `try_join`

src/commands/dev/socket.rs
```diff
@@ -16,51 +20,78 @@ const KEEP_ALIVE_INTERVAL: u64 = 10;
...
-    // parse every incoming message and print them
-    let print_ws_messages = {
-        read.for_each(|message| async {
-            let message = message.unwrap().into_text().unwrap();
-            log::info!("{}", message);
-            let message: Result<protocol::Runtime, failure::Error> = serde_json::from_str(&message)
-                .map_err(|e| failure::format_err!("this event could not be parsed:\n{}", e));
-            if let Ok(protocol::Runtime::Event(event)) = message {
-                println!("{}", event);
+    async move {
```
ObsidianMinor, line 23: `async move { }.await` is redundant in this context.

src/commands/dev/socket.rs
```diff
@@ -16,51 +20,78 @@ const KEEP_ALIVE_INTERVAL: u64 = 10;
...
+
+            // every 10 seconds, send a keep alive message on the channel
+            let heartbeat = keep_alive(keep_alive_tx);
+
+            // when the keep alive channel receives a message from the
+            // heartbeat future, write it to the websocket
+            let keep_alive_to_ws = keep_alive_rx
+                .map(Ok)
+                .forward(write)
+                .map_err(|e| failure::format_err!("{:?}", e));
```
ObsidianMinor, line 57: This can also be mapped like so since it returns an error that implements `std::error::Error`:
```suggestion
                .map_err(Into::into);
```

src/commands/dev/socket.rs
```diff
@@ -16,51 +20,78 @@ const KEEP_ALIVE_INTERVAL: u64 = 10;
...
+
+            // parse all incoming messages and print them to stdout
+            let printer = print_ws_messages(read).map_err(|e| failure::format_err!("{:?}", e));
```
ObsidianMinor, line 60: `print_ws_messages` already returns a `failure::Error`, so this map is also redundant
```suggestion
            let printer = print_ws_messages(read);
```

src/commands/dev/socket.rs
EverlastingBugstopper: oops - yes that is from when i was using `BoxFuture`, good catch

src/commands/dev/socket.rs
EverlastingBugstopper: nice catch

EverlastingBugstopper:
@ObsidianMinor thanks so much for the review 😄

PR #1281, opened 2020-05-12
wip add cors options

EverlastingBugstopper:
this doesn't work yet but im thinking of making `--cors` take an allowed origin and setting `Access-Control-Allow-Origin` depending on that arg.

`wrangler dev` wouldn't do anything different
`wrangler dev --cors localhost:3000` would allow requests from `localhost:3000`
`wrangler dev --cors localhost:3000 --cors localhost:3001` would allow requests from `localhost:3000` and `localhost:3001`.

cc @tibotiber @PierBover @dhaynespls

### DIFF
--- a/src/commands/dev/gcs/server.rs
+++ b/src/commands/dev/gcs/server.rs
@@ -8,7 +8,7 @@ use chrono::prelude::*;
 
 use hyper::client::{HttpConnector, ResponseFuture};
 use hyper::header::{HeaderName, HeaderValue};
-use hyper::http::uri::InvalidUri;
+use hyper::http::{uri::InvalidUri, Method};
 use hyper::service::{make_service_fn, service_fn};
 use hyper::{Body, Client as HyperClient, Request, Response, Server, Uri};
 
@@ -48,6 +48,23 @@ pub(super) async fn serve(
                 // split the request into parts so we can read
                 // what it contains and display in logs
                 let (parts, body) = req.into_parts();
+
+                if parts.method == Method::OPTIONS && server_config.allowed_origins.len() > 0 {
+                    return async move {
+                        let response_builder = Response::builder()
+                            .header("Access-Control-Allow-Credentials", "true")
+                            .header(
+                                "Access-Control-Allow-Methods",
+                                "GET, HEAD, POST, PUT, DELETE, OPTIONS",
+                            )
+                            .header("Access-Control-Allow-Headers", "Content-Type");
+                        for allowed_origin in server_config.allowed_origins {
+                            response_builder
+                                .header("Access-Control-Allow-Origin", allowed_origin.to_string());
+                        }
+                        Ok::<_, failure::Error>(response_builder.body(()))
+                    };
+                }
 
                 let req_method = parts.method.to_string();
 
--- a/src/commands/dev/server_config/mod.rs
+++ b/src/commands/dev/server_config/mod.rs
@@ -4,10 +4,13 @@ mod listening_address;
 use host::Host;
 use listening_address::ListeningAddress;
 
+use http::Uri;
+
 #[derive(Debug, Clone)]
 pub struct ServerConfig {
     pub host: Host,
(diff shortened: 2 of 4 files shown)


src/commands/dev/gcs/server.rs
```diff
@@ -49,6 +49,24 @@ pub(super) async fn serve(
...
+                            )
+                            .header("Access-Control-Allow-Origin", "http://localhost:3000")
+                            .header("Access-Control-Allow-Headers", "Content-Type");
+                        for allowed_origin in server_config.allowed_origins {
+                            response_builder
+                                .header("Access-Control-Allow-Origin", allowed_origin.to_string());
+                        }
+                        Ok::<_, failure::Error>(response_builder.body(()))
+                    };
+                }
```
EverlastingBugstopper, line 68: this part super doesnt compile right now but this is sorta what im thinking for handling OPTIONS requests when `--cors` args are passed

src/commands/dev/server_config/mod.rs
```diff
@@ -4,28 +4,36 @@ mod listening_address;
...
     pub listening_address: ListeningAddress,
+    pub allowed_origins: Vec<Uri>,
 }
 
 impl ServerConfig {
     pub fn new(
         host: Option<&str>,
         ip: Option<&str>,
         port: Option<&str>,
+        allowed_origins: &[&str],
```
EverlastingBugstopper, line 21: starting to wonder if we want to have some sort of definition in `wrangler.toml` for stuff like this, seems like stuff you'd want to configure and share across a team

dhaynespls: having a [dev] block in the wrangler toml makes sense for stuff like this

PierBover:
Looking good!

FYI you'd need to return these CORS headers on all responses (200, 400, 500), not only on OPTIONS requests:

```
response.headers.set('Access-Control-Allow-Origin', 'http://localhost:3000');
response.headers.set('Access-Control-Allow-Credentials', 'true');
```

I'm not an expert on CORS but maybe another thing that could be useful is to allow more headers to be added instead of hardcoding `'Content-Type'`.

I'm referring to this: `'Access-Control-Allow-Headers': 'Content-Type'`

What if someone wants to use some special authorization header?

dhaynespls:
@PierBover tbh it can probably just be * -- We're already saying that if you set this argument then we're going to allow requests in for that origin you tell us, so I think a natural assumption is that any headers can also arrive.

PierBover:
> @PierBover tbh it can probably just be `'Access-Control-Allow-Headers': '*'` -- We're already saying that if you set this argument then we're going to allow requests in for that origin you tell us, so I think a natural assumption is that any headers can also arrive.

It cannot be `*` if you are using credentials.

> For requests without credentials, the literal value "*" can be specified, as a wildcard; the value tells browsers to allow requesting code from any origin to access the resource. Attempting to use the wildcard with credentials will result in an error.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin

PierBover:
@dhaynespls that would not work if there are cookies.

> The value "*" only counts as a special wildcard value for requests without credentials

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers

tibotiber:
I can't comment on rust code itself but looks good overall. Definitely in line with expected functionality. 

For `Access-Control-Allow-Headers`, maybe we could use `*` as default since it would support all headers out of the box for non authorized request, and add a note that authorized requests need headers manually set. Since this introduces more complex settings, the `[dev]` block in `wrangler.toml` starts to make sense. I wonder though, how are CORS middlewares typically managing this? I've used many but have never had to set individual headers 🤔.

Is there any existing CORS middleware for the server you are using btw?

EverlastingBugstopper:
ill get to this but i dont think it will be soon :(

PR #1288, opened 2020-05-15
Adds Developing Wrangler section to Contribution Guide

EverlastingBugstopper:
Yay external contributors! Fixes #1288

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,9 +5,11 @@ and the opportunity to contribute to their developer tools. Wrangler is also a p
 delivered by Cloudflare, so it's important to clarify how we think about issue triage and
 contributions.
 
+If you want to learn about how to get started with Wrangler, [click here](#Developing-Wrangler)
+
 ## People
 
-Wrangler is maintained by @ashleygwilliams, and her team, Workers Developer Experience.
+Wrangler is owned by the [Cloudflare Workers](https://workers.cloudflare.com) Team and maintained by community members like you! The core maintainers are @EverlastingBugstopper and @ashleymichal, and everything that gets merged must be approved by at least one team member.
 
 ## Primary Issue Triage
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -65,3 +64,30 @@ should be triaged immediately upon open by the PR author.
...
+To build Wrangler from source, fork the repo, clone it, and `cd wrangler`.
+
+There are many ways to build and execute a development version of Wrangler:
+
+`cargo build` will produce a local binary at `./target/debug/wrangler` that you can execute.
+`cargo install --debug --path .` will replace any globally installed wrangler with the one you've just built from source.
+
+You can read more about cargo [here](https://doc.rust-lang.org/cargo/), just find something that works for you.
+
+### Command-line argument parsing
```
ashleymichal, line 85: can we also call out cloudflare-rs? i feel like noting client might be good.

CONTRIBUTING.md
```diff
@@ -65,3 +64,30 @@ should be triaged immediately upon open by the PR author.
...
+
+The primary framework we use for developing Wrangler is called [clap](https://clap.rs), which provides fast and structured argument parsing. This is how different features are exposed to users, and most of that logic lives in [main.rs](./src/main.rs).
+
+Each of the commands supported by Wrangler have entrypoints in [./src/commands](./src/commands). It's useful to understand the [module system](https://doc.rust-lang.org/rust-by-example/mod.html) if you will be adding new commands or need to refactor/organize imports.
+
```
ashleymichal, line 90: it is also nice to call out pre-push steps, like test, fmt, and clippy, and the fact that if CI fails your PR may not be reviewed (if it is a legit bug). i also wouldn't mind calling out that additions to the readme/testing suite should be considered part of Definition of Done, but maybe we need to hash that out and add it later.

CONTRIBUTING.md
```diff
@@ -5,24 +5,27 @@ and the opportunity to contribute to their developer tools. Wrangler is also a p
 delivered by Cloudflare, so it's important to clarify how we think about issue triage and
 contributions.
 
+If you want to learn about how to get started with Wrangler, [click here](#Developing-Wrangler)
+
 ## People
 
-Wrangler is maintained by @ashleygwilliams, and her team, Workers Developer Experience.
+Wrangler is maintained by the Workers Developer Productivity team. The core maintainers are @EverlastingBugstopper and @ashleymichal, and everything that gets merged must be approved by at least one team member.
```
ispivey, line 12: I'd suggest just saying the "Workers" team. It has the benefit of being both true and more inclusive; I also like how it helps abstract the most-relevant concept of "core maintainers" away from Cloudflare's organizational structure.

ispivey:
This is great, and an unequivocal improvement 🤗 

I'll leave it to you and @ashleymichal to agree on what's good enough to merge, but consider this a 👍  from me.

EverlastingBugstopper:
Comments should be addressed!

CONTRIBUTING.md
xtuc: I'm not sure it's more inclusive. What about just maintained by Cloudflare?

EverlastingBugstopper: could say `owned by the Cloudflare Workers team and maintained by anybody who wants to contribute`

PR #1289, opened 2020-05-19
removes --release from publish help

EverlastingBugstopper:
we deprecated this a while ago, we can still make it behave the same (alias of wrangler publish) but remove it from the help text :)

before

```console
$ wrangler publish --help
wrangler-publish 
🆙  Publish your worker to the orange cloud

USAGE:
    wrangler publish [FLAGS] [OPTIONS]

FLAGS:
    -h, --help       Prints help information
        --release    [deprecated] alias of wrangler publish
        --verbose    toggle verbose output

OPTIONS:
    -e, --env <env>    environments to publish to
```

after

```console
$ wrangler publish --help
wrangler-publish 
🆙  Publish your worker to the orange cloud

USAGE:
    wrangler publish [FLAGS] [OPTIONS]

FLAGS:
    -h, --help       Prints help information
        --verbose    toggle verbose output

OPTIONS:
    -e, --env <env>    environments to publish to
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -470,6 +470,7 @@ fn run() -> Result<(), failure::Error> {
                 )
                 .arg(
                     Arg::with_name("release")
+                        .hidden(true)
                         .long("release")
                         .takes_value(false)
                         .help("[deprecated] alias of wrangler publish")


LICENSE-MIT
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2018 Ashley Williams <ashley666ashley@gmail.com>
+Copyright (c) 2020 Cloudflare, Inc. <wrangler@cloudflare.com>
```
ashleymichal, line 1: now this PR does two things, idk how  licensing things  work

src/main.rs
```diff
@@ -470,6 +470,7 @@ fn run() -> Result<(), failure::Error> {
                 )
                 .arg(
                     Arg::with_name("release")
```
ashleymichal, line 472: if we deprecated it, should it just go away?

LICENSE-MIT
EverlastingBugstopper: uhh woops

src/main.rs
EverlastingBugstopper: eh - doesnt really hurt to keep it around so the command doesnt fail in case folks have it in CI but if you want we can remove

ashleymichal: if  it's going into a  point release, we should probably leave it. would be a shitty reason for your ci  to fail. maybe  it doesn't get pulled until 2.0? idk.

EverlastingBugstopper: yeah, makes sense

PR #1291, opened 2020-05-19
[dev] connect to edge websocket

EverlastingBugstopper:
this should Just Work once API change goes through to move websocket auth to url from header


src/commands/dev/edge/mod.rs
```diff
@@ -20,9 +20,22 @@ pub fn dev(
...
+        &mut target,
+        None,
+        &deploy_config,
+        &user,
+        init.preview_token.clone(),
+    )?;
     let mut runtime = TokioRuntime::new()?;
-    runtime.block_on(server)
+    runtime.block_on(async {
+        let devtools_listener = tokio::spawn(socket::listen(init.websocket_url));
```
jwheels, line 32: does this take care of adding the token to the URL as a GET param? (the websocket URL is just the base url, without the token inside, although, I could change that too...)

EverlastingBugstopper: mmmm - nope, will add

EverlastingBugstopper: updated 😄

PR #1294, opened 2020-05-20
Adds basic Dependabot config

ispivey:
The app appears to be installed on the repo, so maybe we just need a config file? 😬 

https://dependabot.com/docs/config-file/


ispivey:
I'm not sure how to test this, other than merging. And, as I re-read the docs, maybe we should try the "live" frequency first, so we can tell if it's working right away?

I define "working" as "opens PRs to update our dependencies".

PR #1306, opened 2020-05-20
fixes more clippy warnings

EverlastingBugstopper:
gotta catch em all!!!

### DIFF
--- a/src/terminal/interactive.rs
+++ b/src/terminal/interactive.rs
@@ -12,12 +12,14 @@ pub fn get_user_input_multi_line(prompt_string: &str) -> String {
     println!("{}", prompt_string);
     let mut input = String::new();
     // are we reading from user input?
-    if atty::is(Stream::Stdin) {
-        input = read!("{}\n");
+    let mut input = if atty::is(Stream::Stdin) {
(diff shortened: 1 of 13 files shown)


src/terminal/interactive.rs
```diff
@@ -12,12 +12,14 @@ pub fn get_user_input_multi_line(prompt_string: &str) -> String {
...
-        input = read!("{}\n");
+    let mut input = if atty::is(Stream::Stdin) {
+        read!("{}\n")
     } else {
         // or is this data from a pipe? (support newlines)
-        drop(io::stdin().read_to_string(&mut input));
-    }
+        let tmp = String::new();
+        let _ = io::stdin().read_to_string(&mut input);
+        tmp
```
dmcgowan, line 21: Is this overwriting input with an empty string? I was trying to trace down an empty input when using pipe in 1.9.1, reverting back to 1.9.0 worked for me.

EverlastingBugstopper: whhooooppss - it looks like that should be `tmp` not `input`. probably gonna need another point release for this one. thanks @dmcgowan

PR #1315, opened 2020-05-21
updates dependencies

EverlastingBugstopper:
doing this all in one go so we only get a trickle of dependabot PRs

there were two breaking changes made here, the only one i was kinda worried about was the `percent_encoding` crate, but i tested things manually and migration seemed to work ok.

@ashleymichal mind double checking on that? it's related to kv stuff


src/commands/kv/mod.rs
```diff
@@ -119,7 +119,7 @@ pub fn get_namespace_id(target: &Target, binding: &str) -> Result<String, failur
 }
 
 fn url_encode_key(key: &str) -> String {
-    percent_encode(key.as_bytes(), PATH_SEGMENT_ENCODE_SET).to_string()
+    utf8_percent_encode(key, CONTROLS).to_string()
```
ashleymichal, line 122: this is probably indeed what we actually want; my only hesitation is that both of these functions existed before... how did you get from one to the other? i notice `as_bytes` is now gone so that is relevant i guess?

ashleymichal: basically, how would we test the relevant edge cases for this? if it was part of sites i'd almost say screw it since we control the asset manifest, but this is used for `kv:key put` and `kv:key get`

EverlastingBugstopper: not sure how to test  edge cases, but they completely removed `PATH_SEGMENT_ENCODE_SET` from the API, and I've tested kv:key put and kv:key get individually and they both work just fine

EverlastingBugstopper: i can revert to old version if we want to address this singular depedency later but i think it should work ok

ashleymichal: the edge cases would be against specific values, but it's.... probably okay?

EverlastingBugstopper: ok - from the docs it says 

> Use URL-encoding to use special characters (e.g. :, !, %) in the key name.

i just tested w/keys that include those characters and everything works as expected

ashleymichal: like you get the same results when you `put` with the old logic and `get` with the new, and vice versa?

EverlastingBugstopper: i tested that kv:key put and kv:key get both work when using a key with those special characters

PR #1316, opened 2020-05-22
Fix secret pipe input

dmcgowan:
Fixes bug where pipe input is always empty

This is my first PR in rust, so I wasn't quite sure if mut should be added to tmp or removed from `read_to_string` here. This fixed the issue I was seeing in 1.9.1 though


ashleymichal:
is there a relevant issue that should be linked here?

EverlastingBugstopper:
Fixes #1322

PR #1324, opened 2020-05-26
Runs clippy in CI + enforce all formatting

EverlastingBugstopper:


.github/workflows/lint.yml
```diff
@@ -14,6 +14,10 @@ jobs:
           rustup update stable
           rustup default stable
           rustup component add rustfmt
+          rustup component add clippy
 
       - name: Check Formatting
-        run: cargo fmt -- --check
+        run: cargo fmt --all -- --check
```
EverlastingBugstopper, line 20: `--all` makes sure that both the binaries and the libraries are formatted

PR #1325, opened 2020-05-26
enforces no compiler warnings in CI

EverlastingBugstopper:


src/main.rs
```diff
@@ -20,6 +21,10 @@ use wrangler::settings::global_user::GlobalUser;
 use wrangler::settings::toml::TargetType;
 use wrangler::terminal::{emoji, interactive, message, styles};
 
+fn unused_code() {
+    println!("THIS IS NEVER USED EITHER");
+}
```
ashleymichal, line 26: just testing it out?

EverlastingBugstopper: yeah - want to make sure it fails in CI. it failed locally but want to double check

src/main.rs
```diff
@@ -20,6 +21,7 @@ use wrangler::settings::global_user::GlobalUser;
 use wrangler::settings::toml::TargetType;
 use wrangler::terminal::{emoji, interactive, message, styles};
 
+#[allow(clippy::cognitive_complexity)]
```
ashleymichal, line 24: 😭

PR #1326, opened 2020-05-26
Made version check only happen once a day

jspspike:


EverlastingBugstopper:
this is perfect. congrats on your first PR of many!

src/version/version.rs
```diff
@@ -44,42 +45,55 @@ fn get_installed_version() -> Result<Version, failure::Error> {
...
-        None => {
-            let latest_version = get_latest_version_from_api(installed_version)?;
-            let updated_file_contents = toml::to_string(&LastCheckedVersion {
-                latest_version: latest_version.to_string(),
-                last_checked: current_time,
-            })?;
-            fs::write(&version_file, updated_file_contents)?;
-            Ok(latest_version)
+
+    Ok((WranglerVersion { current, latest }, checked))
```
ashleymichal, line 69: this tuple return is awkward. will review in the morning and suggest a different tactic.

ashleymichal, line 69: this tuple return is awkward. will review in the morning and suggest a different tactic.

jspspike: I can add it as a field in the `WranglerVersion` struct and add comments to make it more clear

ashleymichal: sgtm, i'll have a look then 😄

src/version/version.rs
```diff
@@ -45,41 +49,60 @@ fn get_installed_version() -> Result<Version, failure::Error> {
...
+        latest,
+        checked,
+    })
+}
+
+/// Reads version out of version file, is `None` if file does not exist
+fn get_version_disk(version_file: &PathBuf) -> Result<Option<LastCheckedVersion>, failure::Error> {
+    let local_version = match fs::read_to_string(&version_file) {
+        Ok(contents) => {
+            let last_checked_version = LastCheckedVersion::from_str(&contents)?;
```
ashleymichal, line 84: I think this fn should return `Option<LastCheckedVersion>` rather than `Result<Option<LastCheckedVersion>, failure::Error>`. I know this was the function signature in the first place, but I can't see a good way for a user to recover from this error; if this goes wrong, it likely means there is an issue with the file format  (we were able to read the file, but its contents were unexpected), and it seems like the best course of action is to return `None` so that the file is rewritten.

Thoughts? cc @EverlastingBugstopper

src/version/version.rs
```diff
@@ -116,7 +139,10 @@ pub fn background_check_for_updates() -> mpsc::Receiver<Version> {
 
     let _detached_thread = thread::spawn(move || match check_wrangler_versions() {
         Ok(wrangler_versions) => {
-            if wrangler_versions.current != wrangler_versions.latest {
+            // If the wrangler version has not been checked within the last day and the versions
+            // are different, print out an update message
+            if !wrangler_versions.checked && (wrangler_versions.current != wrangler_versions.latest)
```
ashleymichal, line 144: nit: this might be easier to grok as a method `outdated(&self)` on the `WranglerVersion` struct.

src/version/version.rs
EverlastingBugstopper: yeah that makes sense - generally if something goes wrong here we should probably just overwrite the file.

src/version/version.rs
```diff
@@ -116,7 +142,11 @@ pub fn background_check_for_updates() -> mpsc::Receiver<Version> {
 
     let _detached_thread = thread::spawn(move || match check_wrangler_versions() {
         Ok(wrangler_versions) => {
-            if wrangler_versions.current != wrangler_versions.latest {
+            // If the wrangler version has not been checked within the last day and the versions
+            // are different, print out an update message
+            if !wrangler_versions.is_outdated()
+                && (wrangler_versions.current != wrangler_versions.latest)
```
ashleymichal, line 148: sorry, i  mean this whole big boolean:

``` rust
fn is_outdated(&self) -> bool {
    !self.checked && (self.current != self.latest)
}

# and then call
if wrangler_versions.is_outdated() {
    ...
}
```

EverlastingBugstopper: yeah this should make it look pretty clean

Issue #1157, opened 2020-03-20
Provide better error message when API token doesn't have permissions


lukin0110:
When I'm using the `apiToken` the Github Actions build fails with:
```
Error: Your user configuration is invalid, please run wrangler config and enter a new set of credentials.
```

This doesn't work for my repo:
```yaml
name: Deploy

on:
  push:
    branches:
      - master

jobs:
  deploy:
    runs-on: ubuntu-latest
    name: Deploy
    steps:
      - uses: actions/checkout@master
      - name: Publish
        uses: cloudflare/wrangler-action@1.1.0
        with:
          apiToken: ${{ secrets.CF_API_TOKEN }}
```

I switched to using the **Global Key**. Cfr: https://github.com/MakeInBelgium/naaiactie/

EverlastingBugstopper:
This is probably worth refactoring to use the logic in `wrangler whoami` to let folks know exactly which permissions they're missing

jspspike:
I'll take a look at this one

jspspike:
Looking at what the api response is now the messages seem pretty clear on what permissions are missing or if your API token doesn't have access. On request of @EverlastingBugstopper I'll make a PR to clean up the messages so they look like so:

`Error: ApiErrors ["user is missing required permission \"#worker:edit\""]`
`Error: ApiErrors ["Authentication error"]`

cyberfox1:
Still hasn't been resolved, still wasting peoples time.

kuberlog:
I login, I then try to publish and get `Error: Your user configuration is invalid, please run wrangler login or wrangler config and enter a new set of credentials.`

I've tried using both tokens and API-key. Nothing works

xtuc:
@kuberlog can you try to reconfigre zone_id/account_id in your wrangler.toml using the information in the dashboard?

kuberlog:
@xtuc it was an account_id problem actually. The account_id referenced my colleague's account, which I obviously was uncredentialed to use. Thanks for the suggestion. Sorry for causing you to reopen this issue because of my user error.

kuberlog:
Perhaps an error message that said I am uncredentialed for that particular account ID would be useful though.

xtuc:
I reopened the issue because i believe wrangler could be more helpful. I also ran into the same issue this morning.

`wrangler whoiam` already show information about the account, it could be used to check if the user has access to the account or zone before publishing.

PR #1329, opened 2020-05-27
Clean up apierrors and only include messages

jspspike:
Makes API errors a little cleaner. Looked like this before
`Error: [ApiError { code: 10037, message: "user is missing required permission \"#worker:edit\"", other: {} }]`

Now it looks like
`Error:  Code 10037: user is missing required permission "#worker:edit"
 Edit your API Token to have correct permissions, or use the 'Edit Cloudflare Workers' API Token template.`

### DIFF
--- a/src/commands/kv/namespace/site.rs
+++ b/src/commands/kv/namespace/site.rs
@@ -41,7 +41,12 @@ pub fn site(
 
                     get_id_from_namespace_list(&client, target, &title)
                 } else {
-                    failure::bail!("{:?}", api_errors.errors)
+                    let error_messages: Vec<String> = api_errors
+                        .errors
+                        .iter()
+                        .map(|e| e.message.clone())
+                        .collect();
+                    failure::bail!("ApiError {:?}", error_messages);
                 }
             }
             ApiFailure::Invalid(reqwest_err) => failure::bail!("Error: {}", reqwest_err),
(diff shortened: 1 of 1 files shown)


jspspike:
Addresses #1157

src/commands/kv/namespace/site.rs
```diff
@@ -41,7 +41,12 @@ pub fn site(
 
                     get_id_from_namespace_list(&client, target, &title)
                 } else {
-                    failure::bail!("{:?}", api_errors.errors)
+                    let error_messages: Vec<String> = api_errors
+                        .errors
+                        .iter()
+                        .map(|e| e.message.clone())
+                        .collect();
+                    failure::bail!("ApiError {:?}", error_messages);
```
EverlastingBugstopper, line 49: oh actually - it looks like we already have a convenience function for this that should make it easier + uniform with the way we display errors.

It is `crate::http::format_error`, and it takes an `ApiFailure`, and an optional function to handle individual error codes.

So you could write something real quick to include the suggestions for codes `10026` and `10014` and pass it to that function. You can see an example of this in `crate::src::commands::route::delete`, at the end there is a `match result` and it calls `http::format_error`

This should allow us to display the error without making it look like an array

EverlastingBugstopper: (if i had thought of this before i would have pointed you in that direction!)

EverlastingBugstopper:
added a `changelog - feature` label to this one

PR #1332, opened 2020-05-27
Alewis/commands refactor kv

ashleymichal:
this looks  like a nightmare, and that's a little  bit because it is.  but this  gets most of the separation of concerns between kv  implementation and the commands, as  well as giving site specific logic  its own home.


src/kv/bulk.rs
```diff
@@ -0,0 +1,67 @@
+extern crate base64;
```
EverlastingBugstopper, line 1: what is the point of `extern crate`? that was deprecated in rust 2018

src/kv/bulk.rs
```diff
@@ -0,0 +1,67 @@
...
+        );
+    }
+
+    match client.request(&WriteBulk {
+        account_identifier: &target.account_id,
+        namespace_identifier: namespace_id,
+        bulk_key_value_pairs: pairs.to_owned(),
+    }) {
+        Ok(_) => Ok(()),
+        Err(e) => failure::bail!("{}", kv::format_error(e)),
```
EverlastingBugstopper, line 36: in another section, we imported this and just called `format_error` - lets make em consistent

src/kv/bulk.rs
ashleymichal: idk! i thought i had  dropped this  but it maybe snuck back in? can update

src/kv/bulk.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    let response = client.request(&DeleteBulk {
+        account_identifier: &target.account_id,
+        namespace_identifier: namespace_id,
+        bulk_keys: keys,
+    });
+
+    match response {
+        Ok(_) => Ok(()),
+        Err(e) => failure::bail!("{}", kv::format_error(e)),
```
EverlastingBugstopper, line 65: same

src/settings/toml/target.rs
```diff
@@ -9,7 +9,7 @@ use std::path::PathBuf;
 
 use serde::{Deserialize, Serialize};
 
-#[derive(Clone, Debug, Deserialize, Serialize)]
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
```
EverlastingBugstopper, line 12: what do we use `Default` for? what is a `Default` target?

ashleymichal: this snuck in while i was doing some work with prototyping on  the deployments project. all a default  here would be is each field would be set to its type's default -- empty `String`, `None`  for `Option`, etc. deriving this here is just a convenience, i can leave it or not.

src/kv/bulk.rs
ashleymichal: ugh i  hate that this is hereeeeeeeee next stop `thiserror`

PR #1336, opened 2020-05-28
Add better error messages without breaking publish

jspspike:
This redoes #1329 without breaking `wrangler publish`. Addresses #1157

### DIFF
--- a/src/commands/kv/namespace/site.rs
+++ b/src/commands/kv/namespace/site.rs
@@ -3,6 +3,7 @@ use cloudflare::framework::apiclient::ApiClient;
 use cloudflare::framework::response::ApiFailure;
 
 use crate::commands::kv;
+use crate::http;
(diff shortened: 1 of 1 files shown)


src/commands/kv/namespace/site.rs
```diff
@@ -63,6 +62,15 @@ fn get_id_from_namespace_list(
             .find(|ns| ns.title == title)
             .unwrap()
             .to_owned()),
-        Err(e) => failure::bail!(kv::format_error(e)),
+        Err(e) => failure::bail!("{}", http::format_error(e, Some(&error_suggestions))),
```
ashleymichal, line 65: kv  actually has a format  error fn:  https://github.com/cloudflare/wrangler/blob/josh/fix-broken-publish/src/commands/kv/mod.rs#L35 try  using/adding to t hat.

PR #1339, opened 2020-05-28
Batched kv put and delete

### LINKED ISSUE #1191: How to Handle Builds Where More Than 10k Pages Change?
### 🐛 Bug Report

My SSG generates new pages each build. The first `wrangler publish` always works, but subsequent ones fail due to the 10k key limit. 

`Number of keys to delete (17397) exceeds max of 10000` 

Is there a workaround for this besides deleting the store ahead of time?  (This works for staging but not production).

`- yes | wrangler kv:namespace delete --namespace-id={ID for __static-staging-workers_sites_assets}`

This said, I was reading the code for for `getAssetFromKV` and it appears that a possible workaround would be to:

1 - Create a new KV store for the content.
2 - Specify wrangler to use that store for the content.
3 - After deploy delete the old KV store. 

Is there a way to do step 2? I've read all of the docs and can't find anything about how to specify the store for the worker to use. 

#### Environment

* operating system: Docker: node:latest

#### Steps to reproduce
1 - Build a project with > 10k pages. 
2 - Make a change to more than 10k of those pages. 

The build be successful.

`Number of keys to delete (17397) exceeds max of 10000`

jspspike:
Made kv bulk put and delete do `MAX_PAIRS` at a time. Fixes #1191


src/commands/kv/bulk/delete.rs
```diff
@@ -65,23 +66,39 @@ pub fn delete_bulk(
...
-    if keys.len() > MAX_PAIRS {
-        failure::bail!(
-            "Number of keys to delete ({}) exceeds max of {}",
-            keys.len(),
-            MAX_PAIRS
-        );
-    }
+    let mut pairs = keys;
+
+    message::working("Starting KV Bulk Delete");
```
EverlastingBugstopper, line 71:
```suggestion
    message::working(&format!("deleting {} key value pairs", pairs.len());
```

src/commands/kv/bulk/delete.rs
```diff
@@ -65,23 +66,39 @@ pub fn delete_bulk(
...
+        if let Err(e) = response {
+            failure::bail!("{}", kv::format_error(e))
+        }
+    }
 
-    match response {
-        Ok(_) => Ok(()),
-        Err(e) => failure::bail!("{}", kv::format_error(e)),
+    if let Some(pb) = &progress_bar {
+        pb.finish_with_message("Done KV Bulk Delete");
```
EverlastingBugstopper, line 100:
```suggestion
        pb.finish_with_message(&format!("deleted {} key value pairs", pairs.len()));
```

src/commands/kv/bulk/put.rs
```diff
@@ -39,20 +40,35 @@ pub fn put(
...
 
-    // Validate that bulk upload is within size constraints
-    if pairs.len() > MAX_PAIRS {
-        failure::bail!(
-            "Number of key-value pairs to upload ({}) exceeds max of {}",
-            pairs.len(),
-            MAX_PAIRS
-        );
-    }
+    message::working("Starting KV Bulk Put");
```
EverlastingBugstopper, line 43:
```suggestion
    message::working(&format!("uploading {} key value pairs", pairs.len()));
```

src/commands/kv/bulk/put.rs
```diff
@@ -39,20 +40,35 @@ pub fn put(
...
+            failure::bail!("{}", kv::format_error(e));
+        }
+
+        if let Some(pb) = &progress_bar {
+            pb.inc(p.len() as u64);
+        }
+    }
+
+    if let Some(pb) = &progress_bar {
+        pb.finish_with_message("Done KV Bulk Put");
```
EverlastingBugstopper, line 68:
```suggestion
        pb.finish_with_message(&format!("uploaded {} key value pairs", pairs.len()));
```

src/commands/kv/bulk/put.rs
```diff
@@ -39,20 +40,35 @@ pub fn put(
...
     }
+
+    message::success("Success");
```
EverlastingBugstopper, line 71: probably dont need this line since we have the `uploaded` message w/the progress bar

if you think we should keep it, we should add it to the bulk delete command as well

jspspike: I just left it from before but if you want it gone that works

EverlastingBugstopper:
so i checked this branch out and ran it locally - it seems like the `finish_with_message` doesn't actually work for some reason. let's see if we can get it to work (it should completely replace the progress bar with the correct message), but if we can't, then lets put a pin in it and do a `message::success()` with the same message we were trying to put in the progress bar

EverlastingBugstopper:
Still looks like we have some inconsistencies w/the output here

```console
wrangler kv:bulk put --binding NAMESPACE ./bulk.json
🌀  Starting KV Bulk Put
███████████████████████████████████████████ 21000/21000
```

```console
$ yes | wrangler kv:bulk delete --binding NAMESPACE ./bulk.json
🌀  deleting 21000 key value pairs
███████████████████████████████████████████ 21000/21000
✨  Success
```

in addition to the difference in output, it seems like for the bulk upload, the progress bar does not appear until after it has completed the first tick, whereas the delete progress bar appears immediately upon sending the first request to the Cloudflare API

EverlastingBugstopper:
ofc as soon as i say it's good to go - there are merge conflicts. 

if they're real rough, tag in @ashleymichal - they're from her recent refactor :)

Issue #1064, opened 2020-02-14
Allow a user to specify which wrangler.toml to use with commands


jroyal:
#### Overview and problem statement

I would like to be able to specify which `wrangler.toml` file to use when running commands like `wrangler publish`. 

I have an example project that I would like to leave an empty toml file in so users can do things like `wrangler generate`. An example repo is https://github.com/cloudflare/worker-template-fetch. However I still want to be able to test it before pushing it out to github, so I need a valid `wrangler.toml` in order to actually publish it. To do that I created a `.gitignored` wrangler.toml that contains my real `wrangler.toml`. 

I currently work around this issue by just changing the names of the template `wrangler.toml` to something else and changing my template to `wrangler.toml` but it would be nice if I could just specify the correct file when running wrangler commands. 

I realize that environments solve most of the problems that having multiple toml files would solve, but in this particular case I don't want to ever commit my toml file to git. So i'm stuck with my current workaround.

#### Basic example

It could work like docker compose, where if you include a `-f` flag it will use that instead of the default `docker-compose.yaml` file.

```wrangler publish -f mywrangler.toml```

EverlastingBugstopper:
Hey so this is something we thought about doing for environments but decided against doing because there are some ways to make templates work without having separate tomls. You can set the environment variables `CF_ACCOUNT_ID` and `CF_ZONE_ID` and leave the `account_id = ""` in your `wrangler.toml` and they will be overridden. If you wanted new team members to know to follow that process, you could have a gitignored `example.wrangler.toml` that has the ids filled in and a comment instructing them to set those environment variables.

That being said, this may still be worth implementing but there'd be a decent bit of cleanup involved (see #478).

What do you think?

jroyal:
Well if it was only account and zone id that would work, but with the release of secrets and environment variables there are some configuration items there that you can't overwrite with system environment variables. So I would still think it would be valuable.

PR #1350, opened 2020-05-30
Added parameter to override default location to config file

luanraithz:
Enables: `cargo build -c ./wrangler-test-1.toml`.

The parameter was added to the following commands:
* kv:namespace
* kv:key
* kv:bulk
* route
* secret
* build
* preview
* dev
* publish
* subdomain
* tail

Which basically were the ones that made use of the `config_path` variable.
I'm still working, but I'm kinda stuck, as I don't really know how to write unity tests for it yet (any reference to help will be welcome!).


luanraithz:
@EverlastingBugstopper Can you take a look at the test case? If it makes sense I may write more cases.

EverlastingBugstopper:
hey @luanraithz - that test case seems good! i need to check this branch out and test it out myself, and maybe do a bit of thinking about what we want to call the argument. `"file"` seems a bit.. vague in my opinion but i'm not sure exactly what i would use instead. i'll do some thinking and get back to you. thanks for the contribution!

EverlastingBugstopper:
I think a good name for the flag would be `--config` instead. @luanraithz - do you mind updating `--file` -> `--config` and resolving the conflicts?

luanraithz:
Thanks for your review! `config` is really better, I will change it!

luanraithz:
@EverlastingBugstopper I also changed the short name to `c` as `f` just stopped making sense.

EverlastingBugstopper:
Hey @luanraithz - so I checked your fork out and ran some tests. First thing I found was that it doesn't seem to work with `wrangler dev` - didn't check the other commands.

```console
$ wrangler dev --config anotherfile.toml
Error: wrangler.toml not found; run `wrangler init` to create one.
```

would probably also be good to update some error messages so they use the value passed in to "--config" instead of printing a hard coded `wrangler.toml`

luanraithz:
The problem is that I get the `config` param using the matches for the whole command instead of using just the matches for a subcommand.
I'm really sorry, I didn't catch that because the test I wrote was for the build subcommand, which gets the path from the subprogramm matches. I will fix that.
Would you rather repeat the logic to get the config inside every subprogramm or add the `config` as a parameter to the whole app @EverlastingBugstopper ?

EverlastingBugstopper:
Hey no worries, that's why we have the review process 😄 

> Would you rather repeat the logic to get the config inside every subprogramm or add the config as a parameter to the whole app

i don't think it makes sense to add it to the whole app because commands like `init` and `generate` do not require a `wrangler.toml` in order to run them, therefore it would not make sense to accept a `--config` flag on those commands. I think adding them everywhere we need them makes the most sense, similar to how we have `--preview` flags sitting around for commands that need them. 😄 

Thanks!

luanraithz:
@EverlastingBugstopper It's seems to be fixed now, I've also changed the message to display the file's name.

I don't really know if it's better to display the path with the file's name, your feedback about that would be nice!

EverlastingBugstopper:
fixes #1064

luanraithz:
I was forced to use the `to_str` in the if statement, as the `file_name` would be equal do `wrangler.toml` and the default path would be `./wrangler.toml`

luanraithz:
Damn, there are a lot of `fmt` commits, once it goes to the master branch, a squash would be nice

EverlastingBugstopper:
@luanraithz - you can for sure feel free to squash these on your own branch if you'd like 😄

EverlastingBugstopper:
ok @luanraithz - this seems to work well! i think in order for this to be considered feature-complete, we'll need to update error messages across the codebase to be formatted with the name of the config file that was passed instead of `wrangler.toml`.

cc @ashleymichal for ideas on how to do this, since there are a _bunch_ of instances lying around. i guess maybe an alternative would be to not say `wrangler.toml` in the error messages at all and instead say "your manifest" or "your wrangler configuration file" or something like that... 

possible solution may be to add `config_file_path` field to `Manifest` and then access that any time an error needs to be thrown referencing a config file.

luanraithz:
Sure! I'm just waiting for @ashleymichal feedback and I will do it.
I think that using a message such as "your configuration file" is better, otherwise when adding new features the developers would need to remember that the wrangler file can be different from `wrangler.toml`.

src/main.rs
```diff
@@ -101,6 +101,12 @@ fn run() -> Result<(), failure::Error> {
         .takes_value(false)
         .help("toggle verbose output");
 
+    let wrangler_file = Arg::with_name("config")
+        .short("c")
+        .takes_value(true)
+        .long("config")
+        .help("File path to override the default wrangler.toml file ");
```
ashleymichal, line 108:
```suggestion
        .help("Path to configuration file. Defaults to `./wrangler.toml`");
```

ashleymichal:
probably just "your configuration file" is sufficient for this change.

luanraithz:
I think I changed all of them (at least the ones that make sense)

EverlastingBugstopper:
we're so close on this one, there's just one thing that i think we should do before we merge this up:

each of the following subcommands _also_ have subcommands:

```
kv:namespace -> create/delete/list
kv:bulk -> delete/put
kv:key -> delete/list/put
route -> delete/list
secret -> delete/list/put
```

in my view, usage for these should be like this:

`wrangler kv:namespace create -c ra.toml BINDING`

instead of how it currently is in this pr:

`wrangler kv:namespace -c ra.toml create BINDING`

luanraithz:
@EverlastingBugstopper Can you take a look?
Two questions that I had while coding:
* Are there any tests for this commands? I didn't find any for this specific commands.
* Does it make any sense having statements such as the following?
```rust
("", None) => message::warn("secret expects a subcommand")
```
Doesn't `clap` ensure that there will be a subcommand?

If it doesn't make sense to have them I would be happy to create another pull request removing it😄.

EverlastingBugstopper:
> are there any tests

the best place to test the changes here are going to be in the `tests` subdirectory and find the ones that actually run the commands with the `Command` struct. if you want to add some extra tests for commands like `preview` you can create a file with a special config name and pass it as an argument in the test. 

we dont really have good tests for commands that require an authenticated user because we haven't spent the time to create robust mocks so for now integration tests on stuff that doesn't require an account is the way to go

> Does it make any sense having statements such as the following?

probably not, feel free to open a PR removing 😄

PR #1353, opened 2020-06-01
Feat/rename subdomain

nprogers:
This is my first PR, so expect there to be a lot of suggestions. My google searches looked something like `rust integers` so that's where we're at right now 😆 

- No error on new subdomain name the same as old subdomain name.
- Interactive, y/n to proceed with rename. Would ideally have more specific messaging around what scripts will be modified.
- No existing subdomain command tests. I need some help to get that started if that's something we should add.
- Second commit is correcting for subdomain availability api response. Not exactly related, but not unrelated either.


EverlastingBugstopper:
@nprogers - i'm assuming you used your work email to commit these, could you add your work email as a valid email address for your account so that the commits are properly attributed to your account? right now they are showing up as anonymous commits

[here](https://help.github.com/en/enterprise/2.15/user/articles/adding-an-email-address-to-your-github-account) is info on how to do that

src/commands/subdomain.rs
```diff
@@ -105,15 +117,27 @@ pub fn set_subdomain(name: &str, user: &GlobalUser, target: &Target) -> Result<(
...
+            let msg = format!("You have previously registered {}.workers.dev", subdomain);
+            message::success(&msg);
+            return Ok(());
         } else {
-            format!("This account already has a registered subdomain. You can only register one subdomain per account. Your subdomain is {}.workers.dev", subdomain)
-        };
-        failure::bail!(msg)
-    } else {
-        register_subdomain(&name, &user, &target)
+            // probably need to list all the effected scripts?
```
EverlastingBugstopper, line 125: yes - should have parity w/what we do in the UI. UI lists all renames with `"worker.avery.workers.dev" => "worker.natalie.workers.dev".

EverlastingBugstopper:
this looks good - just need to add that additional messaging about moving. as for tests - we need a solution to mock out API calls before we could test this reasonably i think.

.gitignore
```diff
@@ -2,3 +2,4 @@
 **/*.rs.bk
 config/default.toml
 npm/node_modules
+.idea/
```
ashleymichal, line 5: ?

nprogers: That is JetBrains IDE.

src/commands/subdomain.rs
nprogers: Done.

src/commands/subdomain.rs
```diff
@@ -105,15 +128,48 @@ pub fn set_subdomain(name: &str, user: &GlobalUser, target: &Target) -> Result<(
     }
     let subdomain = Subdomain::get(&target.account_id, user)?;
     if let Some(subdomain) = subdomain {
-        let msg = if subdomain == name {
-            format!("You have previously registered {}.workers.dev", subdomain)
+        if subdomain == name {
+            let msg = format!("You have previously registered {}.workers.dev", subdomain);
```
ashleymichal, line 132: nit: (i realize this was the pre-existing verbiage)
```suggestion
            let msg = format!("You have already registered {}.workers.dev", subdomain);
```

EverlastingBugstopper:
re: gitignore - i could not care less

.gitignore
ashleymichal:
```suggestion
```

ashleymichal: i'm  going to  remove for now i think.

PR #1357, opened 2020-06-02
adds preview_id to namespace manifest definition

### LINKED ISSUE #1032: Use preview namespace for regular workers
For Workers Sites, we create a separate preview namespace so that we do not impact production namespaces. We do not do this for regular workers and we should.

EverlastingBugstopper:
fixes #1032


src/upload/mod.rs
```diff
@@ -19,7 +19,7 @@ pub fn script(
         target.account_id, target.name,
     );
 
-    let script_upload_form = form::build(target, asset_manifest)?;
+    let script_upload_form = form::build_production(target, asset_manifest)?;
```
ashleymichal, line 22: 💡 maybe we pass preview to `get_target` to skip all this preview bool mess?

src/settings/toml/target.rs
```diff
@@ -13,7 +14,7 @@ use serde::{Deserialize, Serialize};
 pub struct Target {
     pub account_id: String,
     #[serde(rename = "kv-namespaces")]
-    pub kv_namespaces: Option<Vec<KvNamespace>>,
+    pub kv_namespaces: Option<Vec<ConfigKvNamespace>>,
```
ashleymichal, line 17: i think i would flip these, actually, and name the thing you attach to `Manifest` `ConfigKvNamespace`  and use just `KvNamespace`s for `Target`

PR #1359, opened 2020-06-03
set kv_namespaces in manifest::get_target

ashleymichal:


src/commands/publish.rs
```diff
@@ -123,19 +123,16 @@ fn validate_target_required_fields_present(target: &Target) -> Result<(), failur
...
-        Some(kv_namespaces) => {
-            for kv in kv_namespaces {
-                if kv.binding.is_empty() {
-                    missing_fields.push("kv-namespace binding")
-                }
-
-                if kv.id.is_empty() {
-                    missing_fields.push("kv-namespace id")
-                }
+    if !target.kv_namespaces.is_empty() {
```
EverlastingBugstopper, line 126: is this necessary? if it's empty wont it just never enter the loop?

src/main.rs
```diff
@@ -580,6 +580,7 @@ fn run() -> Result<(), failure::Error> {
         .get_matches();
 
     let config_path = Path::new("./wrangler.toml");
+    let mut preview = false;
```
EverlastingBugstopper, line 583:
```suggestion
    let mut is_preview = false;
```

src/preview/upload.rs
```diff
@@ -188,15 +183,15 @@ fn unauthenticated_upload(target: &Target) -> Result<Preview, failure::Error> {
...
             "KV Namespaces are not supported in preview without setting API credentials and account_id",
         );
         let mut target = target.clone();
-        target.kv_namespaces = None;
-        upload::form::build_preview(&target, None)?
+        target.kv_namespaces = Vec::new();
+        upload::form::build(&target, None)?
     } else {
-        upload::form::build_preview(&target, None)?
+        upload::form::build(&target, None)?
```
EverlastingBugstopper, line 194: i wonder if we should just do the warning check in its own if and then assign `script_upload_form` outside of the conditional. doesn't look like we need the conditional here since both arms return the same fn call

src/commands/publish.rs
ashleymichal: good catch lol i was just blindly correcting compiler errors.

PR #1371, opened 2020-06-08
Pretty print recieved json

### LINKED ISSUE #1249: [dev] get better console.log() messages
### 💡 Feature request

It would be great if `dev` provided better `console.log()` messages.

When using `console.log()` objects get truncated. Not only nested objects and array but look at the ellipsis at the far right:

image

A common trick in Node is to use `JSON.stringify` to get the full log, but when doing this line breaks and white space are printed quite literally...

image

The best solution would be to print complete objects and nested objects n levels deep, but if that is not feasible at least interpret strings correctly.

jspspike:
Update chrome-devtools-rs to unscape strings fixes #1249


jspspike:
Closes #1249

src/commands/dev/socket.rs
```diff
@@ -71,12 +71,23 @@ async fn print_ws_messages(
...
                         failure::format_err!("this event could not be parsed:\n{}", e)
                     });
+
                 if let Ok(protocol::Runtime::Event(event)) = parsed_message {
-                    println!("{}", event);
+                    // Try to parse json to pretty print, otherwise just print string
+                    let json_parse: Result<serde_json::Value, serde_json::Error> =
+                        serde_json::from_str(&*event.to_string());
+                    if let Ok(json) = json_parse {
+                        if let Ok(json_str) = serde_json::to_string_pretty(&json) {
```
EverlastingBugstopper, line 85: if this fails we should print `&json` by itself, not do nothing

jspspike: Done

Issue #521, opened 2019-09-04
The env value should be passed to webpack


xtuc:
_Originally posted by @xtuc in https://github.com/cloudflare/wrangler/pull/500#issuecomment-526829405_

anapsix:
Having webpack aware of "env", passed via `wrangler build --env "some-env"` would allow for greater flexibility during webpack action.
Looking forward this feature!

pyk:
For those who are stumbling upon this thread from search engine.

The alternative is to use `process.env.VARIABLE_NAME` inside your webpack config.

For example:

```js
module.exports = (env, args) => {
    const myVar = process.env.MY_VARIABLE;
};
```

Then run your wrangler using the following command:

```
MY_VARIABLE=production wrangler publish -e production
```

I hope it helps 🙏

tguvdamm:
Introducing env vars works, but it's messy and I really want to avoid them wherever possible. Being able to pass on some parameters to the underlying builder would help a lot, or just even the env var.

PR #1385, opened 2020-06-15
Pass wrangler env to webpack

xtuc:
Closes https://github.com/cloudflare/wrangler/issues/521

This change passes the first argument `env` when webpack exposes a
function to the env passed to wrangler.

The object form of webpack's env is not supported, `env` will always be
a single string.

### DIFF
--- a/src/build/mod.rs
+++ b/src/build/mod.rs
@@ -6,7 +6,7 @@ use crate::{commands, install};
 use std::path::PathBuf;
 use std::process::Command;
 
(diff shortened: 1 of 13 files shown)


xtuc:
Maybe there's a better way than passing the `build_env` everywhere. We could move it and the `verbose` flag to a build option struct.

ashleymichal:
> Maybe there's a better way than passing the `build_env` everywhere. We could move it and the `verbose` flag to a build option struct.

👍 i did something similar with "DeployConfig" and intended to extend it to build and upload steps anyway.

PR #1392, opened 2020-06-16
increases default HTTP timeout to 1 min

### LINKED ISSUE #1386: Error sending request for url (...): operation timed out
The issue lies [here](https://github.com/cloudflare/wrangler/blob/master/src/http/legacy.rs#L40).

I'm writing some web workers on a slow internet speed. When I want to preview my web worker (`wrangler preview --watch`), my internet speed requires a full second to upload 50KiB, and I have 2.3MiB of uncompressed JavaScript to upload. This is 46 seconds, and surpasses the 30 second timeout limit.

A simple fix would be to utilize the user's configuration (located at <user's home directory>\.wrangler\config\default.toml) and allow them to specify an `http_timeout` in seconds.

Current workaround is to compile from source with a longer timeout.

EverlastingBugstopper:
should fix #1386

### DIFF
--- a/src/http/mod.rs
+++ b/src/http/mod.rs
@@ -2,6 +2,7 @@ pub(self) mod cf;
 pub(crate) mod feature;
 pub(self) mod legacy;
 
+pub const DEFAULT_HTTP_TIMEOUT_SECONDS: u64 = 60;
 pub use cf::{cf_v4_api_client_async, cf_v4_client, featured_cf_v4_client, format_error};
 pub use feature::Feature;
 pub use legacy::{client, featured_legacy_auth_client, legacy_auth_client};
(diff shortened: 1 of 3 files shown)


src/http/mod.rs
```diff
@@ -2,6 +2,7 @@ pub(self) mod cf;
 pub(crate) mod feature;
 pub(self) mod legacy;
 
+pub const DEFAULT_HTTP_TIMEOUT_SECONDS: u64 = 5 * 60;
```
ispivey, line 5: This is 5 minutes, not 1 minute, right? Or am I misunderstanding how this value is used?

ashleymichal: bah. you are correct. the  five minutes should be for the  other thing.

ashleymichal, line 5:
```suggestion
pub const DEFAULT_HTTP_TIMEOUT_SECONDS: u64 = 1 * 60;
```

PR #1400, opened 2020-06-19
move batching into bulk module code

### LINKED ISSUE #1398: Critical: Error: Number of keys to delete (16010) exceeds max of 10000
#### Environment

* operating system:
* output of `rustc -V`: rustc 1.41.0
* output of `node -v`: v10.20.1
* output of `wrangler -V`: wrangler 1.9.2
* contents of `wrangler.toml`:

```
name = "jwr-nuxt"
type = "webpack"
account_id = "0de41866465843875a6d452ca5bd6e1f"
workers_dev = false
route = "thejewelleryroom.com/*"
zone_id = "5323273892234c0ce2ebcd182644e8fb"
kv-namespaces = [
	{ binding = "REDIRECTS", id = "a79e38aac1fe464f99a8e8fc75d2b16f" }
]
[site]
bucket = "dist"
entry-point = "workers-site"

[env.beta]
name = "jwr-nuxt-beta"
entry-point = "workers-site-beta"
bucket = "dist"
route = "beta.thejewelleryroom.com/*"
kv-namespaces = [
	{ binding = "REDIRECTS", id = "a79e38aac1fe464f99a8e8fc75d2b16f" }
]
```

#### Steps to reproduce

Running wrangler publish

Should publish

    Error: Number of keys to delete (16010) exceeds max of 10000

This is pretty critical. Our site is currently down because I by mistake published our beta site to production, and now I cannot publish our production site anymore.

ashleymichal:
fixes #1398


src/kv/bulk.rs
```diff
@@ -33,36 +39,106 @@ pub fn put(
...
+    Ok(())
+}
+
+fn batch_keys_values(mut pairs: Vec<KeyValuePair>) -> Vec<Vec<KeyValuePair>> {
+    let mut batches: Vec<Vec<KeyValuePair>> = Vec::new();
 
-    match response {
-        Ok(_) => Ok(()),
-        Err(e) => failure::bail!("{}", format_error(e)),
+    if !pairs.is_empty() {
```
ashleymichal, line 95: this if is probably unnecessary; it just saves initializing a few local variables. felt cute might delete later.

src/kv/bulk.rs
```diff
@@ -33,36 +39,106 @@ pub fn put(
...
+
+        if let Some(pb) = &progress_bar {
+            pb.inc(b.len() as u64);
+        }
+    }
+
+    Ok(())
+}
+
+fn batch_keys_values(mut pairs: Vec<KeyValuePair>) -> Vec<Vec<KeyValuePair>> {
```
ashleymichal, line 92: standardizing this to use the logic from sites, which takes into account not only the number of keys but the maximum total payload size.

src/main.rs
```diff
@@ -757,9 +757,7 @@ fn run() -> Result<(), failure::Error> {
         let mut target = manifest.get_target(env, is_preview)?;
         let deploy_config = manifest.deploy_config(env)?;
 
-        let verbose = matches.is_present("verbose");
```
ashleymichal, line 760: removed this bc the reason for the flag made  sense when we were printing all of the file names that were being uploaded; now we just use a progress bar so we should just show it all the time and make the verbose silent.

src/preview/upload.rs
```diff
@@ -70,15 +69,15 @@ pub fn upload(
                         message::info("Uploading updated files...");
                     }
 
-                    upload_files(target, user, &site_namespace.id, to_upload)?;
+                    bulk::put(target, user, &site_namespace.id, to_upload, &None)?;
```
ashleymichal, line 72: should  we use a progress bar here? didn't before but can add.

src/sites/upload.rs
```diff
@@ -1,82 +0,0 @@
-use indicatif::{ProgressBar, ProgressStyle};
```
ashleymichal, line 1: moved this logic to `bulk::batch_keys_values` since it was no longer specific to sites anyway

src/preview/upload.rs
jspspike: That would make sense to me, right now it just says nothing and it might look like it's hanging

PR #1413, opened 2020-06-25
[dev] allow multiple response header values

### LINKED ISSUE #1412: [dev] handle multiple value headers
I'm unable to get `wrangler dev` to return multiple `Set-Cookie` headers in a worker response. I've tried in both of the following ways: 

_explicitly appending them_
```js
// not working, only last gets sent (`cookie2`)

resp.headers.append('Set-Cookie', 'cookie1=true')
resp.headers.append('Set-Cookie', 'cookie2=true')
```
_passing them in the options object of the response_
```js
// not working, only last gets sent (`cookie2`)

headers: {
   'Set-Cookie': 'cookie1=true',
   'Set-Cookie': 'cookie2=true'
}
```

I've so far only been trying to get this to work using `wrangler dev`... perhaps it's a wider Workers issue/limitation? Perhaps I'm missing something obvious here?

_Originally posted by @yet-another-dev in https://github.com/cloudflare/wrangler/issues/1047#issuecomment-649705103_

---

The issue here is that we do not properly handle response headers that have multiple values. We iterate over them all and end up overwriting each response header with the following one, which is why @yet-another-dev only sees the last cookie header in the response

EverlastingBugstopper:
fixes #1412

Before, `wrangler dev` would not properly handle response headers that have multiple values. We would iterate over all response headers coming from the Workers runtime, and "insert" them into the header map instead of appending them. This is no longer the case and response headers should now work as expected.

### DIFF
--- a/src/commands/dev/gcs/headers.rs
+++ b/src/commands/dev/gcs/headers.rs
@@ -60,7 +60,7 @@ fn strip_response_headers_prefix(parts: &mut ResponseParts) -> Result<(), failur
         if name.starts_with(HEADER_PREFIX) {
             let header_name = &name[HEADER_PREFIX.len()..];
             let header_name = HeaderName::from_bytes(header_name.as_bytes())?;
-            headers.insert(header_name, value.clone());
+            headers.append(header_name, value.clone());
         }
     }
     parts.headers = headers;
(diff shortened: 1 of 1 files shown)


src/commands/dev/gcs/headers.rs
```diff
@@ -60,7 +60,7 @@ fn strip_response_headers_prefix(parts: &mut ResponseParts) -> Result<(), failur
         if name.starts_with(HEADER_PREFIX) {
             let header_name = &name[HEADER_PREFIX.len()..];
             let header_name = HeaderName::from_bytes(header_name.as_bytes())?;
-            headers.insert(header_name, value.clone());
+            headers.append(header_name, value.clone());
```
ashleymichal, line 63: should we  write a unit test for this?

EverlastingBugstopper: should probably write unit tests for dev in general

EverlastingBugstopper: but - yes

ashleymichal: sure; this  one  seems simple  enough to get the ball rolling tho

src/commands/dev/gcs/headers.rs
```diff
@@ -81,3 +81,29 @@ fn set_response_status(parts: &mut ResponseParts) -> Result<(), failure::Error>
...
+            .header("cf-ew-raw-Set-Cookie", &second_cookie)
+            .body(())
+            .unwrap();
+        let (mut parts, body) = response.into_parts();
+        strip_response_headers_prefix(&mut parts).unwrap();
+        let response = Response::from_parts(parts, body);
+        let cookie_jar = response.headers().get_all("Set-Cookie");
+
+        let mut iter = cookie_jar.iter();
+        assert_eq!(&first_cookie, iter.next().unwrap());
```
ashleymichal, line 105: nit: do we care that these are in order? or just that they are both included in the headers?

EverlastingBugstopper: just that they are both there

ashleymichal: i wanna say there's a `.contains` you can use...

PR #1414, opened 2020-06-25
allow creation of preview namespace when namespace exists in toml

EverlastingBugstopper:
when we introduced KV preview namespaces, we made sure to add nice messages when creating new namespaces so people could easily add the new namespace id to their wrangler.toml in the correct place. 

However, we missed a very common case where folks already have a production namespace defined in their wrangler.toml and they want to add a preview namespace. When this is the case, we return an error message intended to only be thrown when running either `wrangler preview` or `wrangler dev`.


src/commands/kv/namespace/create.rs
```diff
@@ -14,19 +14,19 @@ pub fn run(
...
+    let mut title = format!("{}-{}", worker_name, binding);
     if is_preview {
         title.push_str("_preview");
     }
     let msg = format!("Creating namespace with title \"{}\"", title);
     message::working(&msg);
 
     let client = http::cf_v4_client(user)?;
-    let result = create(&client, &target, &title);
+    let result = create(&client, &account_id, &title);
```
ashleymichal, line 29: we should validate the presence of account id in this command; since manifest allows it to be missing or empty, we will get back the cryptic api gateway error.

PR #1445, opened 2020-07-10
[dev] [wip] accept https connections

EverlastingBugstopper:
http://127.0.0.1:8787 should preview http://myzone.com and https://127.0.0.1:8787 should preview https://myzone.com


sample.pem
```diff
@@ -0,0 +1,113 @@
+-----BEGIN CERTIFICATE-----
```
EverlastingBugstopper, line 1: eventually this will be replaced by auto-generated self-signed certificates with the ability for a user to provide their own - this is just for a proof of concept 😄

EverlastingBugstopper:
for posterity - was encountering this issue:

when compiling i get this error:

```console
$ cargo check
error: future cannot be sent between threads safely
   --> src/commands/dev/gcs/mod.rs:75:22
    |
75  |         let server = tokio::spawn(serve(server_config, Arc::clone(&preview_id)));
    |                      ^^^^^^^^^^^^ future returned by `serve` is not `Send`
    | 
   ::: /Users/averyharnish/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-0.2.21/src/task/spawn.rs:127:21
    |
127 |         T: Future + Send + 'static,
    |                     ---- required by this bound in `tokio::task::spawn::spawn`
    |
    = help: the trait `std::marker::Send` is not implemented for `dyn futures_core::stream::Stream<Item = std::result::Result<tokio_rustls::server::TlsStream<tokio::net::tcp::stream::TcpStream>, std::io::Error>>`
note: future is not `Send` as it awaits another future which is not `Send`
   --> src/commands/dev/gcs/server.rs:123:21
    |
123 |     if let Err(e) = server.await {
    |                     ^^^^^^ await occurs here on type `hyper::server::Server<commands::dev::tls::HyperAcceptor<'_>, hyper::service::make::MakeServiceFn<[closure@src/commands/dev/gcs/server.rs:59:35: 112:6 client:hyper::client::Client<hyper_rustls::connector::HttpsConnector<hyper::client::connect::http::HttpConnector>>, server_config:commands::dev::server_config::ServerConfig, preview_id:std::sync::Arc<std::sync::Mutex<std::string::String>>]>>`, which is not `Send`

error: aborting due to previous error

error: could not compile `wrangler`.

To learn more, run the command again with --verbose.
```

this is a bit strange since it seems to me like the `server.await` should still be the same type. I wonder if I need to do something with `HyperAcceptor` to make it `Send`...

EverlastingBugstopper:
currently running into some more issues:

when curling https:

```console
$ curl https://127.0.0.1:8787 --insecure
curl: (92) HTTP/2 stream 1 was not closed cleanly: INTERNAL_ERROR (err 2)
```

---

when curling http

```console
$ curl http://127.0.0.1:8787            
curl: (52) Empty reply from server

```

```console
$ wrangler dev
server error: error accepting connection: TLS Error: Custom { kind: InvalidData, error: CorruptMessage }
```

---

so - i need to figure out why https isn't working properly, and i need to also modify my code to allow http requests in _addition_ to https requests

EverlastingBugstopper:
ok - i have it accepting `curl https://127.0.0.1:8787` just fine now! big problem is that it no longer accepts `http` connections, AND if it happens to receive an http connection, it then seems to corrupt the parser. To reproduce:

```console
$ wrangler dev
💁  watching "./"
executing worker
[2020-07-14 11:51:03] GET example.com/ HTTP/1.1 200 OK
server error: error accepting connection: TLS Error: Custom { kind: InvalidData, error: CorruptMessage }
```

```console
$ curl --insecure https://127.0.0.1:8787
Hello, World!
```

```console
$ curl http://127.0.0.1:8787
curl: (52) Empty reply from server
```

```console
$ curl --insecure https://127.0.0.1:8787
curl: (7) Failed to connect to 127.0.0.1 port 8787: Connection refused
```

EverlastingBugstopper:
i found these two examples: https://github.com/izderadicka/tokio-tls/blob/new-tokio/examples/hyper-server.rs and https://users.rust-lang.org/t/use-hyper-server-with-tls/18535, but they seem to use a lower level `Http` API that I'm not currently using, and I'm not quite sure how to fit this in with the way things are currently set up.

PR #1451, opened 2020-07-14
[dev] upload workers sites bindings for edge dev

EverlastingBugstopper:
before we weren't handling any workers sites stuff for edge dev - now we do


src/commands/dev/edge/setup.rs
```diff
@@ -31,6 +50,14 @@ pub(super) fn upload(
         .send()?
         .error_for_status()?;
 
+    if !to_delete.is_empty() {
```
jwheels, line 53: Just double checking... this won't cause us to delete any production files, right?

ashleymichal: that really depends on whether they put the same namespace in for preview_id.

EverlastingBugstopper: yup - the namespace returned here is the one they have defined for their preview sesssions. we dont stop them from having production namespace in preview, but we print a warning out when they're the same that they are in dangerous territory.

jwheels: cool, sounds good!

PR #1461, opened 2020-07-23
Check if rustc exists before building a Rust project type

### LINKED ISSUE #487: Ensure that rustc and other dependencies are installed before building a rust project
### 🐛 Bug Report

`wrangler build` produces an unhelpful error message when rustc isn't in the path.

#### Environment

* operating system:
```
Microsoft Windows [Version 10.0.18362.295]
Command Prompt and Bash has the same problem though.
```
* output of `rustc -V`:
```
'rustc' is not recognized as an internal or external command,
operable program or batch file.
```
* output of `node -v`:
```
v8.11.3
```
* output of `wrangler -V`:
```
 wrangler 1.1.1
```

#### Steps to reproduce

I installed wrangler using npm, and created a rust project, then called `wrangler build`.

Either a successful build, or a helpful error message prompting me to use rustup to install rustc and the dependencies.

When I called `wrangler build` I got the error:
```
>wrangler build
⬇️ Installing wasm-pack...
 Compiling your project to WebAssembly...
Error: The system cannot find the file specified. (os error 2)
Error: failed to execute `"cmd" "/C" "...\\AppData\\Local\\.wrangler\\wasm-pack-7824e3bac6af138a\\wasm-pack.exe" "build" "--target" "no-modules"`: exited with exit code: 1
```
The underlying issue is that wasm-pack doesn't return a helpful error either, just an exit code.

ObsidianMinor:
Fixes #487

### DIFF
--- a/src/build/mod.rs
+++ b/src/build/mod.rs
@@ -13,6 +13,9 @@ pub fn build(target: &Target) -> Result<(), failure::Error> {
             message::info("JavaScript project found. Skipping unnecessary build!")
         }
         TargetType::Rust => {
+            let _ = which::which("rustc")
+                .map_err(|e| failure::format_err!("'rustc' not found: {}", e))?;
+
             let binary_path = install::install_wasm_pack()?;
             let args = ["build", "--target", "no-modules"];
 
--- a/src/build/mod.rs
+++ b/src/build/mod.rs
@@ -1,5 +1,5 @@
 use crate::settings::toml::{Target, TargetType};
-use crate::terminal::message;
+use crate::terminal::{message, styles};
 use crate::wranglerjs;
 use crate::{commands, install};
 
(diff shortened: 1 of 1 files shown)


src/build/mod.rs
```diff
@@ -13,6 +13,9 @@ pub fn build(target: &Target) -> Result<(), failure::Error> {
             message::info("JavaScript project found. Skipping unnecessary build!")
         }
         TargetType::Rust => {
+            let _ = which::which("rustc")
+                .map_err(|e| failure::format_err!("'rustc' not found: {}", e))?;
```
EverlastingBugstopper, line 17:
```rust
use wrangler::terminal::styles;
```

```rust
let rust_url = styles::url("https://www.rust-lang.org/tools/install");
```

```suggestion
                .map_err(|e| failure::format_err!("'rustc' not found, installation documentation can be found here: {}, rust_url))?;
```

EverlastingBugstopper: lets point the devs in the right direction for installation here just so it's easy for them to find.

ObsidianMinor: It might be good to still print the error message. Maybe:
```
                .map_err(|e| failure::format_err!(
"'rustc' not found: {}. Installation documentation can be found here: {}",
e, styles::url("https://www.rust-lang.org/tools/install")))?;
```

EverlastingBugstopper: what does the error message look like when it's printed out there? can you give an example?

ObsidianMinor: I believe it'd look something like:
`'rustc' not found: cannot find binary path. Installation documentation can be found here: https://www.rust-lang.org/tools/install`

Where "cannot find binary path" would be any of the messages from `which::Error`
```rust
#[derive(thiserror::Error, Copy, Clone, Eq, PartialEq, Debug)]
pub enum Error {
    #[error("bad absolute path")]
    BadAbsolutePath,
    #[error("bad relative path")]
    BadRelativePath,
    #[error("cannot find binary path")]
    CannotFindBinaryPath,
    #[error("cannot get current directory")]
    CannotGetCurrentDir,
    #[error("cannot canonicalize path")]
    CannotCanonicalize,
}
```

EverlastingBugstopper: 👍 sounds good to me

PR #1471, opened 2020-07-30
Wrangler Login

jspspike:
This is set up in staging but is ready for review.


.github/workflows/release.yml
```diff
@@ -89,6 +89,15 @@ jobs:
...
 
+      - name: Install OpenSSL via vcpkg (Windows)
+        if: matrix.build == 'windows'
+        run: |
+          git clone https://github.com/microsoft/vcpkg
+          .\vcpkg\bootstrap-vcpkg.bat
+          .\vcpkg\vcpkg install openssl:x64-windows
+        env:
+            VCPKGRS_DYNAMIC: 1
+
```
EverlastingBugstopper, line 100: mmmm - if we need an extra install step for windows then i dont think we should use openssl crate to do this. generally we want wrangler to be a standalone binary that does not require extra system configuration

Cargo.toml
```diff
@@ -67,9 +68,9 @@ fs_extra = "1.1.0"
 predicates = "1.0.5"
 
 [features]
-# We build Linux binaries with vendored openssl
-# to avoid SSL issues
-vendored-openssl = ['openssl/vendored']
```
EverlastingBugstopper, line 72: we should probably keep the `vendored-openssl` feature flag around so that if people have CI set up to install from cargo with that feature it doesn't just start failing

src/login/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+
+pub fn run() -> Result<(), failure::Error> {
+    let rsa = Rsa::generate(1024)?;
+    let pubkey = rsa.public_key_to_pem_pkcs1()?;
+
+    // Convert key to string and remove header and footer
+    let pubkey_str = str::from_utf8(&pubkey)?;
+    let pubkey_filtered = pubkey_str
+        .lines()
+        .filter(|line| !line.starts_with('-'))
```
EverlastingBugstopper, line 22: is it ever possible for an rsa key to have a `-` at the beginning of a line other than the `--BEGIN` and `--END` lines? also - is there a reason we dont just strip the first and last line of the string?

src/login/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+        .fold(String::new(), |mut data, line| {
+            data.push_str(&line);
+            data
+        });
+    let pubkey_encoded = percent_encode(pubkey_filtered.as_bytes(), NON_ALPHANUMERIC).to_string();
+
+    let browser_permission =
+        interactive::confirm("Allow Wrangler to a open page in your browser?")?;
```
EverlastingBugstopper, line 30:
```suggestion
        interactive::confirm("Allow Wrangler to open a page in your browser?")?;
```

src/login/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+    if !browser_permission {
+        failure::bail!("In order to use login you must allow Wrangler to open pages in your browser. If you don't want to do this consder using `wrangler config`");
```
EverlastingBugstopper, line 32:
```suggestion
        failure::bail!("In order to log in you must allow Wrangler to open your browser. If you don't want to do this consider using `wrangler config`");
```

src/login/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+fn poll_token(token_id: String) -> Result<String, failure::Error> {
+    let mut request_params = HashMap::new();
+    request_params.insert("token-id", token_id);
+
+    let client = reqwest::blocking::Client::new();
+    let timer = Timer::new().interval_ms(1000).iter();
+
+    let style = ProgressStyle::default_spinner().template("{spinner}   {msg}");
+    let spinner = ProgressBar::new_spinner().with_style(style);
+    spinner.set_message("Waiting to be sent api token...");
```
EverlastingBugstopper, line 70:
```suggestion
    spinner.set_message("Waiting for API token...");
```

src/login/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+            return Ok(body.result);
+        }
+
+        if seconds >= 500 {
+            break;
+        }
+    }
+
+    failure::bail!(
+        "Timed out while waiting for api token. Try using `wrangler config` if login fails to work."
```
EverlastingBugstopper, line 90:
```suggestion
        "Timed out while waiting for API token. Try using `wrangler config` if login fails to work."
```

src/main.rs
```diff
@@ -550,7 +550,7 @@ fn run() -> Result<(), failure::Error> {
         .subcommand(
             SubCommand::with_name("config")
                 .about(&*format!(
-                    "{} Set up wrangler with your Cloudflare account",
+                    "{} Authenticate Wrangler by copying in your Cloudflare API Token",
```
EverlastingBugstopper, line 553:
```suggestion
                    "{} Authenticate Wrangler with a Cloudflare API Token or Global API Key",
```

src/main.rs
```diff
@@ -615,6 +615,9 @@ fn run() -> Result<(), failure::Error> {
                 )
                 .arg(verbose_arg.clone())
         )
+        .subcommand(
+            SubCommand::with_name("login")
+                .about(&*format!("{} Authenticate Wrangler by logging in with your Cloudflare account", emoji::UNLOCKED)))
```
EverlastingBugstopper, line 620:
```suggestion
                .about(&*format!("{} Authenticate Wrangler with your Cloudflare username and password", emoji::UNLOCKED)))
```

src/login/mod.rs
jspspike: There shouldn't be a `-` at the beginning and I thought this was a good way to remove the first and last while also removing newlines. I changed `-` to `---` just to be safe

jspspike:
Won't merge this until service changes go into prod

EverlastingBugstopper:
will want to change the API its using to prod as well once that happens (unless you already did and i missed it)

jspspike:
> will want to change the API its using to prod as well once that happens (unless you already did and i missed it)

I haven't, will do

PR #1477, opened 2020-08-03
Added dev args to wrangler.toml

### LINKED ISSUE #1282: [dev] RFC: add dev configuration to wrangler.toml
# Situation

Currently, `wrangler dev` takes some options when you start it up, namely `--ip` and `--port`. We're talking about adding `--cors` or something similar (maybe combo of `--allowed-hosts` and `--allowed-headers`?), and I imagine there are further bells/whistles we will want to add as more folks adopt `wrangler dev` as the way to develop Workers.

# Problem

Right now there's no way for devs using `wrangler dev` to persist their development settings. If they want anything customized, they have to type out the arguments they use all the time and likely use `.bash_history` to remember what they do. This... isn't great. This problem gets even worse the more we extend the functionality of `wrangler dev`. 

# Solution

## `webpack-dev-server`

`webpack` has **TONS** of customizable options for `webpack-dev-server` which can be defined in `webpack.config.js`. Documentation for that can be found [here](https://webpack.js.org/configuration/dev-server/). A lot of the options they provide here can be defined in both `webpack.config.js` _AND_ as a command line argument. 

## `wrangler`

I don't think we need to worry about being quite as extensible as `webpack-dev-server`, but their idea of allowing configuration in a persistent file is quite appealing to me. I imagine folks working together on a single project would likely want to have the ability to customize their dev setup and commit it to version control - especially for things like `--allowed-hosts`.

I think to start I would allow anything that is currently configurable for dev (besides `--host` for [reasons](https://github.com/cloudflare/wrangler/pull/1085)) to be defined in `[dev]`. And for future features I'd allow any new extension to `dev` to be both a command line argument and configurable in `wrangler.toml`.

# Not Doing

Another thing to note here is that some developer-specific configurations should _not_ be committed to version control. The most common use case for this I can see is per-developer cloud credentials. Those concerns are separate and tracked [here](https://github.com/cloudflare/wrangler/issues/1229).

# Open Questions

- [ ] should we do this at all?
- [ ] should `[dev]` be configurable for different environments, or should it be top level only?

jspspike:
Fixes #1282

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -751,17 +751,35 @@ fn run() -> Result<(), failure::Error> {
         commands::preview(target, user, options, verbose)?;
     } else if let Some(matches) = matches.subcommand_matches("dev") {
         log::info!("Starting dev server");
-        let port: Option<u16> = matches
-            .value_of("port")
-            .map(|p| p.parse().expect("--port expects a number"));
-        let host = matches.value_of("host");
-        let ip = matches.value_of("ip");
+
         let config_path = Path::new(
             matches
                 .value_of("config")
                 .unwrap_or(commands::DEFAULT_CONFIG_PATH),
         );
         let manifest = settings::toml::Manifest::new(config_path)?;
+
+        let mut host: Option<&str> = matches.value_of("host");
+        let mut ip: Option<&str> = matches.value_of("ip");
+        let mut port: Option<u16> = matches
+            .value_of("port")
+            .map(|p| p.parse().expect("--port expects a number"));
+
+        // Check if arg not given but present in wrangler.toml
+        if let Some(d) = &manifest.dev {
+            if host.is_none() && d.host.is_some() {
+                host = d.host.as_deref();
+            }
+
+            if ip.is_none() && d.ip.is_some() {
+                ip = d.ip.as_deref();
+            }
+
+            if port.is_none() && d.port.is_some() {
+                port = d.port;
+            }
+        }
+
         let env = matches.value_of("env");
         let deploy_config = manifest.deploy_config(env)?;
         is_preview = true;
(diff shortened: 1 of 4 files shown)


src/settings/toml/dev.rs
```diff
@@ -0,0 +1,9 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct Dev {
+    pub host: Option<String>,
```
EverlastingBugstopper, line 6: im a bit wary of putting `host` in `wrangler.toml` because it's kinda half deprecated now because of edge dev...

src/main.rs
```diff
@@ -759,18 +759,14 @@ fn run() -> Result<(), failure::Error> {
...
 
         // Check if arg not given but present in wrangler.toml
         if let Some(d) = &manifest.dev {
-            if host.is_none() && d.host.is_some() {
-                host = d.host.as_deref();
-            }
-
             if ip.is_none() && d.ip.is_some() {
                 ip = d.ip.as_deref();
             }
```
EverlastingBugstopper: i'd also say that flags should override toml settings, so if they pass `--port 9090` and in their `wrangler.toml` it says `[dev.port] = 1010` then `9090` should be chosen

jspspike: That's how it should be working now, `ip` is an Option taken from clap. If it's none and there exists something in the toml then it changes ip, otherwise it'll be the same as before

EverlastingBugstopper:
doesn't seem to work?

---

**wrangler.toml**
```toml
name = "worker"
type = "javascript"
route = "averyharnish.com/test"

[dev]
ip = "[::1]"
port = "9000"
```
---

**console**
```console
$ wrangler dev
💁  watching "./"
```

PR #1485, opened 2020-08-06
Wrangler dev https server

### LINKED ISSUE #1204: [dev] HTTPS support

It would be very nice if `wrangler dev` server allows HTTPS access. I think the self-signed certificate is enough.

I maintain Cloudflare Workers to redirect to HTTPS URL on HTTP access on specific pages.
(The website I'm involved is a bit large and I cannot configure that with Page Rules since you can create only 100 Page Rules available.)

I tried to test HTTPS redirection with `wrangler dev`, but as expected, I opened "http://localhost:8787", then the Worker redirected to "https://localhost:8787" and the error page is shown.

@EverlastingBugstopper #1047

jspspike:
This will make the `wrangler dev` use https for the local server on gcs and edge. Edge can take in `--http` option to use http instead. Borrows some stuff from #1445 

Closes #1204


src/commands/dev/edge/server/https.rs
```diff
@@ -0,0 +1,105 @@
...
+            }
+        })
+        .boxed();
+    let server = Server::builder(tls::HyperAcceptor {
+        acceptor: incoming_tls_stream,
+    })
+    .serve(service);
+
+    println!("{} Listening on https://{}", emoji::EAR, listening_address);
+    message::info("Generated certifiacte is not verified, browsers will give a warning and curl will require `--inscure`");
```
EverlastingBugstopper, line 98:
```suggestion
    message::info("Generated certifiacte is not verified, browsers will give a warning and curl will require `--insecure`");
```

src/commands/dev/edge/server/https.rs
```diff
@@ -0,0 +1,105 @@
...
+                Ok(x) => x,
+                Err(e) => {
+                    println!("Failed to accept client");
+                    return Some(Err(e));
+                }
+            };
+            match tls_acceptor.accept(client).await {
+                Ok(x) => Some(Ok(x)),
+                Err(e) => {
+                    println!("Client connection error {}", e);
```
EverlastingBugstopper, line 85:
```suggestion
                    eprintln!("Client connection error {}", e);
```

EverlastingBugstopper: when i go to `https://127.0.0.1:8787` in my browser i get `ERR_SSL_KEY_USAGE_INCOMPATIBLE` instead of being able to click through a warning and i get this in my `dev` session:

```
[2020-08-10T17:39:41Z ERROR rustls::session] TLS alert received: Message {
        typ: Alert,
        version: TLSv1_3,
        payload: Alert(
            AlertMessagePayload {
                level: Fatal,
                description: IllegalParameter,
            },
        ),
    }
Client connection error received fatal alert: IllegalParameter
```

src/commands/dev/gcs/server.rs
```diff
@@ -85,15 +90,46 @@ pub(super) async fn serve(
...
+                Ok(x) => x,
+                Err(e) => {
+                    println!("Failed to accept client");
+                    return Some(Err(e));
+                }
+            };
+            match tls_acceptor.accept(client).await {
+                Ok(x) => Some(Ok(x)),
+                Err(e) => {
+                    println!("Client connection error {}", e);
```
EverlastingBugstopper, line 109:
```suggestion
                    eprintln!("Client connection error {}", e);
```

src/commands/dev/gcs/server.rs
```diff
@@ -85,15 +90,46 @@ pub(super) async fn serve(
...
+    })
+    .serve(service);
     println!(
-        "{} Listening on http://{}",
+        "{} Listening on https://{}",
         emoji::EAR,
         listening_address.to_string()
     );
+
+    message::info("Generated certificate is not verified, browsers will give a warning and curl will require `--inscure`");
```
EverlastingBugstopper, line 127:
```suggestion
    message::info("Generated certificate is not verified, browsers will give a warning and curl will require `--insecure`");
```

src/commands/dev/mod.rs
```diff
@@ -16,21 +17,24 @@ pub fn dev(
...
     match user {
         // authenticated users connect to the edge
-        Some(user) => edge::dev(target, user, server_config, deploy_config, verbose),
+        Some(user) => edge::dev(target, user, server_config, deploy_config, http, verbose),
 
         // unauthenticated users connect to gcs
-        None => gcs::dev(target, server_config, verbose),
+        None => {
+            if http {
+                failure::bail!("Unauthenticated dev must use https")
```
EverlastingBugstopper, line 34: i dont think this should be true - though we should figure out how `--host` and `--http` interact. my guess is that if the protocol mismatch in the two args then it should fail but other than that it should act the same as edge dev

src/commands/dev/tls/certs.rs
```diff
@@ -0,0 +1,88 @@
...
+    let home = get_wrangler_home_dir()?.join("config");
+    let cert = home.join("dev-cert.pem");
+    let privkey = home.join("dev-privkey.rsa");
+
+    if cert.exists() && privkey.exists() {
+        Ok(None)
+    } else {
+        fs::create_dir_all(&home)?;
+
+        message::info(format!("Generating certificate and private key for https server, if you would like to use your own you can replace `dev-cert.pem` and `dev-privkey.rsa` at {}", home.to_str().unwrap()).as_str());
```
EverlastingBugstopper, line 25: this makes sense to me - in the future we can also probably add something in `wrangler.toml` to override this location. this is more than good enough for now though

src/commands/dev/tls/certs.rs
```diff
@@ -0,0 +1,88 @@
...
+
+        Ok(Some((cert, privkey)))
+    }
+}
+
+/// Generate cert and private key for local https server
+pub fn generate_cert() -> Result<(), failure::Error> {
```
EverlastingBugstopper, line 32: this is super cool, i'd love if you could add a comment describing where you found the example or docs for why each of these settings are applied

src/commands/dev/tls/certs.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    let (cert_file, priv_file) = files.unwrap();
+
+    let rsa = Rsa::generate(2048)?;
+    let privkey = PKey::from_rsa(rsa)?;
+
+    let mut x509_name = X509NameBuilder::new()?;
+    x509_name.append_entry_by_text("C", "US")?;
+    x509_name.append_entry_by_text("ST", "TX")?;
```
EverlastingBugstopper, line 45: should we always be making TX certs for folks?

src/commands/dev/tls/certs.rs
```diff
@@ -0,0 +1,88 @@
...
+    x509_name.append_entry_by_text("O", "Example")?;
+    x509_name.append_entry_by_text("CN", "example")?;
```
EverlastingBugstopper, line 47: what's happening here?

src/commands/dev/edge/server/https.rs
jspspike: I didn't catch this earlier because firefox didn't check `KeyUsage` in the cert while chrome did. Should be fixed now

src/commands/dev/tls/certs.rs
jspspike: So this and the previous comment are the [attributes](https://docs.oracle.com/cd/E24191_01/common/tutorials/authz_cert_attributes.html) for the certificate. Usually they're the name of the company/organization for the certificate authority or who created the certificate. I was just using Wrangler or example.

src/commands/dev/tls/certs.rs
jspspike: https://github.com/cloudflare/wrangler/pull/1485#discussion_r468141988

src/commands/dev/edge/server/https.rs
koeninger: certifiacte / certificate

jspspike: Wow I'm really bad at spelling

jspspike:
So after talking about it with @EverlastingBugstopper we thought about how to structure https with dev. With dev there's the local server which accepts your requests and there's what you upstream to. We were thinking that making https default for both local and upstream would add a lot of friction to users when they run into browser warnings and curl failing without `--insecure`. Plus it seems most use cases won't use this. Instead making http default for both would have less friction and for other use cases we can add "local-https" and "upstream-https" cli argument (and `wrangler.toml` argument). Https will opt where those users can deal with the friction. This covers use cases where users rely on https behavior or if someone has a host that uses https but wants their local to be http.

kentonv:
> Instead making http default for both

To clarify... Are you saying that `wrangler dev` would by default connect to Cloudflare's edge via HTTP, not HTTPS?

I think that would be a dangerous default. We should be encrypting all network traffic unless the user very explicitly tells us not to.

Moreover, most sites these days enable "always use HTTPS" which causes all HTTP requests to return a redirect to HTTPS, which would likely lead to confusion when using `wrangler dev`.

I do agree that traffic to `localhost` should _not_ use HTTPS since there's no such thing as a valid certificate for `localhost`, and there's also no possibility of MITM.

jspspike:
> To clarify... Are you saying that `wrangler dev` would by default connect to Cloudflare's edge via HTTP, not HTTPS?
> 
> I think that would be a dangerous default. We should be encrypting all network traffic unless the user very explicitly tells us not to.

That's what I was intending but this makes sense. Local http and upstream https by default sound good? @EverlastingBugstopper @ashleymichal

EverlastingBugstopper:
the following configurations should be possible. when i say upstream i mean cloudflare's edge OR the --host specified for gcs dev

upstream https + local http (default)
upstream https + local https
upstream http + local http

---

upstream http + local https should not be possible i think

src/commands/dev/edge/mod.rs
```diff
@@ -19,7 +19,8 @@ pub fn dev(
     user: GlobalUser,
     server_config: ServerConfig,
     deploy_config: DeployConfig,
-    http: bool,
+    local_https: bool,
+    upstream_http: bool,
```
EverlastingBugstopper, line 23:
```suggestion
    local_protocol: Protcol,
    upstream_protocol: Protocol,
```

```
enum Protocol {
  Http,
  Https
}
```

src/commands/dev/mod.rs
```diff
@@ -18,23 +18,32 @@ pub fn dev(
...
-    http: bool,
+    local_https: bool,
+    upstream_http: bool,
     verbose: bool,
 ) -> Result<(), failure::Error> {
     // before serving requests we must first build the Worker
     build(&target)?;
 
+    if server_config.host.is_https() && upstream_http {
+        failure::bail!("Can't upstream http with https host")
```
EverlastingBugstopper, line 29:
```suggestion
        failure::bail("protocol mismatch: protocol in --host and protocol in --upstream-protocol must match")
```

src/commands/dev/mod.rs
```diff
@@ -18,23 +18,32 @@ pub fn dev(
...
+    } else if local_https && upstream_http {
+        failure::bail!("Can't have local https and upstreamed http")
```
EverlastingBugstopper, line 31:
```suggestion
        failure::bail("--local-protocol cannot be https if --upstream-protocol is http")
```

maybe suggest using console::style for the arguments here

src/commands/dev/server_config/mod.rs
```diff
@@ -15,6 +15,7 @@ impl ServerConfig {
         host: Option<&str>,
         ip: Option<&str>,
         port: Option<u16>,
+        upstream_http: bool,
```
EverlastingBugstopper, line 18: use the protocol enum i mentioned above

src/commands/dev/tls/certs.rs
```diff
@@ -132,6 +133,7 @@ pub fn generate_cert() -> Result<(), failure::Error> {
     cert_builder.append_extension(BasicConstraints::new().build()?)?;
 
     cert_builder.append_extension(
+        //Extensions requried by browsers to be a valid cert
```
EverlastingBugstopper, line 136:
```suggestion
        // Extensions requried by browsers to be a valid cert
```

src/main.rs
```diff
@@ -522,9 +522,15 @@ fn run() -> Result<(), failure::Error> {
                         .takes_value(true)
                 )
                 .arg(
-                    Arg::with_name("http")
-                    .help("Runs local server on http instead of https, only valid when running dev authenticated")
-                    .long("http")
+                    Arg::with_name("local-https")
+                    .help("Takes requests for local server via https instead of http")
```
EverlastingBugstopper, line 526: `--local-protocol http` or `--local-protocol https` -> need some parsing/validation to put in enum

src/commands/dev/edge/server/mod.rs
```diff
@@ -16,7 +17,7 @@ fn preview_request(
     client: HyperClient<HttpsConnector<HttpConnector>>,
     preview_token: String,
     host: String,
-    http: bool,
+    http: Protocol,
```
EverlastingBugstopper, line 20:
```suggestion
    protocol: Protocol,
```

src/commands/dev/tls/certs.rs
jspspike: #1497

PR #1498, opened 2020-08-12
Have --host run dev unauthenticated

### LINKED ISSUE #1494: Make `--host` in `wrangler dev` use unauthenticated instead of ignoring argument
### 💡 Feature request

Currently if you pass `--host` to `wrangler dev` while authenticated it will just ignore it since it makes sense to use the host provided in the `wrangler.toml`. Instead it should run unauthenticated dev using the provided host.

jspspike:
Closes #1494

### DIFF
--- a/src/commands/dev/mod.rs
+++ b/src/commands/dev/mod.rs
@@ -20,7 +20,6 @@ pub fn dev(
     deploy_config: DeployConfig,
     user: Option<GlobalUser>,
     server_config: ServerConfig,
-    host: bool,
     local_protocol: Protocol,
     upstream_protocol: Protocol,
     verbose: bool,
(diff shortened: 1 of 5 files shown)


src/commands/dev/mod.rs
```diff
@@ -40,19 +40,29 @@ pub fn dev(
...
-            local_protocol,
-            upstream_protocol,
-            verbose,
-        ),
-
-        // unauthenticated users connect to gcs
-        None => gcs::dev(target, server_config, local_protocol, verbose),
+    if let Some(user) = user {
+        if server_config.host.is_default() {
+            //Authenticated and no host provided, run on edge with user's zone
```
EverlastingBugstopper, line 45:
```suggestion
            // Authenticated and no host provided, run on edge with user's zone
```

PR #1507, opened 2020-08-17
Add output option for wrangler publish

nataliescottdavidson:
`wrangler build | publish --output json`

Wrangler supports structured output for core commands `build` and `publish`. `JSON` string goes to stdout, stderr displays user output.

js
```
nat@cf:~/examplejs$wrangler build
nat@cf:~/examplejs$wrangler build --output json
{"success":true,"name":"examplejs"}
nat@davidson:~/examplejs$ ../wrangler/target/debug/wrangler publish --output json

```
^ No output. Confused why that is happening as the logic to create structured output in the publish path should be the same regardless of project type.
Webpack `build` and `publish`
```
nat@davidson:~/demo-site$ ../wrangler/target/debug/wrangler build --output json
{"success":true,"name":"demo-site"}
nat@davidson:~/demo-site$ ../wrangler/target/debug/wrangler publish --output json
{"success":true,"name":"demo-site","urls":["https://demo-site.nats.workers.dev"]}

```

I'm unfamiliar with assert_cmd and will ramp up on that to make some solid tests.

### DIFF
--- a/src/commands/build.rs
+++ b/src/commands/build.rs
@@ -4,7 +4,7 @@ use clap::ArgMatches;
 
 use crate::build;
 use crate::settings::toml::Manifest;
-
+use crate::terminal;
 use super::DEFAULT_CONFIG_PATH;
 
 pub fn run(matches: &ArgMatches) -> Result<(), failure::Error> {
@@ -14,5 +14,16 @@ pub fn run(matches: &ArgMatches) -> Result<(), failure::Error> {
     let manifest = Manifest::new(&config_path)?;
     let env = matches.value_of("env");
     let target = &manifest.get_target(env, false)?;
+    if matches.is_present("output") {
+        if matches.value_of("output") == Some("json") {
+            terminal::message::set_output_type(terminal::message::OutputType::Json)
+        }
+        else {
+            terminal::message::user_error("json is the only valid value for output flag");
+        }
+    }
+    else {
+        terminal::message::set_output_type(terminal::message::OutputType::Human)
+    }
     build(&target)
 }
--- a/src/commands/build.rs
+++ b/src/commands/build.rs
@@ -17,8 +17,7 @@ pub fn run(matches: &ArgMatches) -> Result<(), failure::Error> {
     if matches.is_present("output") {
         if matches.value_of("output") == Some("json") {
             terminal::message::set_output_type(terminal::message::OutputType::Json)
-        }
-        else {
+        } else {
             terminal::message::user_error("json is the only valid value for output flag");
         }
         let result = build(&target);
@@ -26,14 +25,13 @@ pub fn run(matches: &ArgMatches) -> Result<(), failure::Error> {
             name: Some(target.name.clone()),
             success: match result {
                 Ok(_) => Some("true".to_string()),
-                Err(_) => Some("false".to_string())
+                Err(_) => Some("false".to_string()),
             },
             url: None,
         };
         terminal::message::jsonout(&jsonoutput);
         result
-    }
-    else {
+    } else {
         terminal::message::set_output_type(terminal::message::OutputType::Human);
         build(&target)
     }
--- a/src/commands/build.rs
+++ b/src/commands/build.rs
@@ -14,12 +14,8 @@ pub fn run(matches: &ArgMatches) -> Result<(), failure::Error> {
     let manifest = Manifest::new(&config_path)?;
     let env = matches.value_of("env");
     let target = &manifest.get_target(env, false)?;
-    if matches.is_present("output") {
-        if matches.value_of("output") == Some("json") {
-            terminal::message::set_output_type(terminal::message::OutputType::Json)
-        } else {
-            terminal::message::user_error("json is the only valid value for output flag");
-        }
+    if matches.is_present("output") && matches.value_of("output") == Some("json") {
+        terminal::message::set_output_type(terminal::message::OutputType::Json);
         let result = build(&target);
         let jsonoutput = terminal::message::PublishOutput {
             name: Some(target.name.clone()),
--- a/src/commands/build.rs
+++ b/src/commands/build.rs
@@ -21,9 +21,20 @@ pub fn run(matches: &ArgMatches) -> Result<(), failure::Error> {
         else {
             terminal::message::user_error("json is the only valid value for output flag");
         }
+        let result = build(&target);
+        let jsonoutput = terminal::message::PublishOutput {
+            name: Some(target.name.clone()),
+            success: match result {
+                Ok(_) => Some("true".to_string()),
+                Err(_) => Some("false".to_string())
+            },
+            url: None,
+        };
+        terminal::message::jsonout(&jsonoutput);
+        result
     }
     else {
-        terminal::message::set_output_type(terminal::message::OutputType::Human)
+        terminal::message::set_output_type(terminal::message::OutputType::Human);
+        build(&target)
     }
-    build(&target)
 }
--- a/src/commands/build.rs
+++ b/src/commands/build.rs
@@ -24,10 +24,10 @@ pub fn run(matches: &ArgMatches) -> Result<(), failure::Error> {
         let jsonoutput = terminal::message::PublishOutput {
             name: Some(target.name.clone()),
             success: match result {
-                Ok(_) => Some("true".to_string()),
-                Err(_) => Some("false".to_string()),
+                Ok(_) => Some(true),
+                Err(_) => Some(false),
             },
-            url: None,
+            urls: Vec::new(),
         };
         terminal::message::jsonout(&jsonoutput);
         result
--- a/src/commands/build.rs
+++ b/src/commands/build.rs
@@ -2,10 +2,10 @@ use std::path::Path;
 
 use clap::ArgMatches;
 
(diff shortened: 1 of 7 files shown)


EverlastingBugstopper:
Agreed, this should all be handled with the message framework, and if there are things we print that don't use that framework they should be migrated

nataliescottdavidson:
> If the main difference being messages go to stderr instead of stdout and the json output I think maybe using [lazy_static](https://docs.rs/lazy_static/) in `terminal::message` to have a static variable that would change all `message` functions to go to stderr instead of stdout would be better than having to pass a mutable reference into every function especially if we plan on adding this to other commands since this
> reference would basically have to passed everywhere. Thoughts? @EverlastingBugstopper @ashleymichal
> 
> Also looks like the lint CI failed, you can fix that by running `cargo fmt` and `cargo clippy`. Cool that you're starting work on Wrangler :)

Thanks for the feedback. It seems like I'll need to set a global variable `output_type`. That's usually a no-no, but I agree it's nicer than passing around a reference to every logging path.

jspspike:
>  It seems like I'll need to set a global variable

It doesn't really need to be global, just a local static variable in `terminal::message` and then in the `message` function check the variable and print out to stdout or stderr.

src/terminal/message.rs
```diff
@@ -1,9 +1,50 @@
...
+use serde::{Deserialize, Serialize};
+
+pub enum OutputType {
+    Json,
+    Human,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct PublishOutput {
+    pub success: Option<String>,
```
jspspike, line 14: I think you make this a bool and it'll serialize as true or false

src/commands/build.rs
```diff
@@ -14,5 +14,25 @@ pub fn run(matches: &ArgMatches) -> Result<(), failure::Error> {
     let manifest = Manifest::new(&config_path)?;
     let env = matches.value_of("env");
     let target = &manifest.get_target(env, false)?;
-    build(&target)
+    if matches.is_present("output") {
```
jspspike, line 17: Usually the clap matching stuff is done in main and then an `Option` with the arg is passed in where you can do this check and the one below. Something like
```rust
if let Some(output) = output {
  if output == "json" {
    ....
  }
}
```

nataliescottdavidson: That makes sense to keep the clap matching in main whenever possible. However, in this case the path for `wrangler build` leaves the matching to this `run` function. I'll do the matching in main and pass an `Option` for the `wrangler publish` path.

src/commands/publish.rs
```diff
@@ -61,8 +61,19 @@ pub fn publish(
...
 
-        deploy::worker(&user, &deploy_config)?;
-
+        if message::is_json_output() {
+            let result = build(&target);
+            let mut jsonoutput = message::PublishOutput {
+                name: Some(target.name.clone()),
+                success: match result {
+                    Ok(_) => Some(true),
+                    Err(_) => Some(true),
```
EverlastingBugstopper, line 70:
```suggestion
                    Err(_) => Some(false),
```

src/main.rs
```diff
@@ -440,6 +440,13 @@ fn run() -> Result<(), failure::Error> {
                         .long("env")
                         .takes_value(true)
                 )
+                .arg(
+                    Arg::with_name("output")
+                    .short("o")
+                    .long("output")
+                    .takes_value(true)
+                    .help("Machine or human readable output")
```
EverlastingBugstopper, line 448:
```suggestion
                    .help("possible values: <json>")
```

EverlastingBugstopper: or.. something like this. possible clap has some levers you can pull here

src/main.rs
```diff
@@ -558,6 +565,13 @@ fn run() -> Result<(), failure::Error> {
                         .long("release")
                         .takes_value(false)
                         .help("[deprecated] alias of wrangler publish")
+                )
+                .arg(
+                    Arg::with_name("output")
+                    .short("o")
+                    .long("output")
+                    .takes_value(true)
+                    .help("Machine or human readable output")
```
EverlastingBugstopper, line 574: same as above re: help message

src/terminal/message.rs
```diff
@@ -1,9 +1,55 @@
...
+    pub urls: Vec<String>,
+}
+
+static OUTPUT_TYPE: OnceCell<OutputType> = OnceCell::new();
+
+pub fn set_output_type(typ: OutputType) {
+    match OUTPUT_TYPE.set(typ) {
+        Ok(_) => {}
+        Err(_) => {
+            message(&"Output type already set".to_string());
```
EverlastingBugstopper, line 27: probably dont need to message here if this is the case, can do a log::info if you want though

src/commands/build.rs
```diff
@@ -14,5 +14,25 @@ pub fn run(matches: &ArgMatches) -> Result<(), failure::Error> {
...
+        if matches.value_of("output") == Some("json") {
+            terminal::message::set_output_type(terminal::message::OutputType::Json)
+        } else {
+            terminal::message::user_error("json is the only valid value for output flag");
```
EverlastingBugstopper, line 21: code maybe works as is? i'd double check this is off the top of my head..  - we want the flag to be highlighted yellow

```suggestion
            let flag = terminal::message::highlight("`--output`");
            terminal::message::user_error(&format!("{} must be json", flag));
```

src/main.rs
```diff
@@ -854,7 +868,15 @@ fn run() -> Result<(), failure::Error> {
         let env = matches.value_of("env");
         let mut target = manifest.get_target(env, is_preview)?;
         let deploy_config = manifest.deploy_config(env)?;
-
+        if matches.is_present("output") {
+            if matches.value_of("output") == Some("json") {
+                message::set_output_type(message::OutputType::Json)
+            } else {
+                message::user_error("json is the only valid value for output flag");
```
EverlastingBugstopper, line 875: same deal here with the terminal highlighting

src/terminal/message.rs
nataliescottdavidson: Would it be ok to just `.unwrap()` the value here? Not sure what the rust convention is when you don't care about a `Result`.

src/commands/publish.rs
nataliescottdavidson: that's a big oof right there- thanks for catching

src/commands/build.rs
nataliescottdavidson: I found the `clap` option for possible values to do this.

src/commands/publish.rs
EverlastingBugstopper: haha thats what code review friends are for!

src/terminal/message.rs
EverlastingBugstopper: no - unwrapping an error will cause it to panic. if you want to ignore you can do this

```rust
let _ = OUTPUT_TYPE.set(r#type);
```

src/terminal/message.rs
```diff
@@ -1,6 +1,19 @@
...
+#[derive(Serialize, Deserialize)]
+pub struct PublishOutput {
+    #[serde(skip_serializing)]
+    pub json : bool,
+    pub success : Option<String>,
+    pub name : Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub url : Option<String>,
+    #[serde(skip_serializing)]
+    pub humanoutput : Vec<String>
```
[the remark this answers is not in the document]
nataliescottdavidson: Implementation changed

src/terminal/message.rs
```diff
@@ -1,9 +1,50 @@
...
+    T: ?Sized + Serialize,
+{
+    println!("{}", &serde_json::to_string(value).unwrap());
+}
 
 fn message(msg: &str) {
-    println!("{}", msg);
+    match OUTPUT_TYPE.get() {
+        Some(OutputType::Json) => {
+            eprintln!("{}", msg);
```
EverlastingBugstopper, line 42: are these backwards? i think json should go to stdout and human readable should be stderr, no?

nataliescottdavidson: > json should go to stdout and human readable should be stderr, no?
Yes- IFF json output flag is given. I made the assumption messages are for human readable output. It made sense to me to make a json type here. I'm a rust novice so let me know if there is a better way. 
`wrangler build | publish --output json`

Wrangler supports structured output for core commands `build` and `publish`. `JSON` string goes to stdout, stderr displays user output.

js
```
nat@cf:~/examplejs$wrangler build
nat@cf:~/examplejs$wrangler build --output json
{"success":true,"name":"examplejs"}
nat@davidson:~/examplejs$ ../wrangler/target/debug/wrangler publish --output json `returns no output, need to fix.

```
Webpack `build` and `publish`
```
nat@davidson:~/demo-site$ ../wrangler/target/debug/wrangler build --output json
{"success":true,"name":"demo-site"}
nat@davidson:~/demo-site$ ../wrangler/target/debug/wrangler publish --output json
{"success":true,"name":"demo-site","urls":["https://demo-site.nats.workers.dev"]}

```
Next steps: extend to other worker types

nataliescottdavidson:
closing for #1538

PR #1512, opened 2020-08-20
Change redirects to host into localhost

jspspike:


src/commands/dev/utils.rs
```diff
@@ -1,8 +1,34 @@
...
+    if resp.status().is_redirection() {
+        let headers = resp.headers_mut();
+        if let Some(destination) = headers.get("Location") {
+            if let Ok(destination) = destination.to_str() {
+                if let Ok(destination_url) = Url::parse(destination) {
+                    if let Some(destination_domain) = destination_url.domain() {
+                        if destination_domain == upstream_host {
+                            let rewritten_destination =
+                                destination.replace(destination_domain, local_host);
+                            println!("rewriting to {:?}", &rewritten_destination);
```
EverlastingBugstopper, line 23:
```suggestion

```

EverlastingBugstopper:
very close - need to rewrite the protocol to match whatever is running localhost as well in order for this to work properly

Issue #1460, opened 2020-07-23
RFC: Machine-readable output for wrangler publish


codewithkristian:
# Situation

When running `wrangler publish`, the user receives the following console output:

```sh
~/src/evergreen/packages/edge$ wrangler publish
✨  Built successfully, built project size is 15 KiB.
🌀  Using namespace for Workers Site "__evergreen-workers_sites_assets"
✨  Success
🌀  Uploading site files
✨  Successfully published your script to https://REDEACTED
💁  Deleting stale files...
```

If something goes wrong, they receive an error string in console output:

```sh
~/src/evergreen/packages/edge$ wrangler publish
✨  Built successfully, built project size is 15 KiB.
Error: ⚠️  Code 10000: Authentication error
```

# Problem

For situations where projects are being deployed _automatically_, via [GitHub Actions](https://github.com/cloudflare/wrangler-action) or something similar, this output is buried inside of the console output. If we were able to _parse_ the information from `wrangler publish`, we would then be able to _pass_ it to other steps in a workflow. 

In https://github.com/cloudflare/wrangler-action/issues/9, I explored potentially parsing the above output via regex, but decided against it because, well, regex. That being said, the advantages of being able to consistently parse `wrangler publish` output is really exciting — we could capture the information and use it to send Slack messages (e.g. "$scriptName has been deployed to $url") or run e2e tests.

In an upcoming project I'm working on with @davidtsong @sgraden on, not having access to this functionality makes it effectively impossible to know _where_ a project has deployed to. Even if a project successfully deploys via GitHub Actions, we can't point them to a URL or even know the name of the script for construct a link to the script in the dashboard (`dash.cloudflare.com/$accountID/workers/$scriptName`).

# Solution

## `wrangler publish --json`

Adding a `--json` flag would return a parseable JSON string that could be passed into `jq` or other tools for getting more information about the published Worker. An example format is shown below:

```sh
$ wrangler publish --json | jq
{
  "success": true,
  "name": "cloudflare-worker-cra",
  "url": "http://cloudflare-worker-cra.signalnerve.workers.dev",
  "environment": "production"
}
```

I would LOVE to have access to `name` and `url` — if the project isn't published under an `environment`, it's possible that field would be confusing or unnecessary (maybe should just be omitted), and `success` is kind of a given if you're getting `name` and `url` back. `success` could be useful given the presence of an `error`:

```sh
$ wrangler publish --json | jq
{
  "success": false,
  "error": {
[...]

PR #1522, opened 2020-08-24
Refactor message module to message::StdOut and message::StdErr

nataliescottdavidson:
Prerequisite for https://github.com/cloudflare/wrangler/issues/1460
Issue surfaced here https://github.com/cloudflare/wrangler/pull/1507#issuecomment-678523080

### DIFF
--- a/src/build/mod.rs
+++ b/src/build/mod.rs
@@ -1,5 +1,6 @@
 use crate::settings::toml::{Target, TargetType};
-use crate::terminal::{message, styles};
+use crate::terminal::message::{Message, StdOut};
+use crate::terminal::styles;
 use crate::wranglerjs;
 use crate::{commands, install};
 
@@ -10,7 +11,7 @@ pub fn build(target: &Target) -> Result<(), failure::Error> {
     let target_type = &target.target_type;
     match target_type {
         TargetType::JavaScript => {
-            message::info("JavaScript project found. Skipping unnecessary build!")
+            StdOut::info("JavaScript project found. Skipping unnecessary build!")
         }
         TargetType::Rust => {
             let _ = which::which("rustc").map_err(|e| {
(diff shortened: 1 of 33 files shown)


src/terminal/message.rs
```diff
@@ -2,59 +2,86 @@ use super::emoji;
...
-        .build();
-    bb.display(msg);
+pub struct StdErr;
+
+impl Message for StdErr {
+    fn message(msg: &str) {
+        eprintln!("{}", msg);
+    }
+
+    fn billboard(_msg: &str) {
```
ashleymichal, line 80: this is unfortunate, esp since these are generally not program output. @EverlastingBugstopper any chance you'd be open to implementing a formatter only billboard? <3

src/build/mod.rs
```diff
@@ -1,5 +1,5 @@
 use crate::settings::toml::{Target, TargetType};
-use crate::terminal::{message, styles};
+use crate::terminal::{message, message::Message, styles};
```
ashleymichal, line 2: nit:
```suggestion
use crate::terminal::message::{Message, StdOut};
use crate::terminal::styles;
```
bums me out that you need the trait in scope 🤦‍♀️

src/terminal/message.rs
EverlastingBugstopper: Yes most definitely, I've been meaning to put some work into that crate

src/build/mod.rs
nataliescottdavidson: agreed, its unfortunate

src/terminal/message.rs
nataliescottdavidson: let's chat about this in wrangler contributors- i'd be willing to take a stab at it if you could explain what `formatter only` means here :)

PR #1528, opened 2020-08-25
Decouple `wrangler build` user output from internal build logic output

nataliescottdavidson:

### DIFF
--- a/src/build/mod.rs
+++ b/src/build/mod.rs
@@ -7,11 +7,14 @@ use crate::{commands, install};
 use std::path::PathBuf;
 use std::process::Command;
 
-pub fn build(target: &Target) -> Result<(), failure::Error> {
+// Internal build logic, called by both `build` and `publish`
+// TODO: return a struct containing optional build info and construct output at command layer
+pub fn build_target(target: &Target) -> Result<String, failure::Error> {
     let target_type = &target.target_type;
     match target_type {
         TargetType::JavaScript => {
-            StdOut::info("JavaScript project found. Skipping unnecessary build!")
+            let msg = "JavaScript project found. Skipping unnecessary build!".to_string();
+            Ok(msg)
         }
         TargetType::Rust => {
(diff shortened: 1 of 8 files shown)


src/build/mod.rs
```diff
@@ -7,11 +7,13 @@ use crate::{commands, install};
 use std::path::PathBuf;
 use std::process::Command;
 
-pub fn build(target: &Target) -> Result<(), failure::Error> {
+// This is called by both build and publish. Should return a struct of whatever info we need to show success
```
ashleymichal, line 10: should this be a struct or a string? i'd be into a struct, but i'm not 100% certain what it would contain.

nataliescottdavidson: Oops, left out a TODO. I think it will need to be a struct in the future once we have different information in the `build` output depending on project type. For now, its just going to be a string describing the result.

src/commands/build.rs
```diff
@@ -2,17 +2,29 @@ use std::path::Path;
...
     let env = matches.value_of("env");
     let target = &manifest.get_target(env, false)?;
-    build(&target)
+    let build_result = build_target(&target);
+    match build_result {
+        Ok(Some(msg)) => {
+            StdOut::success(&msg);
+            Ok(())
+        }
+        Ok(None) => {
```
ashleymichal, line 23: actually this match seems unnecessary. i think i would prefer that build target just return `Result<String>` rather than `Result<Option<String>>` and backfilling the none  as  it does here.

src/build/mod.rs
```diff
@@ -29,13 +32,19 @@ pub fn build(target: &Target) -> Result<(), failure::Error> {
             let command_name = format!("{:?}", command);
 
             commands::run(command, &command_name)?;
+            Ok(None)
```
ashleymichal, line 35: so i think that means moving the default message here

Issue #896, opened 2019-11-21
include asset manifest in worker site built script size


guaka:
### 🐛 Bug Report

```
 Uploading...
 Uploading...
 Success
⬇️ Installing wranglerjs...
⬇️ Installing wasm-pack...
 Built successfully, built project size is 11 KiB.
Error: Something went wrong! Status: 400 Bad Request, Details {
  "result": null,
  "success": false,
  "errors": [
    {
      "code": 10027,
      "message": "workers.api.error.script_too_large"
    }
  ],
  "messages": []
}
```

#### Environment

* ubuntu
* output of `rustc -V`: not found: rustc
* output of `node -v`: v10.17.0
* output of `wrangler -V`: wrangler 1.6.0

#### Steps to reproduce

`wrangler publish --verbose --env=prod`

on a directory with 60k+ files of size  4.2G.

#### What did you expect to see?

Success

#### What did you see instead?

`      "message": "workers.api.error.script_too_large"`

ashleymichal:
hey @guaka !

This issue reveals an interesting bug in how wrangler calculates build size. The workers platform calculates your project's bundle size based on the total size of the script and all of its embedded assets, including wasm modules and text blobs. The way we handle cache invalidation for Workers Sites is by generating an asset manifest that maps paths to kv keys, and the kv keys include a content hash to "fingerprint" them as unique cache keys. We bundle that asset manifest as a text blob binding when we upload your worker, and the size of that text blob counts against the 1MB bundle size limit.

I can imagine that a project with 60k+ assets could potentially come up against that limit.

There are a few ways we can try to alleviate this issue, but at minimum we should include the generated asset manifest in our bundle size calculations during a build so that our warn is still effective. we should also adjust the messaging to include the size of individual assets (script, wasm module, text blob) as an itemized set to show folks where their size issues are.

guaka:
Cool. That makes sense.

A related issue is that `Uploading...` takes quite a while. It would be nice if there was some kind of indication of what it's uploading or how many files have been uploaded so far.

(I also posted 2 other queries at https://community.cloudflare.com/u/fifi/activity)

tmikaeld:
This is an ongoing struggle for Cloudflare it seems, we hit a lot of API 500 errors today that came and went for a few hours. I'm wondering if they can handle it, considering that CF workers are growing in popularity.

sureshkumargondi:
[...]

PR #1543, opened 2020-09-03
Add support for text blob bindings

### LINKED ISSUE #483: Allow file bindings as text_blob

We have a worker that does a HTML transformation that adds a new JavaScript script tag, called client code. We want to store the client code along the server-side transformation.

The server-side and client-side codes need to keep sync, even during deployements, so I think KV wouldn't be a good fit.

Alternatively, we can use webpack to inline the client-side code into the worker. But allowing files as text_blob seems useful (a customer requested to inline images, we can use the same technique). 

#### Basic example

worker.js:
```js
})

  return new Response(CLIENT_SIDE_CODE);
}
```

metadata.json:
```json
{
  "body_part": "foo",
  "bindings": [
    {
      "name": "CLIENT_SIDE_CODE",
      "type": "text_blob",
      "part": "CLIENT_SIDE_CODE"
    },
  ]
}
```

Then the multipart from needs to include the `CLIENT_SIDE_CODE` part. Note that the size limit needs to take metadata.json into account too.

xortive:
closes #483 

I need to write docs, and confirm I got the env inheritance stuff right (I copied how it works for vars, need to ensure that is correct.)

But my baby test passes :)

I probably need to ensure size limits are still taken into account here as well.

changelog/docs:

Wrangler now supports text blobs! text blobs are values to use in your workers, but read from a file instead of a string in your TOML.

Usage:

text_blobs = { FOO = "path/to/foo.txt", BAR = "path/to/bar.txt" }


ashleymichal:
> I probably need to ensure size limits are still taken into account here as well.

correct. there's another ticket for that: #896

src/settings/toml/manifest.rs
```diff
@@ -43,6 +43,7 @@ pub struct Manifest {
     pub kv_namespaces: Option<Vec<ConfigKvNamespace>>,
     pub env: Option<HashMap<String, Environment>>,
     pub vars: Option<HashMap<String, String>>,
+    pub blobs: Option<HashMap<String, String>>,
```
ashleymichal, line 46: you should be able to use a `<HashMap<String, PathBuf>>` explicitly

src/settings/toml/manifest.rs
```diff
@@ -239,6 +241,9 @@ impl Manifest {
 
             // don't inherit vars
             target.vars = environment.vars.clone();
+
+            // don't inherit blobs
+            target.blobs = environment.blobs.clone();
```
ashleymichal, line 246: i don't remember why we chose not to inherit vars... 🤔

ashleymichal (commented):
i'm not sure about the name `blobs`. this likely deserves some product attention.

src/settings/toml/manifest.rs
xortive: I'd have used PathBuf, but webpack_config above uses String so I used that, but I can change it if you'd like.

ispivey:
@rita3ko : "Let's call 'em text blobs specifically, and just support those"

src/settings/toml/manifest.rs
ashleymichal: yes please 😄

src/settings/toml/manifest.rs
ashleymichal: i think these should likely be inherited; cc @rita3ko any thoughts?

ashleymichal (approved):

ashleymichal:
lgtm once the name change is complete.

src/settings/toml/manifest.rs
ashleymichal: these should be inherited i think.

xortive: My only reason for not inheriting these, is it feels weird to do something different than for vars.

ashleymichal:
@xortive are you in a position to address the above comments or should someone pick this up and carry it over the finish line?

xortive:
> @xortive are you in a position to address the above comments or should someone pick this up and carry it over the finish line?

Sorry I've been unresponsive with this PR, I've got some free time today so I'm going to get this out the door

src/settings/toml/manifest.rs
xortive: @ashleymichal @rita3ko Do y'all have a final opinion here? Y'all have more context than I do about what should/should not get inherited

xortive:
name change done, going to resolve conflicts now

xortive:
Alright, looks like this should be ready to go, unless we choose to inherit text_blobs.

xtuc (approved):

nataliescottdavidson (approved):
Could you write some docs / small lil paragraph explaining usage for the changelog?

PR #1551, opened 2020-09-09
Wrangling Actors

### LINKED ISSUE #1550: RFC: Durable Objects support
I propose a new per-environment, non-inherited key in wrangler.toml, called `actors-beta`.
This key will be a table with two properties. Actor namespaces have an `id`, similar to the id of a KV namespace.

- `uses`: array of tables
  - `binding` - same meaning as kv_namespaces
    - Then the names XOR ids this binding uses
    - fields for the name variant
      - `name` - name of the actors namespace
      - `preview_name` - name of the actors namespace to be used in preview
    - fields for the id variant
      - `id` - id of the actors namespace
      - `preview_id` - name of the actors namespace to be used in preview
  - `implements`: array of names
    - `name` - name of the actors namespace
      - If a name is provided here, wrangler will prompt to create the namespace for the user.
    - I'm not sure if an id possibility makes sense here, as there is a chicken/egg problem.
       It might be useful for silencing any warnings for a project with no routes or workers_dev = true,
       which would be the case for a project which only implements an actor.

Errors should be thrown if a namespace id or name in `uses` does not exist.

Supporting only IDs in bindings saves us an API request on deploy, but composes weirdly with the `implements` field which only makes sense for `name`.

All possible deployment scenarios supported by the runtime/platform, such as a worker implementing an actor namespace and having a route, should be supported by wrangler, but handling dependency loops when a script both uses and implements the same actor namespace is not in scope. However, we should warn in these scenarios, as there is a possibility for eyeball errors if not handled carefully by the user.

I chose to add a new key specific to actors, rather than a new project type, or a generic invocation_type field, to potentially support other invocation types in the future that may coexist in the same project.

Other considerations:
The actors api has new token scopes, so wrangler login will need to be done again.
Wrangler’s DeployConfig type does not support uploading scripts with no route. This used to not make sense, but it does now if you have a script which only purpose is to implement an actor namespace.
Wrangler's DeployConfig type has a semantic restriction that environments can only deploy to two invocation types (zoned or zoneless). It makes a lot of sense to have a single environment that both implements an actor namespace, and is invokable from workers.dev or a zoned route.

xortive:
Will implement #1550 
closes #1550

### DIFF
--- a/src/actors/mod.rs
+++ b/src/actors/mod.rs
@@ -1,0 +1,74 @@
+use crate::settings::toml::{ActorNamespace, ActorNamespaceNoId};
+use reqwest::blocking::Client;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct ApiActorNamespace {
+    pub name: String,
+    pub id: String,
+}
+
+#[derive(Deserialize)]
+pub struct ListActorNamespaceResponse {
+    pub result: Option<Vec<ApiActorNamespace>>,
+}
+
+pub fn list_namespaces_by_name_to_id(
+    client: &Client,
+    account_id: &str,
+) -> Result<HashMap<String, String>, failure::Error> {
+    let mut map = HashMap::new();
+    let namespaces = list_namespaces(client, account_id)?;
+    for namespace in namespaces {
+        map.insert(namespace.name, namespace.id);
+    }
+
+    Ok(map)
+}
+
+pub fn convert_namespace_names_to_ids(
+    client: &Client,
+    account_id: &str,
+    namespaces_noid: &[ActorNamespaceNoId],
+) -> Result<Vec<ActorNamespace>, failure::Error> {
+    let namespaces = list_namespaces_by_name_to_id(client, account_id)?;
+    namespaces_noid.iter().map(|ns| {
+        if let Some(id) = namespaces.get(&ns.name) {
+            Ok(ActorNamespace {
+                binding: ns.binding.clone(),
+                id: id.clone()
+            })
+        } else {
+            failure::bail!("actor namespace binding {} is configured with {}, a namespace that does not exist", ns.binding, ns.name);
+        }
+    }).collect()
+}
+
+fn list_namespaces(
+    client: &Client,
+    account_id: &str,
+) -> Result<Vec<ApiActorNamespace>, failure::Error> {
+    let list_addr = format!(
+        "https://api.cloudflare.com/client/v4/accounts/{}/workers/actors/namespaces",
+        account_id,
+    );
+
+    let res = client
+        .get(&list_addr)
+        .header("Content-type", "application/json")
+        .send()?;
+
+    if !res.status().is_success() {
+        failure::bail!(
+            "Something went wrong! Status: {}, Details {}",
+            res.status(),
+            res.text()?
+        )
+    }
+
+    Ok(res
+        .json::<ListActorNamespaceResponse>()?
+        .result
+        .unwrap_or_default())
+}
--- a/src/commands/dev/edge/setup.rs
+++ b/src/commands/dev/edge/setup.rs
@@ -41,7 +41,8 @@ pub(super) fn upload(
     let session_config = get_session_config(&deploy_config.http_routes);
     let address = get_upload_address(target);
 
-    let script_upload_form = upload::form::build(target, asset_manifest, Some(session_config))?;
+    let script_upload_form =
+        upload::form::build(target, Vec::new(), asset_manifest, Some(session_config))?;
 
     let response = client
         .post(&address)
--- a/src/commands/dev/edge/setup.rs
+++ b/src/commands/dev/edge/setup.rs
@@ -93,6 +93,9 @@ impl Session {
                 let subdomain = namespaces[1];
                 format!("{}.{}.workers.dev", target.name, subdomain)
             }
+            DeployConfig::NoRoutes => {
+                failure::bail!("wrangler dev does not support previewing scripts that are not reachable from a route or your workers.dev subdomain.");
+            }
         };
 
         let client = crate::http::legacy_auth_client(&user);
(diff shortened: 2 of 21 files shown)


src/settings/toml/deploy_config.rs
```diff
@@ -109,37 +116,31 @@ pub struct Zoned {
...
     fn has_conflicting_targets(&self) -> bool {
-        if self.is_zoneless() {
-            self.has_routes_defined()
-        } else if let Some(routes) = &self.routes {
-            !routes.is_empty() && self.route.is_some()
-        } else {
-            false
-        }
+        [
+            self.is_zoneless() && self.is_zoned(),
```
xortive, line 130: this case confused me a bit, since `is_zoneless && has_routes_defined` seemed like a less strict way of specifying `is_zoneless && is_zoned`.

I found that in `settings/toml/environments.rs`, the top level zone id was being copied into an environments InvocationConfig even if the environment was zoneless. I changed this behavior since it didn't make sense, but if I'm missing some reason that it exists, I can revert and add a comment explaining why.

src/settings/toml/tests/deploy_config.rs
```diff
@@ -12,17 +12,6 @@ const PATTERN: &str = "hostname.tld/*";
 const ACCOUNT_ID: &str = "fakeaccountid";
 
 // TOP LEVEL TESTS
-#[test]
-fn it_errors_on_empty_deploy_config() {
```
xortive, line 16: it might make sense to check the warnings in this case, instead of just deleting the tests. I need to figure out how, but I wanted to get this commit up earlier rather than later.

xortive:
basic functionality works, but there are easy to hit edge cases and absolutely no testing. I'm a bit unhappy with some of the threading through of configuration for setting up actor namespace bindings, but I think this is unavoidable.

src/commands/dev/edge/setup.rs
```diff
@@ -114,30 +128,32 @@ impl Session {
...
             "https://api.cloudflare.com/client/v4/zones/{}/workers/edge-preview",
             config.zone_id
         ),
         // TODO: zoneless is probably wrong
-        DeployConfig::Zoneless(config) => format!(
+        HttpRouteDeployConfig::Zoneless(config) => format!(
             "https://api.cloudflare.com/client/v4/accounts/{}/workers/subdomain/edge-preview",
             config.account_id
         ),
+        HttpRouteDeployConfig::NoRoutes => unimplemented!(),
```
koeninger, line 156: is there a possible implementation of either of these NoRoutes case?

xortive: not for wrangler dev, since currently it only supports workers that are invokable by a zone route or dotdev subdomain. the unimplemented! here is b/c these two will never be called, as we error out of wrangler dev in the NoRoutes case much earlier in the process.

src/actors/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+) -> Result<Vec<ActorNamespace>, failure::Error> {
+    let namespaces = get_existing_namespaces(client, account_id)?;
+    namespaces_noid.iter().map(|ns| {
+        if let Some(id) = namespaces.get(&ns.name) {
+            Ok(ActorNamespace {
+                binding: ns.binding.clone(),
+                id: id.clone()
+            })
+        } else {
+            failure::bail!("actor namespace binding {} is configured with {}, a namespace that does not exist", ns.binding, ns.name);
```
xortive, line 43: I'd love to give an error if the binding name here is also in the implements block... but that means making it accessible here :/

src/commands/dev/edge/setup.rs
```diff
@@ -38,10 +39,20 @@ pub(super) fn upload(
         (Vec::new(), None, None)
     };
 
-    let session_config = get_session_config(deploy_config);
+    let session_config = get_session_config(&deploy_config.http_routes);
     let address = get_upload_address(target);
 
-    let script_upload_form = upload::form::build(target, asset_manifest, Some(session_config))?;
+    let actor_namespaces = actors::convert_namespace_names_to_ids(
```
xortive, line 45: This is wrong in retrospect, as it tries to upload every implemented actor namespace when uploading to wrangler dev.

The edge-preview api does not consider them either, so we should probably remove this and pass an empty vec.

xortive:
some things that are missing currently, is being able to delete actor namespaces, and giving a better error when trying to delete a script bound to an actor namespace.

also, implementing an actor namespace that is currently bound to a different script, does not work.
We should allow this (with a warning that you are changing the implementing script).

xortive:
Implements now updates existing namespaces instead of doing nothing, and actor namespaces are not sent to wrangler dev's backend api since they are not supported there anymore.

I'm going to do the commands in their own PR.

src/settings/toml/tests/deploy_config.rs
xortive: Couldn't check for warnings, but I added the tests back into check for HttpRouteDeployConfig::NoRoutes instead.

src/settings/toml/manifest.rs
```diff
@@ -235,7 +245,13 @@ impl Manifest {
             }
 
             // don't inherit kv namespaces because it is an anti-pattern to use the same namespaces across multiple environments
-            target.kv_namespaces = get_namespaces(environment.kv_namespaces.clone(), preview)?;
+            target.kv_namespaces = get_kv_namespaces(environment.kv_namespaces.clone(), preview)?;
+
+            // don't inherit actor namespaces because it could be potentially dangerous to use the same namespaces across environments.
+            target.actor_namespaces = get_actor_namespaces(
```
xortive, line 251: based on the comments above... shouldn't site have an entry in this block too? It looks as if it will be inherited currently.

src/commands/dev/edge/setup.rs
koeninger: Yeah, it's a little unfortunate to turn what should be compile-time exhaustiveness checks into runtime errors, but I don't have a better suggestion ;)

ashleymichal:
closing this in favor of a different implementation to come later.

PR #1552, opened 2020-09-09
Fix the link to the update instructions

stof:
The old URL now triggers an error page.


xortive (approved):

xortive:
Thank you!

stof:
what would have been good is for old URLs to redirect to the appropriate place in the new website, but at least the readme now works.

PR #1566, opened 2020-09-21
Upload .well-known files

### LINKED ISSUE #980: Always upload `.well-known` dotfiles
Wrangler doesn't appear to publish hidden files. In particular, I'm attempting to publish a Hugo site with public/.well-known/* (I.e. .well-known/keybase.txt, .well-known/matrix/server, etc.) These files are definitely in public/, but don't appear in the key list for the created namespace.

nataliescottdavidson:
Hack to allow uploading .well-known files as requested by customers using wrangler for static site hosting- some frameworks use these dotfiles

closes #980

### DIFF
--- a/src/sites/mod.rs
+++ b/src/sites/mod.rs
@@ -68,23 +68,43 @@ pub fn add_namespace(
 pub fn directory_keys_values(
     target: &Target,
     directory: &Path,
-) -> Result<(Vec<KeyValuePair>, AssetManifest), failure::Error> {
+) -> Result<(Vec<KeyValuePair>, AssetManifest, Vec<String>), failure::Error> {
     match &fs::metadata(directory) {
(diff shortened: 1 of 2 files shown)


ashleymichal (approved):
whew. do you think you could write an RFC for replacing walkdir?

nataliescottdavidson:
> whew. do you think you could write an RFC for replacing walkdir?

Yep, will do.

nataliescottdavidson:
Right now we use walkdir to traverse the filesystem with an iterator, and for the include/exclude glob logic. Ideally the include/exclude logic would have regex. 
To replace walkdir, we would need to write our own directory iterator and include/exclude logic. I think that forking walkdir/include and modifying include to use regex makes the most sense.
For now, I think this hack is worth it to unblock users.

PR #1573, opened 2020-09-28
Allow site key to be configured by environment

### LINKED ISSUE #1567: Bug: Inconsistency for configuration inheritance
### 🐛 Bug Report

I want to deploy a specific folder for my site depending on the environment using wrangler **1.11.0**:
```toml
# wrangler.toml
[env.production.site]
bucket = "../app/dist-production"
entry-point = "."
```
However this configuration is ignored and the top-level site is always deployed.

-----

According to the [documentation](https://developers.cloudflare.com/workers/cli-wrangler/configuration#keys), the `site` key is  `not herited` (which is different from `Top level only`).
> Not inherited: Must be defined for every environment individually.

I took a look at the source code, and it seems that the code is not in sync with the doc (or the other way around :-), and treating this key as `top level only` (which does not help me either).

https://github.com/cloudflare/wrangler/blob/c76aab4b8fcf217b37e41313f136bab49bf1f016/src/settings/toml/manifest.rs#L228-L242

`target` is the top-level configuration, and only `name`, `account_id`, `webpack_config`, `kv_namespaces` and `vars` gets inherited (**`site` is missing**)

I guess it is just a matter of adding
```rust
     if let Some(site) = &environment.site { 
         target.site = Some(site.clone()); 
     } 
```

Do you mind considering adding a PR supporting this?

-----

Background information:
I want to deploy another site depending on the environment since the generation of my site depends on the environment as well (some config differs). As-is, there is a very high risk, that I deploy the staging code to my production URL (which would be avoided if I could specify another folder and generate my site in `dist-staging` and `dist-production`)

oliverpool:
Closes #1567


oliverpool:
@ashleymichal do you have any feedback on this PR ? (or an expected delay until it gets reviewed?)

nataliescottdavidson:
@oliverpool update and rebase and I will approve.

nataliescottdavidson (approved):

PR #1592, opened 2020-10-10
Add support for Cron Triggers

### LINKED ISSUE #1574: Mini RFC: Cron Triggers
Cron triggers are out, but there isn't a way to configure them from wrangler. To support cron triggers, I propose we add a new `triggers` table with a `crons` field as an array of strings.

### Example

```toml
name = "scheduled"
type = "javascript"
account_id = "..."

[triggers]
crons = ["* * * * *"]
```

```toml
name = "weekly"
type = "javascript"
account_id = "..."

[triggers]
crons = ["0 0 * JAN-JUN FRI", "0 0 * JUL-DEC SAT"]
```

ObsidianMinor:
* This also refactors deployments to accept multiple deploy targets so you
   aren't limited to zoned, zoneless, or scheduled for one script.

 * It also removes the constraint that only route or routes can be used.
   If both are specified, routes will have route inserted at the beginning

This fixes #1574


src/deploy/mod.rs
```diff
@@ -1,75 +1,38 @@
...
-            let addresses: Vec<String> =
-                published_routes.iter().map(|r| format!("{}", r)).collect();
+/// A set of deploy targets.
+pub type DeploymentSet = Vec<DeployTarget>;
 
-            Ok(addresses)
-        }
-    }
+#[derive(Debug, PartialEq, Clone)]
+pub enum DeployTarget {
```
ObsidianMinor, line 15: @ashleymichal So it turns out removing all type information by using a boxed trait object makes testing and checks in `wrangler dev` really hard so I decided to just use an enum instead to simplify a lot of stuff...

ashleymichal:
can you please address the clippy issue?

src/commands/publish.rs
```diff
@@ -25,11 +25,31 @@ pub struct PublishOutput {
...
     out: Output,
 ) -> Result<(), failure::Error> {
     validate_target_required_fields_present(target)?;
 
+    let deploy = |target: &Target| match deploy::worker(&user, &deployments) {
+        Ok(urls) => {
+            let result_msg = if !urls.is_empty() {
+                format!("Successfully published your script to {}", urls[0])
+            } else {
+                "Successfully published your script".to_owned()
```
ashleymichal, line 38: i assume this is the cron only response? it would be good for this output to include feedback on crons published as well, i think; if possible next approximate scheduled time or something similar. if that detail isn't possible, you can repeat back the cron schedule that was set.

to me this suggests that the deploy worker function result should perhaps become a struct with `urls` and `crons` or `schedules` to facilitate this, which would turn this into a kind of list, and in the case of json output, would add another field (`urls` and `schedules`)

src/deploy/mod.rs
```diff
@@ -1,75 +1,41 @@
...
-        .body(build_subdomain_request())
-        .send()?;
-
-    if !res.status().is_success() {
-        failure::bail!(
-            "Something went wrong! Status: {}, Details {}",
-            res.status(),
-            res.text()?
-        )
+    deployment: &[DeployTarget],
```
ashleymichal, line 23: nit: call this `deploy_targets`

src/deploy/mod.rs
```diff
@@ -1,75 +1,41 @@
...
+) -> Result<Vec<String>, failure::Error> {
```
ashleymichal, line 24: based on analysis from the publish command, we should change this result to a struct, `DeployResults` containing `urls` and `schedules`, each of which is a `Vec<String>`

src/deploy/zoned.rs
```diff
@@ -7,11 +7,70 @@ use cloudflare::framework::apiclient::ApiClient;
...
+    }
+
+    pub fn deploy(&self, user: &GlobalUser) -> Result<Vec<String>, failure::Error> {
+        log::info!("publishing to zone {}", self.zone_id);
+
+        let published_routes = publish_routes(&user, self)?;
+
+        let display_results: Vec<String> = published_routes.iter().map(|r| r.to_string()).collect();
+
+        StdOut::success(&format!(
```
ashleymichal, line 62: idk if this was here before, but stdout/stderr calls should pushed to the `command` layer to maintain control over output.

src/deploy/zoneless.rs
```diff
@@ -0,0 +1,69 @@
...
+#[derive(Clone, Debug, PartialEq)]
+pub struct ZonelessTarget {
+    pub account_id: String,
+    pub script_name: String,
+}
+
+impl ZonelessTarget {
+    pub fn build(script_name: &str, route_config: &RouteConfig) -> Result<Self, failure::Error> {
+        match route_config.account_id.as_ref() {
+            // TODO: Deserialize empty strings to None; cannot do this for account id
```
ashleymichal, line 16: good callout. we should make an issue referencing this line.

src/deploy/zoneless.rs
```diff
@@ -0,0 +1,69 @@
...
+            failure::bail!(
+                "Something went wrong! Status: {}, Details {}",
+                res.status(),
+                res.text()?
+            )
+        }
+
+        let deploy_address = format!("https://{}.{}.workers.dev", self.script_name, subdomain);
+
+        StdOut::success(&format!(
```
ashleymichal, line 58: same issue here with stdout. i suspect this simply hasn't happened yet, but since we're here let's get it.

src/settings/toml/environment.rs
```diff
@@ -32,17 +34,8 @@ impl Environment {
...
-            self.account_id.clone()
-        };
-
-        let zone_id = if self.zone_id.is_none() {
-            top_level_zone_id
-        } else {
-            self.zone_id.clone()
-        };
+        let account_id = self.account_id.clone().or(Some(top_level_account_id));
+        let zone_id = self.zone_id.clone().or(top_level_zone_id);
```
ashleymichal, line 38: nice tidy

src/settings/toml/manifest.rs
```diff
@@ -156,28 +163,86 @@ impl Manifest {
...
+                .map(|t| (t.crons.as_slice(), &self.account_id)),
+        };
+
+        if let Some((crons, account)) = crons {
+            let scheduled =
+                deploy::ScheduleTarget::build(account.clone(), script.clone(), crons.to_vec())?;
+            deployments.push(DeployTarget::Schedule(scheduled));
+        }
+
+        if !deployments.is_empty() {
```
ashleymichal, line 242: nit: flip this

src/settings/toml/tests/deployments.rs
```diff
@@ -13,18 +13,18 @@ const ACCOUNT_ID: &str = "fakeaccountid";
...
 
     let environment = None;
 
-    assert!(manifest.deploy_config(environment).is_err());
+    assert!(manifest.get_deployments(environment).is_err());
 }
 
 #[test]
-fn it_errors_on_conflicting_deploy_configs() {
+fn it_errors_on_conflicting_get_deploymentss() {
```
ashleymichal, line 27: update the name of the test to reflect expected behavior (we no longer expect this scenario to error, but rather to allow many kinds of deploy target).

src/settings/toml/tests/deployments.rs
```diff
@@ -94,58 +109,147 @@ fn it_errors_on_zoneless_deploy_config_workers_dev_false() {
...
+        account_id: ACCOUNT_ID.to_owned(),
+        script_name: worker_name,
+        crons: env_crons,
+    })];
+    let actual_deployments = manifest.get_deployments(environment).unwrap();
+
+    assert_eq!(actual_deployments, expected_deployments);
+}
+#[test]
+fn it_errors_on_no_schedules_in_env() {
```
ashleymichal, line 227: should this be true? i think when we talked about this the assumption was that crons should be able to inherit.

src/settings/toml/tests/deployments.rs
```diff
@@ -319,11 +423,11 @@ fn it_errors_on_multi_route_deploy_config_empty_route() {
 
     let environment = None;
 
-    assert!(manifest.deploy_config(environment).is_err());
+    assert!(manifest.get_deployments(environment).is_err());
 }
 
 #[test]
-fn it_errors_on_deploy_config_route_and_routes() {
+fn it_errors_on_get_deployments_route_and_routes() {
```
ashleymichal, line 430: rename to reflect current expectations

src/settings/toml/tests/deployments.rs
```diff
@@ -332,13 +436,27 @@ fn it_errors_on_deploy_config_route_and_routes() {
...
     let environment = None;
+    let actual_deployments = manifest.get_deployments(environment).unwrap();
 
-    assert!(manifest.deploy_config(environment).is_err());
+    assert_eq!(actual_deployments, expected_deployments);
 }
 
 #[test]
-fn it_errors_on_deploy_config_route_and_workers_dev_true() {
+fn it_errors_on_get_deployments_route_and_workers_dev_true() {
```
ashleymichal, line 459: i think this should now not error; this was part of relaxing restrictions.

src/settings/toml/tests/deployments.rs
```diff
@@ -348,11 +466,11 @@ fn it_errors_on_deploy_config_route_and_workers_dev_true() {
 
     let environment = None;
 
-    assert!(manifest.deploy_config(environment).is_err());
+    assert!(manifest.get_deployments(environment).is_err());
 }
 
 #[test]
-fn it_errors_on_deploy_config_routes_and_workers_dev_true() {
+fn it_errors_on_get_deployments_routes_and_workers_dev_true() {
```
ashleymichal, line 473: same here, this should no longer error

src/settings/toml/tests/deployments.rs
```diff
@@ -748,14 +866,14 @@ fn when_top_level_zoneless_env_zoned_multi_route_route_empty() {
...
+    let actual_deployments = manifest.get_deployments(Some(TEST_ENV_NAME));
 
-    assert!(actual_deploy_config.is_err());
+    assert!(actual_deployments.is_err());
 }
 
 #[test]
 fn when_top_level_zoneless_env_zoned_multi_route_route_key_present() {
-    // when route key also present, error
+    // when route key also present, append routes to route
```
ashleymichal, line 876: i think this case didn't get applied at the top level?

ashleymichal (changes requested):
for future PRs, it would be helpful to separate the refactors and feature work into different commits.

let's tidy up the output and tests. also have a look to see if there's changes needed to the readme.

src/deploy/zoneless.rs
ObsidianMinor: This was an existing comment, so an issue may already exist.

nataliescottdavidson:
I checked this out, added `crons = ["****"]` to my wrangler.toml, and published my worker. It doesn't show up in the cron triggers view on that worker in the dashboard, though. Am I getting the usage wrong?

ObsidianMinor:
@nataliescottdavidson It needs to be added under a `triggers` table like this:

```toml
name = "wrangled-scheduled"
type = "javascript"
account_id = "blah"

[triggers]
crons = ["* * * * *"]
```

src/deploy/zoneless.rs
ObsidianMinor: It may also not be a relevant comment anymore? Most strings use a `string_empty_as_none` deserializer now so it could just be outdated.

src/commands/publish.rs
```diff
@@ -20,16 +20,49 @@ pub struct PublishOutput {
...
+                ),
+            };
+            StdErr::success(&result_msg);
+            if out == Output::Json {
+                StdOut::as_json(&PublishOutput {
+                    success: true,
+                    name: target.name.clone(),
+                    urls,
+                    schedules,
+                });
```
ashleymichal, line 59: excellent

src/deploy/zoneless.rs
ashleymichal: i'll try to get through inline TODOs at some stage

ashleymichal (commented):
thanks mucho!

PR #1611, opened 2020-10-22
Add printing URL info in terminal::open_browser

### LINKED ISSUE #1544: Wrangler Login should print the URL to the console in addition to opening it
`wrangler login` opens a browser window to the Cloudflare dashboard in order to generate a token.

Some folks, like me, have multiple browser profiles. Opening a browser using the default mechanism doesn't necessarily use the correct profile.

Unfortunately, the URL that `wrangler login` opens seems to immediately redirect to something else, such that if I copy/paste the URL out of the address bar and into the correct browser profile, it doesn't work as expected.

Most tools like this also print the URL to the terminal, so that it can be copy/pasted. It would be great if Wrangler did so as well. It would be especially great if it did so even if I chose "no" when asked "Allow Wrangler to open a page in your browser?"

encadyma:
Resolves #1544 .

### DIFF
--- a/src/terminal/browser.rs
+++ b/src/terminal/browser.rs
@@ -1,3 +1,4 @@
+use crate::terminal::message::{Message, StdOut};
 use std::process::{Command, Stdio};
 
 pub fn open_browser(url: &str) -> Result<(), failure::Error> {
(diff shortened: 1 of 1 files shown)


nataliescottdavidson:
@encadyma, update and rebase your changes into one commit and I will approve.

encadyma:
Apologies for the messy force-push >_<. This branch should now be rebased with the upstream master.

nataliescottdavidson (approved):

PR #1631, opened 2020-11-01
Upload .well-known files using default ignore override

ObsidianMinor:
This removes all standard filters and replaces the faulty regex hack with a new default override that is used on top of any specified excludes.

Also this is getting more complex than it should be. Can we just use `.gitignore` or `.ignore` files on top of `include`/`exclude` and get rid of default "removed items" like node_modules and hidden files? Maybe just add a warning or add an option to preview a publish to see what would be uploaded?

### DIFF
--- a/src/sites/mod.rs
+++ b/src/sites/mod.rs
@@ -78,18 +78,6 @@ pub fn directory_keys_values(
             let spinner_style =
                 ProgressStyle::default_spinner().template("{spinner}   Preparing {msg}...");
             let spinner = ProgressBar::new_spinner().with_style(spinner_style);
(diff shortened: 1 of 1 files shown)


nataliescottdavidson:
@ObsidianMinor, thanks for fixing this.  Re `.gitignore`, docs for the ignore crate [claim it respects that by default](https://docs.rs/ignore/0.4.16/ignore/#example). I haven't tested to make sure that's the cause. 

Adding a `--dry-run` option to publish is a great idea. I'll make an issue.

nataliescottdavidson (approved):

PR #1649, opened 2020-11-17
Fix bug template "About" not to be too long

rita3ko:
I made the "about" too long which made the template go away :(


nataliescottdavidson (approved):

nataliescottdavidson:
@rita3ko oh, crap. That's on me for not properly reviewing your last docs change.

rita3ko:
@nataliescottdavidson it didn't give a very obvious error message, and it's pretty hard to preview the templates so no worries. But I do think the email will be better than the link so that's an improvement :)

PR #1680, opened 2020-12-13
Decorate the YAML workflow files with emoji

jbampton:


xtuc (changes requested):
this is not fixing an issue we have

xtuc:
Please raise an issue to discuss what we need to fix

PR #1717, opened 2021-01-11
list more than 100 kv namespaces

ags799:
Use the KV API's pagination to accumulate all the namespaces in an account.

### DIFF
--- a/src/kv/namespace/list.rs
+++ b/src/kv/namespace/list.rs
@@ -44,11 +44,10 @@ pub fn list(
 fn get_total(list_response: &ApiSuccess<Vec<WorkersKvNamespace>>) -> Result<usize, failure::Error> {
     match list_response.result_info.clone() {
         Some(r) => {
-            let result_info: ListResponseResultInfo =
-                serde_json::from_value(r)?;
+            let result_info: ListResponseResultInfo = serde_json::from_value(r)?;
             Ok(result_info.total_count)
-        },
-        None => failure::bail!("KV list response lacks result_info field")
+        }
+        None => failure::bail!("KV list response lacks result_info field"),
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/kv/namespace/upsert.rs
```diff
@@ -31,7 +31,6 @@ pub fn upsert(
                     log::info!("Namespace {} already exists.", title);
 
                     let namespace = list(&client, target)?
-                        .result
                         .iter()
                         .find(|ns| ns.title == title)
                         .unwrap()
```
koeninger, line 36: Is this the unwrap that was panicking originally?  Should we go ahead and fix it to return an error, even if it "shouldn't" happen anymore?

src/kv/namespace/list.rs
```diff
@@ -4,24 +4,50 @@ use cloudflare::endpoints::workerskv::list_namespaces::ListNamespaces;
...
+            }
+            Err(e) => failure::bail!("{}", kv::format_error(e)),
+        }
+    }
+    Ok(namespaces)
+}
+
+fn get_total(list_response: &ApiSuccess<Vec<WorkersKvNamespace>>) -> Result<usize, failure::Error> {
+    let result_info: ListResponseResultInfo =
+        serde_json::from_value(list_response.result_info.clone().unwrap())?;
```
koeninger, line 46: I'd like to get out of the habit of bare unwraps in the codebase, especially when this function is already returning a result.

src/kv/namespace/upsert.rs
ags799: done

src/kv/namespace/list.rs
ags799: done

koeninger (approved):

src/kv/namespace/upsert.rs
```diff
@@ -30,13 +30,12 @@ pub fn upsert(
...
-                    let namespace = list(&client, target)?
-                        .result
+                    match list(&client, target)?
                         .iter()
-                        .find(|ns| ns.title == title)
-                        .unwrap()
-                        .to_owned();
-                    Ok(UpsertedNamespace::Reused(namespace))
+                        .find(|ns| ns.title == title) {
+                        Some(namespace) => Ok(UpsertedNamespace::Reused(namespace.to_owned())),
```
xortive, line 36: :smile:

xortive (approved):

PR #1722, opened 2021-01-12
Fix filtering by extension for js assets

### LINKED ISSUE #1719: JSON files included in the worker script
### 🐛 Bug Report

JSON files are included in the worker script processed by the webpack.
This occurs in the `filterByExtension` function in `wranglerjs` because it evaluates to true with the `js` extension passed as an argument and matches for `js` and `json` files.

#### Environment

* operating system:
* output of `rustc -V`: 1.49.0
* output of `node -v`: 14.15
* output of `wrangler -V`: 1.12.3
* contents of `wrangler.toml`

```toml
account_id = "$CF_ACCOUNT_ID"
name = "example"
route = ""
type = "webpack"
webpack_config = "node_modules/flareact/webpack"
workers_dev = true
zone_id = ""

[site]
bucket = "out"
entry-point = "./"
```

#### Steps to reproduce

1. Create a worker site using the Flareact template: `wrangler generate my-project https://github.com/flareact/flareact-template`
2. Add a json file to the `public` folder to be served as a static asset
3. Build and deploy: `yarn deploy`

The content of the JSON file is not included in the worker's script file.

The worker script file includes the content of the JSON file.

EDIT:

Here's an incorrectly generated script file and the attempt to publish it:

Screenshot 2021-01-13 at 14 00 59
Screenshot 2021-01-13 at 14 01 10

rubenamorim:
Fixes #1719


tests/build.rs
```diff
@@ -331,7 +331,7 @@ fn it_builds_with_webpack_name_output() {
     build_creates_assets(&fixture, vec!["script.js"]);
 
     let out = fs::read_to_string(fixture.get_output_path().join("script.js")).unwrap();
-    assert!(out.contains(r#"//# sourceMappingURL=worker.js.map{"version":3,"file":"worker.js""#));
```
xtuc, line 334: why did you need to change the test?

rubenamorim: I think the output assertion is wrong because it contains a JSON file that is not parsed correctly as a JS file and throws an expected token error.

Here's an incorrectly generated script file and the attempt to publish it:

Screenshot 2021-01-13 at 14 00 59
Screenshot 2021-01-13 at 14 01 10

Btw, this is the first time I've dealt with Rust 🙈

Electroid: Yes, it looks like that test is wrong, based on what it should look like: https://developer.mozilla.org/en-US/docs/Tools/Debugger/How_to/Use_a_source_map

Electroid (approved):

PR #1730, opened 2021-01-14
remove webpack specific change

xtuc:
We used to hook directly into webpack internals to rewrite the runtime
that loads Wasm to make it point to the Worker binding instead of a
network fetch.

This change removes the webpack specific change and injects a generic
runtime to load Wasm from the binding.

Demo of the webpack's generated code: https://gist.github.com/xtuc/1a96d6793a3729016a518416ecf5b585#file-demo-js-L2-L9. I used https://github.com/xtuc/rust-webpack-worker-demo to test.

### DIFF
--- a/src/wranglerjs/bundle.rs
+++ b/src/wranglerjs/bundle.rs
@@ -42,6 +42,24 @@ impl Bundle {
             wasm_file.write_all(&wasm)?;
         }
 
+        if self.has_wasm() {
+            script_file.write_all(
+                format!(
+                    r#"
+                        WebAssembly.instantiateStreaming =
+                            async function instantiateStreaming(req, importObject) {{
+                          const module = {};
+                          return {{
+                            module,
+                            instance: new WebAssembly.Instance(module, importObject)
+                          }}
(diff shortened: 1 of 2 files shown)


src/wranglerjs/bundle.rs
```diff
@@ -42,6 +42,22 @@ impl Bundle {
             wasm_file.write_all(&wasm)?;
         }
 
+        script_file.write(
+            format!(
+                r#"
+                WebAssembly.instantiateStreaming =
```
xortive, line 48: this is a neat trick! can we provide a real value for module or is it not possible?

xortive: edit: didn't notice the format!, disregard

xtuc: sorry what do you mean by real value?

xortive (approved):
Tested with provided template, works great :)

xtuc:
changed to emit the JS runtime only if wasm is actually used. Avoids breaking source maps for others

PR #1740, opened 2021-01-20
kv put with metadata

### LINKED ISSUE #1734: RFC: KV put with metadata
### 💡 Feature request

Perform a [KV put with metadata](https://developers.cloudflare.com/workers/runtime-apis/kv#metadata) from Wrangler:
```
wrangler kv:key put my-key my-value --metadata '{"some": "json"}'
```
Any kind of JSON works:
* boolean
* number
* object
* array
* strings

ags799:
Closes #1734

### DIFF
--- a/src/commands/kv/key/put.rs
+++ b/src/commands/kv/key/put.rs
@@ -13,9 +13,9 @@ use crate::http;
 use crate::settings::global_user::GlobalUser;
 use crate::settings::toml::Target;
 use crate::terminal::message::{Message, StdOut};
+use regex::Regex;
 use reqwest::blocking::multipart;
 use reqwest::blocking::Body;
-use regex::Regex;
 
 pub struct KVMetaData {
     pub namespace_id: String,
@@ -37,10 +37,13 @@ pub fn parse_metadata(arg: Option<&str>) -> Result<Option<serde_json::Value>, fa
                     // try to help users that forget to double-quote a JSON string
                     let re = Regex::new(r#"^['"]?[^"'{}\[\]]*['"]?$"#)?;
                     if re.is_match(s) {
-                        failure::bail!("did you remember to double quote strings, like --metadata '\"{}\"'", s)
+                        failure::bail!(
+                            "did you remember to double quote strings, like --metadata '\"{}\"'",
+                            s
+                        )
                     }
                     failure::bail!(e.to_string())
-                },
+                }
             }
         }
     }
@@ -167,9 +170,11 @@ mod tests {
                 Ok(_) => return Err("illegal value was parsed successfully"),
                 Err(e) => {
                     let expected_message = format!(
-                        "did you remember to double quote strings, like --metadata '\"{}\"'", input);
+                        "did you remember to double quote strings, like --metadata '\"{}\"'",
+                        input
+                    );
                     assert_eq!(expected_message, e.to_string());
-                },
+                }
             }
         }
         Ok(())
--- a/src/commands/kv/key/put.rs
+++ b/src/commands/kv/key/put.rs
@@ -37,8 +37,7 @@ pub fn parse_metadata(arg: Option<&str>) -> Result<Option<serde_json::Value>, fa
                     let re = Regex::new(r#"^['"]?[^"'{}\[\]]*['"]?$"#)?;
                     if re.is_match(s) {
                         failure::bail!(
-                            "did you remember to double quote strings, like --metadata '\"{}\"'",
-                            s
+                            "did you remember to double quote strings, like --metadata '\"made with 🤠 wrangler\"'"
                         )
                     }
                     failure::bail!(e.to_string())
--- a/src/commands/kv/key/put.rs
+++ b/src/commands/kv/key/put.rs
@@ -15,7 +15,6 @@ use crate::settings::toml::Target;
 use crate::terminal::message::{Message, StdOut};
 use regex::Regex;
 use reqwest::blocking::multipart;
-use reqwest::blocking::Body;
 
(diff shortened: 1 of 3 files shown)


src/commands/kv/key/put.rs
```diff
@@ -82,3 +72,78 @@ pub fn put(target: &Target, user: &GlobalUser, data: KVMetaData) -> Result<(), f
...
+fn get_response(
+    data: KVMetaData,
+    user: &GlobalUser,
+    url: &Url,
+) -> Result<reqwest::blocking::Response, failure::Error> {
+    let url_into_str = url.to_string();
+    let client = http::legacy_auth_client(user);
+    let res = match data.metadata {
+        Some(metadata) => {
+            let part = if data.is_file {
```
koeninger, line 85: data.is_file is being handled differently from get_request_body, which means people are going to get confusingly different errors for invalid files depending on whether they added metadata or not.

src/commands/kv/key/put.rs
```diff
@@ -13,13 +13,24 @@ use crate::http;
...
     pub expiration: Option<String>,
     pub expiration_ttl: Option<String>,
+    pub metadata: Option<serde_json::Value>,
+}
+
+// to be used with clap's Arg::validator
+pub fn metadata_validator(arg: String) -> Result<(), String> {
+    serde_json::from_str::<serde_json::Value>(&arg)
+        .map(|_| ())
+        .map_err(|e| e.to_string())
```
koeninger, line 33: The error serde gives isn't very helpful.  It's going to be especially confusing because as things stand

    --metadata '"somestring"'

will work, but not
    
     --metadata "'somestring'"
     --metadata "somestring"

If we cant fix that, we should at least make the error more helpful

src/main.rs
```diff
@@ -1045,13 +1054,18 @@ fn run() -> Result<(), failure::Error> {
                 let expiration_ttl = put_key_matches
                     .value_of("expiration-ttl")
                     .map(|t| t.to_string());
+                let metadata = put_key_matches
+                    .value_of("metadata")
+                    .map(serde_json::from_str)
+                    .map(Result::unwrap); // success is guaranteed in the argument's `validator` call
```
koeninger, line 1060: imho the interface for validator is bad (it's throwing away parsed information that you're going to need anyway)... but that shouldn't excuse using bare unwraps

src/commands/kv/key/put.rs
```diff
@@ -82,3 +72,78 @@ pub fn put(target: &Target, user: &GlobalUser, data: KVMetaData) -> Result<(), f
...
+            Ok(file_type) if file_type.is_file() => {
+                let file = fs::File::open(&data.value)?;
+                Ok(file.into())
+            }
+            Ok(file_type) if file_type.is_dir() => failure::bail!(
+                "--path argument takes a file, {} is a directory",
+                data.value
+            ),
+            // last remaining value is symlink
+            Ok(_) => failure::bail!("--path argument takes a file, {} is a symlink", data.value),
```
koeninger, line 117: I'm a little confused as to why this prior code was calling out symlinks... I'd expect it to work, and I read this api doc https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.is_file  as suggesting you should try opening the file rather than explicitly testing if it's a regular file.

src/commands/kv/key/put.rs
ags799: Okay. We cannot use a multipart upload without metadata. So we can:
1. Edit `get_request_body` to always return a `String`, so errors like "file not found" are always in the same format.
2. Massage all errors from `multipart::Part::file` and `std::fs::File::open` to look the same.
3. Edit the API to allow for multipart uploads without metadata, so this code will always use `multipart::Part::file`.

I think 2 is too much work. What do you think between 1 & 3?

src/commands/kv/key/put.rs
ags799: I just tried it on master, symlinks work. I can replace the error text to say
```
--path argument points to unknown entity
```

src/commands/kv/key/put.rs
koeninger: I see what you're saying.
I don't think we want to change the api.
Part is just a wrapper around Body, so it seems like you should be able to use get_request_body as is, but the constructor for Part that takes a Body is private >_<
Regarding using String as lowest common denominator instead of Vec<u8>, what does the current implementation do with a file of arbitrary bytes?

src/commands/kv/key/put.rs
ags799: I added something to help out here. I think it's good to require the double quotes so that we maintain a sensible input space.

src/main.rs
ags799: I removed the validator and the bare unwrap.

src/commands/kv/key/put.rs
ags799: done

src/commands/kv/key/put.rs
ags799: Good point, we need to respect arbitrary bytes. I used `Vec<u8>`.

src/commands/kv/key/put.rs
koeninger: Yeah, totally.

I think it's probably better to use a fixed valid example string in the error message, rather than their input, because adding quotes to their input won't necessarily make it valid.

Otherwise LGTM thanks

ags799: done

koeninger (approved):

xortive (approved):

Issue #1102, opened 2020-02-21
Clarity on the respective purposes of build and publish


JamesHenry:
#### Overview and problem statement

I guess this would likely fall into a request for documentation updates...

I just went to set up a CI/CD pipeline for an experimental Cloudflare Worker I am working on and wrangler did not behave how I intuitively expected it to.

I set up a Build stage to my pipeline in which I ran `wrangler build`, then published the built artifact for consumption in my Deploy stage. Pretty standard stuff I think.

In my Deploy stage I set up `wrangler publish` and expected it to be responsible for taking the result of `wrangler build` (a nice lightweight artifact) and publishing it to my Cloudflare account.

Instead it seems like `publish` essentially needs the full repo source and npm installation all over again, which suggests it is not intended to be combined with `build` at all, which has left me confused as to the design/intended use-cases of the commands.

Would love to hear your thoughts on this setup and intuition, maybe I have just messed up my approach.

Thanks!

victoriabernard92:
I agree with this report. It is unclear and frustrating to have to build every time on publish. JS developers often know what steps to get a project built and prefer to have control of that.  I think rather than documenting this we should consider solving the problem at the design of the CLI level. 

Possible solutions:
- `publish --skip-build`. Which would simply publish to CF but do no building. Radical, but if this flag gets popular consider sunsetting the old `publish` to skip builds by default
- In`wrangler.toml`, have `type='javascript'` not install wasm-pack or build anything it simply talks to CF's API

JamesHenry:
@victoriabernard92 thanks for acknowledging - do you help run the project? Would you be willing to accept PRs to implement these proposed solutions?

victoriabernard92:
@JamesHenry while we do accept PRs this proposal needs design from and buy in from our team. I am looking forward to hearing their suggestions and getting back to you.

JamesHenry:
@victoriabernard92 Are there individuals we can ping to make this a little more concrete?

Again, I am happy to do all the work here, I just need the go ahead

JamesHenry:
This should remain open please

JamesHenry:
This should remain open please.

@ashleymichal - is there a way I can avoid this now being my job to keep this open? 😅 

This seems to have been acknowledged by the team as something that could be improved and will be addressed?

ashleymichal:
[...]

PR #1741, opened 2021-01-20
Add 'deploy' alias for 'publish'

### LINKED ISSUE #1739: Add `wrangler deploy`

Brief explanation of the requested feature, and a description of the problem it would solve.

Many people  — including myself — have found `wrangler publish` to be an unintuitive way to deploy scripts. `wrangler deploy` is much more intuitive. Adding this as an alias or making that command print something out like `did you mean wrangler publish?` would be an easy improvement to make usability a little better.

ObsidianMinor:
I've always wanted this tbh.

Closes #1739

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -539,6 +539,7 @@ fn run() -> Result<(), failure::Error> {
         )
         .subcommand(
             SubCommand::with_name("publish")
+                .visible_alias("deploy")
                 .about(&*format!(
                     "{} Publish your worker to the orange cloud",
                     emoji::UP


ObsidianMinor:
rebased

xtuc (approved):
I agree that it's a better name for it

xtuc:
Would you mind changing the README and adding a note to the changelog?

Electroid:
Do we want to do this quite yet? `publish` does a build + deploy, but there have been talks previously to have `deploy` not build since it's not-obvious that it re-builds. If we alias this now, it will be complicated to fix later.

xortive:
I agree with @Electroid, #1102 has some previous discussion here, and with the custom builds/modules PR landing soon I think adding a deploy or upload command that doesn't build makes a lot of sense.

Another thing to think about is "publish" really encompasses 3 steps currently, not 2:

- building your project into a workers script
- uploading the script to the API
- configuring how your script can be run such as routes, cron triggers, or implementing durable object namespace

These three steps to me would have commands like `wrangler build`, `wrangler upload`, and `wrangler deploy`, with `wrangler publish` being the shortcut version for when you want all three.

ObsidianMinor:
Why offer a separate command to skip builds instead of a CLI option like `--skip-build`. The CLI option is more obvious since `deploy` and `publish` are effectively synonyms.

Electroid:
We need to support `deploy`, but for the above reasons, I think we'll need more design before adding the alias.

PR #1745, opened 2021-01-25
use toml-edit for modifying provided wrangler.toml for wrangler generate to preserve formatting

xortive:
This came up when testing out #1677, `wrangler generate`'s outputted toml for the testing durable objects `wrangler.toml` is rather ugly. Keeping a nice, understandable format for `wrangler.toml` is an important part of reducing user confusion, and I think it's worth the small risk of having two toml parsers in wrangler.

example output from this pr vs current implementation coming:

### DIFF
--- a/src/settings/toml/manifest.rs
+++ b/src/settings/toml/manifest.rs
@@ -85,60 +85,91 @@ impl Manifest {
         config_path: &PathBuf,
         site: Option<Site>,
     ) -> Result<Manifest, failure::Error> {
-        let config_file = config_path.join("wrangler.toml");
-        let template_config_content = fs::read_to_string(&config_file);
-        let template_config = match &template_config_content {
-            Ok(content) => {
-                let config: Manifest = toml::from_str(content)?;
-                config.warn_on_account_info();
-                if let Some(target_type) = &target_type {
-                    if config.target_type != *target_type {
-                        StdOut::warn(&format!("The template recommends the \"{}\" type. Using type \"{}\" may cause errors, we recommend changing the type field in wrangler.toml to \"{}\"", config.target_type, target_type, config.target_type));
-                    }
-                }
-                Ok(config)
-            }
-            Err(err) => Err(err),
-        };
-        let mut template_config = match template_config {
-            Ok(config) => config,
-            Err(err) => {
-                log::info!("Error parsing template {}", err);
-                log::debug!("template content {:?}", template_config_content);
+        let config_file = &config_path.join("wrangler.toml");
+        let config_template_str = fs::read_to_string(config_file).unwrap_or_else(|err| {
+            log::info!("Error reading config template: {}", err);
+            log::info!("Using default instead");
+            toml::to_string_pretty(&Manifest::default())
+                .expect("serializing the default toml should never fail")
+        });
+
+        let config_template =
+            toml::from_str::<Manifest>(&config_template_str).unwrap_or_else(|err| {
+                log::info!("Error parsing config template: {}", err);
+                log::info!("Using default instead");
                 Manifest::default()
-            }
-        };
+            });
 
-        let default_workers_dev = match &template_config.route {
-            Some(route) => {
-                if route.is_empty() {
-                    Some(true)
-                } else {
-                    None
-                }
-            }
-            None => Some(true),
-        };
-
-        template_config.name = name;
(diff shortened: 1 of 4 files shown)


xortive:
what wrangler generate on #1677 outputs for https://github.com/xortive/durable-objects-template

```
name = "current"
type = "javascript"
account_id = ""
workers_dev = true
route = ""
zone_id = ""

[builder_config]
build_command = "npm install && npm run build"
build_dir = "dist"
upload_format = "modules"
src_dir = "src"
[[durable_objects.uses]]
binding = "Counter"
namespace_name = "Counter"

[[durable_objects.implements]]
namespace_name = "Counter"
class_name = "Counter"
```

what this PR outputs:
```
name = "toml-edit"
type = "javascript"
account_id = ""
workers_dev = true
route = ""
zone_id = ""

[builder_config]
build_command = "npm install && npm run build"
upload_format = "modules"
src_dir = "src"
build_dir = "dist"

[durable_objects]
implements = [{class_name = "Counter", namespace_name = "Counter"}]
uses = [{binding = "Counter", namespace_name = "Counter"}]
```

xortive:
This PR needs to be tested with our various templates as well as workers sites, to ensure we aren't now making a confusing toml.

src/settings/toml/manifest.rs
```diff
@@ -85,60 +85,91 @@ impl Manifest {
...
+                        arr.push(i.as_ref()).unwrap();
+                    });
+                    config_template_doc["target_type"] = toml_edit::value(arr);
+                }
+                if let Some(exclude) = &site.exclude {
+                    let mut arr = toml_edit::Array::default();
+                    exclude.iter().for_each(|i| {
+                        arr.push(i.as_ref()).unwrap();
+                    });
+                    config_template_doc["target_type"] = toml_edit::value(arr);
```
ags799, line 160: The last 3 values are all `target_type`?

src/settings/toml/manifest.rs
```diff
@@ -85,60 +85,91 @@ impl Manifest {
...
+
+                if let Some(entry_point) = &site.entry_point {
+                    config_template_doc["target_type"] =
+                        toml_edit::value(entry_point.to_string_lossy().as_ref());
+                }
+                if let Some(include) = &site.include {
+                    let mut arr = toml_edit::Array::default();
+                    include.iter().for_each(|i| {
+                        arr.push(i.as_ref()).unwrap();
+                    });
```
ags799, line 152: I think you can do this in one line:
```rust
let arr = toml_edit::Array::from(include.iter().map(String::as_ref).collect());
```
Same for `exclude` below.

src/settings/toml/manifest.rs
xortive: bah, I did a copy pasta. Haven't tested with sites yet so didn;t notice

src/settings/toml/site.rs
```diff
@@ -13,7 +13,7 @@ const SITE_ENTRY_POINT: &str = "workers-site";
 pub struct Site {
     pub bucket: PathBuf,
     #[serde(rename = "entry-point")]
-    entry_point: Option<PathBuf>,
+    pub entry_point: Option<PathBuf>,
```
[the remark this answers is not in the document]
xortive: need to directly access entry_point in manifest.rs now. Before, we'd copy the entire Site struct into the generated Manifest struct, but toml_edit won't let us do that so we need to set each field by hand :/

src/settings/toml/manifest.rs
xortive: There's no `From<_>` impl that will make this work unfortunately.

xortive:
Tested `wrangler generate --site`

```
account_id = ""
name = "site-test"
type = "webpack"
workers_dev = true
site = { bucket = "./public" }
```

Outputs inline table instead of separate table currently. This is equivalent to the existing output, and satisfies a TODO comment in `manifest.rs` as a better solution to #798

xortive:
took a look at the templates linked to from the docs and none of them are doing anything unique in their wrangler.toml except for the sites one and the scala KV one, and both of those outputted good tomls when testing.

ags799 (approved):

xortive:
rerunning tests since windows test flaked on preview

nataliescottdavidson (approved):

Issue #820, opened 2019-10-29
Allow users to customize `wrangler build` step


ashleymichal:
Today, Wrangler is rather prescriptive about what happens during the build stage; this works relatively well for Rust projects in particular, but begins to cause headaches for more complex javascript/webpack work flows. We should provide and document some way for users to customize their build step.

Related issues: 

* https://github.com/cloudflare/wrangler/issues/778
* https://github.com/cloudflare/wrangler/issues/428

ashleymichal:
bug report from #428 

> We assume that the existence of node_modules in a webpack project means the dependencies are installed and we don't need to run npm install again. However, if the user does ctrl-c during wrangler build, wrangler publish or wrangler preview while the npm install stage is running, it is possible to have a node_modules folder with invalid contents.

mrw:
This would be super helpful for my use case as well. I proposed a "pre-publish" command in #1224 but customizing the build process in general would definitely accomplish the same end-goal:

#### Overview and problem statement

I think it'd be helpful for us Workers Sites users if `wrangler.toml` allowed a "pre-publish" command to be set.

In my case I'm using Jekyll to generate the static files prior to publishing with wrangler. When running `jekyll build` for production, we'll often include different command line variables or Jekyll config than we do locally/on staging so the output is optimized for production. It'd be great if wrangler allowed a "pre-publish" command to be set so it automatically runs the proper Jekyll build command before publishing.

#### Basic example

In `wrangler.toml`, something like this:

```toml
[site]
bucket = "./_site"
entry-point = "workers-site"
pre-publish = "jekyll build --config ./jekyll-production-config.yml"
```

Thank you!

erickskrauch:
Bump.

jamesarosen:
Our use-case: we have a monorepo containing a number of different workers plus packages with shared worker code:
```
my-repo/
  packages/
    worker-library-1/
      package.json
      index.js
    worker-library-2/
      package.json
      index.js
    worker-a/
      package.json
      index.js
      wrangler.toml
    worker-b/
      package.json
      index.js
      wrangler.toml
```
We use a yarn workspace and `yarn install` to install the dependencies. (`pnpm` would work just as well for our case.)

We can't use [this suggestion](https://github.com/cloudflare/wrangler/issues/778#issuecomment-541193202) to include an empty `package.json` in the worker directories because those `package.json` files declare real dependencies to other packages in the workspace.

thefill:
[...]

Issue #482, opened 2019-08-27
Allow users to specify worker directory when running `wrangler publish`


gabbifish:
#### Overview and problem statement
There are Wrangler users that may use workers in a monorepo; that is, they will have multiple generated workers in one repository. This means that, for actions like CI, `wrangler publish` may be run on a specific worker within a monorepo. This is a case where providing an optional path argument for `wrangler publish` may be useful.

#### Basic example

Include a basic code example of this new feature if possible. Omit this section if not applicable.
Let's say we have a repo with multiple generated workers directories:
```
.
├── a
└── b
```
If we want to deploy the worker in directory `a` from the root of this repo, we would run
```
$ wrangler publish a
``` 
If we are just deploying from a worker directory, the preexisting `wrangler publish` invocation should still work.

EverlastingBugstopper:
+1

PR #1748, opened 2021-02-01
Feature/monorepo support

thefill:
Improvement of the detection of node_modules dir - adds bottom-up traversing to find it in parent dirs.

This should resolve some issues experienced with monorepos (e.g. with nx).

Example of an issue this can fix: https://github.com/cloudflare/wrangler/issues/820#issuecomment-762716331

To resolve most of the monorepo-related issues, we would need to refactor src/wranglerjs/mod.rs a bit more to establish the common root of node project (package.json and node_module location) and share it across a few other functions - still, that's a bit more work. I can do this some other time if you are keen on that.

### DIFF
--- a/src/wranglerjs/mod.rs
+++ b/src/wranglerjs/mod.rs
@@ -234,7 +234,7 @@ fn run_npm_install(dir: &PathBuf) -> Result<(), failure::Error> {
 
     if !dir.join("node_modules").exists() {
         // no dir in current path, search for closest
-        match find_closest_dir("node_modules", dir.as_path()) {
+        match find_closest("node_modules", dir.as_path()) {
             Some(_) => log::info!("skipping npm install because node_modules exists in parent dir"),
             None => {
                 let mut command = build_npm_command();
@@ -262,7 +262,7 @@ fn run_npm_install(dir: &PathBuf) -> Result<(), failure::Error> {
     Ok(())
 }
 
-// find closest (up-the-tree) location of a file or dir
+// find closest (bottom-up traversal) location of a file or dir
 fn find_closest<'a>(name: &str, path: &'a Path) -> Option<&'a Path>{
     if has_dir(name, path) {
         return Option::from(path);
--- a/src/wranglerjs/mod.rs
+++ b/src/wranglerjs/mod.rs
@@ -263,7 +263,7 @@ fn run_npm_install(dir: &PathBuf) -> Result<(), failure::Error> {
 }
 
 // find closest (bottom-up traversal) location of a file or dir
-fn find_closest<'a>(name: &str, path: &'a Path) -> Option<&'a Path>{
+fn find_closest<'a>(name: &str, path: &'a Path) -> Option<&'a Path> {
     if has_dir(name, path) {
         return Option::from(path);
     }
@@ -272,7 +272,7 @@ fn find_closest<'a>(name: &str, path: &'a Path) -> Option<&'a Path>{
     let parent = path.parent();
     match parent {
         Some(parent_path) => find_closest(name, parent_path),
-        None => None
+        None => None,
     }
 }
 
--- a/src/wranglerjs/mod.rs
+++ b/src/wranglerjs/mod.rs
@@ -4,7 +4,6 @@ pub mod output;
 
 pub use bundle::Bundle;
 
-use std::env;
 use std::fs;
 use std::fs::File;
 use std::iter;
@@ -13,6 +12,7 @@ use std::process::Command;
 use std::sync::mpsc::{channel, Sender};
 use std::thread;
 use std::time::Duration;
+use std::{env, io};
 
 use fs2::FileExt;
 use notify::{self, RecursiveMode, Watcher};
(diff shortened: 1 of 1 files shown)


xtuc (approved):

thefill:
Another issue that can be resolved by this PR: https://github.com/cloudflare/wrangler/issues/482

Limitation: As long as there is an existing "webpack.config.json" in the worker dir, users can successfully execute any cli subcommands (dev, build etc.) without triggering local "npm install". If there is none cli will be seeking package.json.

src/wranglerjs/mod.rs
```diff
@@ -256,6 +262,32 @@ fn run_npm_install(dir: &PathBuf) -> Result<(), failure::Error> {
...
+    let parent = path.parent();
+    match parent {
+        Some(parent_path) => find_closest(name, parent_path),
+        None => None,
+    }
+}
+
+// check if dir has file with provided name
+fn has_dir(name: &str, path: &Path) -> bool {
+    for entry in fs::read_dir(path).unwrap() {
```
ags799, line 281: Mind using `?` instead of `unwrap`?

thefill: yup, good call

thefill: added a bit more Err handling

ags799 (approved):
Thanks!

thefill:
@ags799 np, glad I could be of help!

thefill:
@ags799 any timeline for this change to be released? Not rushing things - just want to use this in a monorepo!

ags799:
Within a week!

thefill:
@ags799 amazing, thanks for an update!

ags799:
@thefill FYI the power outages in Texas have affected several of our engineers. The release could be delayed for another week.

thefill:
> @thefill FYI the power outages in Texas have affected several of our engineers. The release could be delayed for another week.

oh my, I have heard it's harsh there -_- no worries & thanks for letting me know.

xortive:
I think we need to revert this PR, unfortunately. It's causing issues for the `cloudflare-docs` repo (and I think all workers-sites projects), I believe because it's detecting the `node_modules` in the root when attempting to run `npm install` in the `workers-site` directory.

@thefill If you could work on a fix, I can take a look at it monday.

thefill:
oh snap, sure will have a look ;-)

thefill:
> I think we need to revert this PR, unfortunately. It's causing issues for the `cloudflare-docs` repo (and I think all workers-sites projects), I believe because it's detecting the `node_modules` in the root when attempting to run `npm install` in the `workers-site` directory.
> 
> @thefill If you could work on a fix, I can take a look at it.

Had a look at the cloudflare-docs@master and on the surface looks like a very simple setup - with nested package.json files, which should not cause problems. But obv the actions themselves have few more steps so I sure I'm missing things.

Any chance you can point me to the build that is failing (you have quite a few failures in actions across the last few days - not sure all relate to the issue).

thefill:
> > I think we need to revert this PR, unfortunately. It's causing issues for the `cloudflare-docs` repo (and I think all workers-sites projects), I believe because it's detecting the `node_modules` in the root when attempting to run `npm install` in the `workers-site` directory.
> > @thefill If you could work on a fix, I can take a look at it.
> 
> Had a look at the cloudflare-docs@master and on the surface looks like a very simple setup - with nested package.json files, which should not cause problems. But obv the actions themselves have few more steps so I sure I'm missing things.
> 
> Any chance you can point me to the build that is failing (you have quite a few failures in actions across the last few days - not sure all relate to the issue).

ok, I see where the problem is - all clear. We need to make it a bit more explicit in detecting its node_modules by correlating it with the location of the closest package.json.

This should follow the principle of most node projects as by design node_module should be located on the same level as package.json

thefill:
So the logic will go as follows:
- Search for the closest package.json 
- if found check if there is node_module in the same dir - if yes all is honky dory
- else follow the current fallback (npm i etc)

That's just a quick draft - I'm sure this will change a bit when having code in front of my face.

LeoColomb:
> * else follow the current fallback (npm i etc)

Then we lose the "monorepo" support, right? It's common to have a `package.json` alone, and another one above in the tree with the `node_modules` directory.

thefill:
> > * else follow the current fallback (npm i etc)
> 
> Then we lose the "monorepo" support, right? It's common to have a `package.json` alone, and another one above in the tree with the `node_modules` directory.

Ah, you mean orphaned package.json for e.g. release purposes? Uhh yeah, that's a good point. 

It all comes back to the question - should wrangler CLI be capable of handling all possible monorepo configs? It's an important one as I have seen in the past mono repo where you have node_modules on different levels - e.g. root provides tools, below one provides specific dependencies. Obv node resolves this but this CLI also have the job of installing wranglerjs module so where in that case it should do this?

This begs for the installation of wranglerjs not during the execution of the "wrangler build" but rather during CLI installation e.g. during "npm i @cloudflare/wrangler -g"

The flow could be as follows:
- user installs wrangler globally
- wrangler installs as a sidecar wranglerjs in its dir
- when user executes "wrangler build" / "wrangler dev"
- wrangler uses its own wranglerjs version

 This would resolve many potential monorepo problems like dir structure or tooling type (yarn / npm / pnpm).

xortive:
I think #1677/custom builds will support the monorepo use case, since it doesn't require `wranglerjs`, and therefore doesn't automatically run `npm install`.

If you agree, I think we should revert this and release `1.14.2`. Once #1677 makes it out into a release candidate which should be soon, folks using monorepos can convert to using custom builds.

thefill:
> I think #1677/custom builds will support the monorepo use case, since it doesn't require `wranglerjs`, and therefore doesn't automatically run `npm install`.
> 
> If you agree, I think we should revert this and release `1.14.2`. Once #1677 makes it out into a release candidate which should be soon, folks using monorepos can convert to using custom builds.

yep, I think that's the safest way going forward - trying to guess what different projects needs is pointless - we will always find an edge case.

Issue #1510, opened 2020-08-19
wrangler dev: panicked at 'Failed to connect to devtools instance'


JakubKoralewski:
#### Environment

* operating system: Microsoft Windows [Version 10.0.18363.1016]
* output of `rustc -V`: rustc 1.44.1 (c7087fe00 2020-06-17)
* output of `node -v`: v14.5.0
* output of `wrangler -V`: wrangler 1.10.3
* contents of `wrangler.toml`

```toml
name = "name-dev"
type = "webpack"
webpack_config = "webpack.dev.js"
account_id = "id ahsjkdhalsdhalkjdh"
workers_dev = true
route = ""
zone_id = ""

[env.production]
name = "name"
webpack_config = "webpack.config.js"
```

#### Steps to reproduce

If I leave `wrangler dev` running for a long time sometimes this happens. It is unclear for me why. It may be my unstable internet connection, but still if so, I'd expect for wrangler to keep retrying until the connection is restored or something like that for a more graceful experience.

#### What did you see instead?

message 'Nieznany host' means 'unknown host'

```
thread 'tokio-runtime-worker' panicked at 'Failed to connect to devtools instance: Io(Os { code: 11001, kind: Other, message: "Nieznany host." })', C:\Users\Admin\.cargo\registry\src\github.com-1ecc6299db9ec823\wrangler-1.10.3\src\commands\dev\socket.rs:29:30
stack backtrace:
   0: backtrace::backtrace::dbghelp::trace
             at C:\Users\VssAdministrator\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.46\src\backtrace/dbghelp.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at C:\Users\VssAdministrator\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.46\src\backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src\libstd\sys_common/backtrace.rs:78
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src\libstd\sys_common/backtrace.rs:59
   4: core::fmt::write
             at src\libcore\fmt/mod.rs:1069
   5: std::io::Write::write_fmt
             at src\libstd\io/mod.rs:1504
   6: std::sys_common::backtrace::_print
             at src\libstd\sys_common/backtrace.rs:62
[...]

PR #1752, opened 2021-02-03
endlessly retry connection to devtools websocket

ags799:
Endlessly retry connection to preview's devtools websocket on `wrangler dev`. With exponential backoff.

Keeps us from panicking in #1510.

Output looks like
Screen Shot 2021-02-03 at 3 47 16 PM

### DIFF
--- a/src/commands/dev/socket.rs
+++ b/src/commands/dev/socket.rs
@@ -85,7 +85,8 @@ async fn connect_retry(
                 ));
                 delay_for(Duration::from_secs(wait_seconds)).await;
                 wait_seconds = wait_seconds.pow(2);
-                if wait_seconds > maximum_wait_seconds { // max out at 60 seconds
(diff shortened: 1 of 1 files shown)


ObsidianMinor (approved):

ObsidianMinor:
We should probably add a limit to the exponential back-off so users aren't forced to wait an extremely long time if they leave their computer for a bit.

ags799:
5 minutes?

ObsidianMinor:
Maybe one minute or 30 seconds? It doesn't really hurt to reconnect and we don't want to force users to wait to long (though they can always just kill it and restart wrangler dev).

ags799:
> Maybe one minute or 30 seconds? It doesn't really hurt to reconnect and we don't want to force users to wait to long (though they can always just kill it and restart wrangler dev).

done

PR #1762, opened 2021-02-08
Set panic to abort in release mode

ObsidianMinor:
This should fix cases where we spawn threads that panic and get us in an invalid state that requires us to get killed anyway.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,3 +79,6 @@ strict = ['openssl/vendored']
 sys-openssl = ['openssl']
 # Keeping feature for users already using this feature flag
 vendored-openssl = ['openssl/vendored']
+
+[profile.release]
+panic = "abort"


ObsidianMinor:
We currently don't handle unwinds so switching to abort shouldn't cause any issues. If we do want to start handling unwinds, we'll need to remove this.

ags799:
This means that if a thread panics other threads can continue? Can you give an example where that would be useful?

ObsidianMinor:
That's the current behavior. Unwind causes panics to unwind the thread. If the panic happens on a thread that's not the main thread, the thread just returns an error on join, but if you drop the join handle (like we do every time we spawn a thread), the thread just dies and everything continues as normal.

That's not what we want to happen, since we want real errors (failure::Error), not panics. Abort changes the behavior to process abort which can never leave us in a bad state if a thread panics. For example, if the fs change notifier thread panics, the process won't be left hanging doing nothing.

ags799 (approved):

xortive:
@ObsidianMinor not super familiar with this, this won't cause errors to not be printed when a thread panics right?

ObsidianMinor:
> this won't cause errors to not be printed when a thread panics right?

Yes, panics still go through the panic handler, but instead of the panic handler unwinding, it simply aborts after printing.

PR #1764, opened 2021-02-10
Fix deeply nested module imports

Electroid:
In modules mode, when importing a file in a directory, it would get uploaded incorrectly.

`index.mjs`:
```js
import file from "./assets/file.txt"
```

`FormData`:
Before: `file.txt => "..."`
Now: `assets/file.txt => "..."`

### DIFF
--- a/src/upload/form/mod.rs
+++ b/src/upload/form/mod.rs
@@ -114,8 +114,14 @@ pub fn build(
                         let entry = entry?;
                         let path = entry.path();
                         if path.is_file() {
-                            log::info!("Adding module {}", path.display());
-                            modules.push(Module::new(path.to_owned())?);
+                            let name = path
+                                .strip_prefix(config.upload_dir.clone())?
+                                .to_owned()
+                                .as_path()
+                                .display()
+                                .to_string();
+                            log::info!("Adding module {}", name);
+                            modules.push(Module::new(name, path.to_owned())?);
                         }
                     }
 
--- a/src/upload/form/mod.rs
+++ b/src/upload/form/mod.rs
@@ -1,26 +1,32 @@
+mod modules_worker;
 mod plain_text;
 mod project_assets;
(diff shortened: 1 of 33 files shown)


src/upload/form/mod.rs
```diff
@@ -117,8 +117,14 @@ pub fn build(
                         let entry = entry?;
                         let path = entry.path();
                         if path.is_file() {
-                            log::info!("Adding module {}", path.display());
-                            modules.push(Module::new(path.to_owned())?);
+                            let name = path
+                                .strip_prefix(config.upload_dir.clone())?
+                                .to_owned()
+                                .as_path()
+                                .display()
```
xortive, line 124: This should use the `to_str` method of path and return an error if the conversion fails (due to non-UTF-8 in the path name)

src/upload/form/mod.rs
```diff
@@ -117,8 +117,14 @@ pub fn build(
...
+                                .to_string();
+                            log::info!("Adding module {}", name);
```
xortive, line 126: would be nice to include the path here as well

src/upload/form/mod.rs
xortive: presumably for a path like like

`/home/matt/worker/dist/nested/nested/index.js`

This returns `nested/nested/index.js`?

xortive (changes requested):
Nice catch, just handle the UTF-8 thing and I'll merge

Electroid:
Realized the previous approach would not work on Windows machines, so had to add a dependency to help us out.

xortive:
@Electroid could you rebase this on the `malonso/modules-rc` branch? I split that out from the old `modules-do` branch, and it looks like this PR is trying to add the durable objects support back in along with your fixes.

xortive (approved):
Thanks @Electroid!

PR #1781, opened 2021-02-25
Durable Objects support

xortive:
same as what was in the last PR


Electroid:
I just ran into an issue. If you implement DO in a script and try to delete it (remove references from both code and wrangler.toml), you can't.
```
    {
      "code": 10064,
      "message": "new version of script does not implement class Counter which is depended on by Durable Object namespace Counter"
    }
```

Moreover, it also doesn't allow me to delete it from the dashboard, but that's an issue for the dashboard.

xortive:
closing in favor of https://github.com/cloudflare/wrangler/pull/1812

PR #1789, opened 2021-03-04
Changelog generator

xortive:
Thanks @ags799 for the idea, this will be a great QoL improvement for those on release duty :)


nataliescottdavidson (approved):
Thanks Matt! This is super helpful

PR #1799, opened 2021-03-05
ci: add build config for aarch64-apple-darwin target

nilslice:
Updates github action to support build & release of the `aarch64-apple-darwin` target needed for M1 devices.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,7 @@
-on:
-  push:
-    tags:
-      - "v*" # Run when tag matches v*, i.e. v1.0, v20.15.10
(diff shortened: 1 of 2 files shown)


xortive:
@nilslice can we configure the action to run on all pull requests (it'll just run on this one until we merge it) so we can test it, then we'll put it back to only running on tagged commits before we merge, so we can test without releasing?

nilslice:
@xortive, good call. It's updated and running now. Once this build passes, I'll revert that change. I also want to file a ticket to track future optimization work so that the apple silicon builds happen from a new matrix entry and job so it's done in parallel with the others.

Issue #1715, opened 2021-01-10
JavaScript project fails without package.json


Electroid:
### 🐛 Bug Report

Wrangler errors without a package.json for JavaScript projects, yet there is no reason to require it.

#### Environment

* output of `wrangler -V`: wrangler 1.12.3
* contents of `wrangler.toml`: type=javascript

#### Steps to reproduce

1. Create a hello world script with `type=javascript`
2. Run `wrangler dev` or `wrangler publish`
3. See the error `Error: Your JavaScript project is missing a package.json file`

#### What did you expect to see?

Wrangler should be able to publish a plain JavaScript project without a package.json.

rbrishabh:
Hi @Electroid, I would like to work on this! Could you please help me get started?

squaresmile:
For others, you need to follow [the required file layout](https://developers.cloudflare.com/workers/platform/sites/configuration/) of the `entry-point` option for worker sites.

For [my project](https://github.com/atlasacademy/apps/blob/master/packages/db-og-worker/wrangler.toml), I copy `package.json` and `package-lock.json` to the `dist` folder and change `main` in `package.json` from `dist/worker.js` to `worker.js`.

PR #1808, opened 2021-03-09
Remove package.json requirement for JS workers

lsrugo:
Fixes issue #1715 
For now it uses index.js as the default entry point. Maybe there should be an option to change that.

### DIFF
--- a/src/settings/toml/target.rs
+++ b/src/settings/toml/target.rs
@@ -12,6 +12,7 @@ pub struct Target {
     pub account_id: String,
     pub kv_namespaces: Vec<KvNamespace>,
     pub name: String,
+    pub main: String,
     pub target_type: TargetType,
     pub webpack_config: Option<String>,
     pub site: Option<Site>,
--- a/src/commands/kv/mod.rs
+++ b/src/commands/kv/mod.rs
@@ -120,6 +120,7 @@ mod tests {
                 },
             ],
             name: "test-target".to_string(),
+            main: "".to_string(),
             target_type: TargetType::Webpack,
             webpack_config: None,
             site: None,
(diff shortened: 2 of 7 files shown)


xtuc:
Could you please add a unit test to cover your change?

lsrugo770:
I updated the unit test by removing the step which creates the package.json, to make sure that it passes without that file

Electroid (changes requested):
Hey @lsrugo770, thanks for working on this!

The team discussed this and one of the points of feedback that came up was the ability to specify the `main` path in the `wrangler.toml`. We've had issues in the past with hardcoded values, like `index.js`. Is that a change you would be willing to add to your PR? It would end up look something like this:

```toml
name = "cool-thing"
type = "javascript"
main = "dist/worker.mjs" # Optional, if not specified, then check package.json
```

If you're not up for this, totally understand as well. Appreciate the time you put into this :)

lsrugo770:
Thank you for getting back to me. This is my first time using Rust, but I will give it a shot.

lsrugo770:
I think it's done. Let me know if I missed something

src/settings/toml/target.rs
```diff
@@ -12,6 +12,7 @@ pub struct Target {
     pub account_id: String,
     pub kv_namespaces: Vec<KvNamespace>,
     pub name: String,
+    pub main: String,
```
xortive, line 15: nit: it would be nice if this was an `Option<String>`, and you used serde's `string_empty_as_none` for the entry in `manifest.rs`

ytbigvinny27 (approved):
The Greed and compromised will always jump first 🙄but soon♥️🔜🤌🏽🔝

brunolca:
hi! any update on this?

threepointone:
wrangler 2 requires an explicit entry point now, either via the `main` field in `wrangler.toml`, or as a cli arg. closing this PR.

PR #1812, opened 2021-03-12
Malonso/durable objects

xortive:
- [x] use class and script name for DO bindings instead of namespace id
- [x] adhoc create/delete/unused migration support
- [ ] tagged create/delete/unused migration support
    - [ ] fetch current migration tag from API
    - [ ] coalesce migrations from wrangler.toml into single API migration
- [ ] support for rename/transfer migrations

### DIFF
--- a/src/commands/publish.rs
+++ b/src/commands/publish.rs
@@ -31,7 +31,7 @@ pub fn publish(
 ) -> Result<(), failure::Error> {
     validate_target_required_fields_present(target)?;
 
-    let deploy = |target: &Target| match deploy::deploy(&user, &deployments) {
+    let run_deploy = |target: &Target| match deploy::deploy(&user, &deployments) {
         Ok(results) => {
             build_output_message(results, target.name.clone(), out);
             Ok(())
@@ -88,10 +88,11 @@ pub fn publish(
 
         let upload_client = http::featured_legacy_auth_client(user, Feature::Sites);
 
+        deploy::pre_upload()?;
         // Next, upload and deploy the worker with the updated asset_manifest
         upload::script(&upload_client, &target, Some(asset_manifest))?;
 
-        deploy(target)?;
+        run_deploy(target)?;
 
         // Finally, remove any stale files
         if !to_delete.is_empty() {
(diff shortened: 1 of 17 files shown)


src/commands/publish.rs
```diff
@@ -112,6 +88,7 @@ pub fn publish(
 
         let upload_client = http::featured_legacy_auth_client(user, Feature::Sites);
 
+        deploy::pre_upload()?;
```
ags799, line 91: Why are we putting this empty function here and elsewhere?

ags799 (approved):

caass (commented):
Looks pretty good! A couple things we can chat about on monday but overall happy with the way this looks -- is there a PR in the works to update docs as well?

src/commands/publish.rs
xortive: `pre_upload()` is for doing tasks that require making an API call before uploading the script. This is going to be used to fetch the current migration tag for tagged migrations, so wrangler can look at the list of migrations in `wrangler.toml` and know how to coalesce them.

I had already done some of the work for tagged migrations and this is part of it, so I figured i'd leave it in

src/settings/toml/durable_objects.rs
```diff
@@ -0,0 +1,79 @@
+use serde::{Deserialize, Serialize};
+
+use crate::settings::binding::Binding;
+
+#[derive(Clone, Debug, Default, Deserialize, PartialEq, Serialize)]
+pub struct DurableObjects {
+    pub classes: Option<Vec<DurableObjectsClass>>,
+    #[serde(skip_deserializing)] // TODO remove when tagged migrations are fully supported
```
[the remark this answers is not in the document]
xortive: migrations are how you tell the platform to
* activate (for lack of a better word) a durable objects class such that you can address it, and create durable objects implemented by that class
* delete all of the durable objects implemented by a durable objects class
* rename durable objects classes
* transfer durable objects between classes

right now wrangler only supports untagged migrations specified when publishing -- this is fine for small projects, but teams will need tagged migrations that can be committed to version control.

The migration tag just serves as a way to get the latest migration that the platform has actually been executed, so when you `publish` with several un-executed migrations after the latest one in `wrangler.toml`, those migrations can get combined into a single migration.

xortive:
Pre-merge tasks:

- [ ] help text for migration directives
- [ ] change modules support to only treat .bin as an arraybuffer -- unknown extensions are an error
- [ ] rename deploy output helper to run_deploy

PR #1816, opened 2021-03-17
#1386: Support a configurable timeout

SirJosh3917:
This PR supplies a robust solution to #1386. It offers three new configurable variables, `http_timeout`, `connect_timeout`, and `bulk_timeout` which are used in the places where timeout durations were previously hardcoded.

Unit tests were added to confirm that the features added work. For a final manual sanity check, I ran `wrangler publish` on a 2.4MiB JS file (on my 0.5Mbps upload internet), one without setting the related environment variables and one with. When the timeout variables weren't set, my internet upload was getting maxed out, and when setting the timeout to `0` the upload instantly failed. huzzah :DDD :tada:

```
$ CF_HTTP_TIMEOUT=0 CF_CONNECT_TIMEOUT=0 CF_BULK_TIMEOUT=0 wrangler publish
Error: error sending request for url (https://api.cloudflare.com/client/v4/accounts/x/workers/scripts/worker): operation timed out
```

Relevant documentation should be added to notify users that they can set these environment variables if they need to increase the timeout.

### DIFF
--- a/src/http/cf.rs
+++ b/src/http/cf.rs
@@ -1,18 +1,16 @@
-use std::time::Duration;
-
 use cloudflare::framework::async_api;
 use cloudflare::framework::auth::Credentials;
 use cloudflare::framework::response::ApiFailure;
 use cloudflare::framework::{Environment, HttpApiClient, HttpApiClientConfig};
 use http::StatusCode;
 
-use crate::http::{feature::headers, Feature, DEFAULT_HTTP_TIMEOUT_SECONDS};
+use crate::http::{feature::headers, Feature};
 use crate::settings::global_user::GlobalUser;
 use crate::terminal::emoji;
 use crate::terminal::message::{Message, StdOut};
 pub fn cf_v4_client(user: &GlobalUser) -> Result<HttpApiClient, failure::Error> {
     let config = HttpApiClientConfig {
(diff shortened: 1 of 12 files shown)


Electroid (changes requested):
Thanks @SirJosh3917, definitely understand the problem of timeouts with slower Internet speeds.

I think a challenge with this PR is it's a lot of new code for a fairly trivial/minor change (though appreciate the tests you included). Do you think we could solve the same problem with a simpler solution like increasing the timeouts to 5 minutes, and 15 minutes for the bulk API? That would also make it easier for us to get this into the next wrangler release.

SirJosh3917:
> Do you think we could solve the same problem with a simpler solution like increasing the timeouts to 5 minutes, and 15 minutes for the bulk API?

I can't disagree - that would solve the problem for the majority of users, and a much simpler one at that. However, one drawback I see with this is that if users cannot reach the target endpoint, they will have to wait for the duration of the timeout before an error occurs. The benefit of a robust solution such as this one allows users to be conscious of how long they have until they time out.

As for specific changes to the timeout values, this problem has been solved for me ever since the timeout was increased to a minute, so I cannot weigh in on a value for a good increase.

> That would also make it easier for us to get this into the next wrangler release.

If this PR is too much to properly code review for the time being, I don't mind it being put on the back burner. The current timeout values satisfy my needs, so I have no dire need for this to get in anytime soon.

PR #1824, opened 2021-03-24
Refine migration support for durable objects

xortive:
This PR adds support for rename and transfer migrations

### DIFF
--- a/src/commands/dev/mod.rs
+++ b/src/commands/dev/mod.rs
@@ -9,7 +9,7 @@ pub use server_config::Protocol;
 pub use server_config::ServerConfig;
 
 use crate::build::build_target;
-use crate::deploy::{self, DeployTarget, DeploymentSet};
+use crate::deploy::{DeployTarget, DeploymentSet};
 use crate::settings::global_user::GlobalUser;
 use crate::settings::toml::Target;
 use crate::terminal::message::{Message, StdOut};
@@ -28,8 +28,6 @@ pub fn dev(
 ) -> Result<(), failure::Error> {
     // before serving requests we must first build the Worker
     build_target(&target)?;
-
-    deploy::pre_upload()?;
 
     let deploy_target = {
         let valid_targets = deployments
--- a/src/settings/toml/manifest.rs
+++ b/src/settings/toml/manifest.rs
@@ -336,6 +336,7 @@ impl Manifest {
             name: self.name.clone(), // Inherited
             kv_namespaces: get_namespaces(self.kv_namespaces.clone(), preview)?, // Not inherited
             durable_objects: self.durable_objects.clone(), // Not inherited
+            migrations: None,        // TODO(soon) Allow migrations in wrangler.toml
             site: self.site.clone(), // Inherited
             vars: self.vars.clone(), // Not inherited
             text_blobs: self.text_blobs.clone(), // Inherited
--- a/src/settings/toml/migrations.rs
+++ b/src/settings/toml/migrations.rs
@@ -1,0 +1,57 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Default, Deserialize, PartialEq, Serialize)]
+pub struct Migrations {
+    pub migrations: Vec<MigrationConfig>,
+}
+
+impl Migrations {
+    pub fn api_migration(&self) -> Result<ApiMigration, failure::Error> {
+        // TODO: make api call to get most recent tag, and coalesce tags afterwards.
+        // For now, migrations will only ever have a single adhoc migration in it.
+        match &self.migrations {
+            migrations if migrations.len() == 1 => Ok(ApiMigration {
+                old_tag: None,
+                new_tag: None,
+                migration: migrations[0].migration.clone(),
+            }),
+            migrations => Self::coalesce_migrations(migrations),
+        }
+    }
+
+    fn coalesce_migrations(
+        _migrations: &[MigrationConfig],
+    ) -> Result<ApiMigration, failure::Error> {
+        unimplemented!()
+    }
+}
+
+#[derive(Clone, Debug, Default, Deserialize, PartialEq, Serialize)]
+pub struct MigrationConfig {
+    pub tag: Option<String>,
+    #[serde(flatten)]
+    pub migration: Migration,
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Serialize)]
+pub struct ApiMigration {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub old_tag: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub new_tag: Option<String>,
+    #[serde(flatten)]
+    pub migration: Migration,
+}
+
+#[derive(Clone, Debug, Default, Deserialize, PartialEq, Serialize)]
+pub struct Migration {
+    #[serde(flatten)]
+    pub durable_objects: DurableObjectsMigration,
+}
+
+#[derive(Clone, Debug, Default, Deserialize, PartialEq, Serialize)]
+pub struct DurableObjectsMigration {
+    pub new_classes: Vec<String>,
+    pub deleted_classes: Vec<String>,
+    pub unused_classes: Vec<String>,
+}
--- a/src/settings/toml/migrations.rs
+++ b/src/settings/toml/migrations.rs
@@ -1,3 +1,4 @@
+use clap::{App, Arg, ArgGroup, ArgMatches};
 use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Debug, Default, Deserialize, PartialEq, Serialize)]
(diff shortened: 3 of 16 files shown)


src/util.rs
```diff
@@ -0,0 +1,10 @@
+use clap::App;
+pub trait ApplyToApp<'a, 'b> {
+    fn apply(self, f: fn(App<'a, 'b>) -> App<'a, 'b>) -> App<'a, 'b>;
+}
+
+impl<'a, 'b> ApplyToApp<'a, 'b> for App<'a, 'b> {
+    fn apply(self, f: fn(App<'a, 'b>) -> App<'a, 'b>) -> App<'a, 'b> {
```
xortive, line 7: the crab language is hard sometimes

caass: i love the smell of lifetimes in the morning...

xortive:
@a-robinson particularly interested in your review of the migration coalescing algorithm
@ags799 @caass tagged y'all for the more wrangler-specific pieces of this PR.

src/util.rs
xortive: This apply thing would be pretty useful for cleaning up main. I added it since I wanted to re-use the setup for the command line arguments in my test

src/settings/toml/migrations.rs
```diff
@@ -0,0 +1,499 @@
...
+                    "https://api.cloudflare.com/client/v4/accounts/{}/workers/scripts",
+                    account_id,
+                );
+
+                let client = http::legacy_auth_client(user);
+                let res = client.get(&list_scripts_addr).send()?;
+                let scripts = {
+                    let res: ListScriptsV4ApiResponse = res.json()?;
+                    res.result
+                };
```
ags799, line 38: purely up to you, but I think starting a new scope here is kind of weird, expected to see
```
let res = ...
let res_json: ... = res.json()?;
let scripts = res.result;
```
but it really doesn't matter

src/settings/toml/migrations.rs
```diff
@@ -0,0 +1,499 @@
...
+
+                let script = scripts.iter().find(|script| script.name == script_name);
+
+                match script {
+                    None => Self::coalesce_migrations(migrations, None),
+                    Some(ApiScript {
+                        name: _,
+                        migration_tag: None,
+                    }) => Self::coalesce_migrations(migrations, None),
```
ags799, line 47: you could use `|` to merge these cases

src/commands/dev/mod.rs
```diff
@@ -29,8 +29,6 @@ pub fn dev(
     // before serving requests we must first build the Worker
     build_target(&target)?;
 
-    deploy::pre_upload()?;
-
```
caass, line 33: Nice

src/settings/toml/manifest.rs
```diff
@@ -336,8 +340,12 @@ impl Manifest {
             name: self.name.clone(), // Inherited
             kv_namespaces: get_namespaces(self.kv_namespaces.clone(), preview)?, // Not inherited
             durable_objects: self.durable_objects.clone(), // Not inherited
-            site: self.site.clone(), // Inherited
-            vars: self.vars.clone(), // Not inherited
+            migrations: self // Top level
+                .migrations
+                .clone()
+                .map(|migrations| Migrations { migrations }),
```
caass, line 346: Wait...is this gonna work if migrations is `None`? I assume it will but I don't see any like, filtering

src/settings/toml/migrations.rs
```diff
@@ -0,0 +1,499 @@
...
+            .takes_value(true)
+            .number_of_values(3)
+            .value_names(&["from script", "from class", "to class"])
+            .multiple(true)
+        )
+        .group(
+            ArgGroup::with_name("migrations")
+            .args(&["new_class", "delete_class", "rename_class", "transfer_class"])
+            .multiple(true)
+        )
```
caass, line 221: dude this pattern is so good...we should make this a trait and impl for subcommands in a future pr

src/util.rs
caass: yeah i left a comment, def think this should become a trait in a future pr

caass (commented):
I'd like to understand what this changes in terms of wrangler usage. This basically adds a [migrations] tag to wrangler.toml? and then you update that every time you migrate?

xortive:
> I'd like to understand what this changes in terms of wrangler usage. This basically adds a [migrations] tag to wrangler.toml? and then you update that every time you migrate?

yup! it's a top-level array of migrations, every time you add/rename/transfer classes, you'd add a new entry to the [[migrations]] array

src/settings/toml/manifest.rs
xortive: yep, map will pass `None` through

src/commands/dev/mod.rs
xortive: I'm still iffy on whether or not it would have been nicer to keep this, or to do all the hackery I did to get a GlobalUser to pass into `api_migrations()`.

src/settings/toml/migrations.rs
xortive: I quite like using scopes to avoid having to name intermediate results

src/settings/toml/migrations.rs
xortive: good catch, will do

src/settings/toml/migrations.rs
nilslice: correct me if I'm mistaken or don't have the full context, but can't you avoid the intermediate result and still try the result like this? 

```rust
let scripts = res.json::<ListScriptsV4ApiResponse>()?.result;
```

(not tested !)

nilslice:
sorry, laptop keyboard went crazy and issued some clicks.. didn't mean to add the `blocked` label 😨

xortive:
fixed some logic bugs with the input to migration coalescing I discovered after testing:

- duplicate migration tags are properly handled
- removed special case for a single migration that incorrectly attempted to push it up with no tags
- the old migration in `wrangler.toml` was incorrectly being included as part of coalescing. we now start at the next migration

src/settings/toml/migrations.rs
```diff
@@ -0,0 +1,504 @@
...
+            .to_vec();
+        let mut transferred_classes = migrations[0]
+            .migration
+            .durable_objects
+            .transferred_classes
+            .to_vec();
+
+        for mc in &migrations[1..] {
+            // Renames will modify existing new, rename, and transfer migrations if they are present
+            for rename in mc.migration.durable_objects.renamed_classes.iter() {
```
a-robinson, line 98: If someone has multiple rename migrations away from the same class name (which is also a target of one of the prior migrations), then this will probably lead to weird error messages from the API (along the lines of "class doesn't exist"). But that should be quite uncommon, and it wouldn't be easy to do better here without a bunch more error checking that's already done in the API.

So this looks good 👍

src/settings/toml/migrations.rs
```diff
@@ -0,0 +1,504 @@
...
+
+            // deletes will remove existing new, rename, and transfer migrations if they are present
+            for delete in mc.migration.durable_objects.deleted_classes.iter() {
+                if let Some(i) = new_classes.iter().position(|c| c == delete) {
+                    new_classes.remove(i);
+                } else if let Some(i) = renamed_classes
+                    .iter_mut()
+                    .position(|existing| &existing.to == delete)
+                {
+                    renamed_classes.remove(i);
```
a-robinson, line 125: We need to apply the delete migration to the `from` class at this point, right? Otherwise it may stick around when it's supposed to be deleted.

src/settings/toml/migrations.rs
```diff
@@ -0,0 +1,504 @@
...
+                } else if let Some(i) = transferred_classes
+                    .iter_mut()
+                    .position(|existing| &existing.to == delete)
+                {
+                    transferred_classes.remove(i);
```
a-robinson, line 130: I think this is just straight-up invalid and should error out. Removing the transfer migration is definitely not the same as transferring a class and then deleting it, but there's also no way to convert this into a logically equivalent migration that deletes the class from the other script.

src/settings/toml/migrations.rs
a-robinson: Although I suppose some of this logic could go awry if the initial set of migrations is invalid. For example, if there are multiple matches for a given classname, won't this code only find and update the first? It's not valid to have multiple migrations targeting the same class, but unless wrangler has already validated that then it isn't really safe to assume here.

I think we either need to do that validation or attempt to update all matching migrations in this logic rather than just the first one. I'd personally suggest doing the former rather than the latter, since it seems simpler and more self-contained. You could probably just create a `set`, and as you process each individual migration (including the first batch of them above this loop), make sure its destination class isn't already the destination of a different migration.

src/settings/toml/migrations.rs
```diff
@@ -0,0 +1,504 @@
...
+                        renamed_classes: vec![rename_class("A"), rename_class("B")],
+                        transferred_classes: vec![transfer_class("A"), transfer_class("B")],
+                    }
+                }
+            })
+        );
+    }
+
+    #[test]
+    fn migration_coalescing() {
```
a-robinson, line 420: I think we'd benefit from some tests that have invalid migrations in them.

src/settings/toml/migrations.rs
```diff
@@ -0,0 +1,504 @@
...
+                } else {
+                    // rename does not modify an existing migration, add it to the final migration
+                    renamed_classes.push(rename.clone());
+                }
+            }
+
+            // deletes will remove existing new, rename, and transfer migrations if they are present
+            for delete in mc.migration.durable_objects.deleted_classes.iter() {
+                if let Some(i) = new_classes.iter().position(|c| c == delete) {
+                    new_classes.remove(i);
```
a-robinson, line 120: Hm, the more I read and comment on this logic the more worried I am that it's going to be very hard to get right here... What if there already is a class in the script (as previously stored in the API) that has durable objects on it? This will behave very differently than if we actually sent all the migrations to the server -- if we sent one batch at a time, it would be an error to pass a new-class migration for a class that already exists. But if we aggregate them like this, then the upload will silently succeed with the old class still in place.

I'm curious if you have ideas, but I see two options to avoid this sort of problem:

1. When we get the migration tag for the script, also get the list of DO classes currently implemented by the script and then use them to avoid problems like this. We'll have to add this to the API, but that's fine -- it'll be useful for other purposes anyway. What might be a bit less fun is duplicating all of the validation logic here, but it wouldn't be terribly hard.
2. Change the API to support passing an array of migration batches. I'm not sure this would actually reduce the amount of new validation code we need to write, since handling multiple batches at once will require a bunch of new logic will be required in the API -- what's there today isn't really structured for that.

So I think I'd lean towards the first option.

xortive: Yeah, I'm starting to agree with you as well here. I dislike the idea of re-implementing the validation logic, is there a particular reason you favor 1. over 2.? If it's going to be a similar amount of work, 2. would have the benefit of non-wrangler users getting to upload multiple migrations as well

xortive:
since migration coalescing / migration in toml seems to be somewhat issue-prone as currently implemented, I'll break out a separate PR for just rename/transfer migrations

src/settings/toml/migrations.rs
xortive: oo, good point

src/settings/toml/migrations.rs
a-robinson: I suppose the main reasons I had in mind are

1. We'll either have to break the existing API (and thus anyone using the current wrangler RC) in order to accept batches of migrations) or put in a hacky compatibility layer in the parsing of the request to deal with some clients sending batches and others assuming that only one batch (not in an array) can be sent, but given that I suppose such a layer is possible it isn't that big of a deal.
2. I don't think it's very likely that anyone other than wrangler will find the option to send multiple rounds of migrations useful, and we'd be making our API a little tougher to use for anyone who only ever sends a single set of migrations at a time.  I don't get the impression we're likely to have many people programming directly against the API, though.

Neither of those points is particularly strong, so I'm left not really caring all that much either way. We can put it in the API if that's your preference.

ags799 (approved):

xortive:
Added the aliases for durable objects classes section. in next RC we will make this a breaking change and remove support for the old name

before:
```toml
[durable_objects]
classes = [{ binding = "Counter", class_name = "Counter" }]
```

```toml
[durable_objects]
# binding = "Counter" would be accepted too, but it feels weird to duplicate "binding" in this manner. We should use name instead IMO, but I don't feel strongly if we want to keep using "binding"
bindings = [{ name = "Counter", class_name = "Counter" }]
```

a-robinson (approved):
LGTM assuming you confirmed the renamed fields work.

PR #1833, opened 2021-04-02
Allow non-Webpack Workers Sites

GregBrimble:
Right now, custom build commands don't work with Workers SItes. This should enable them.

### DIFF
--- a/src/settings/toml/manifest.rs
+++ b/src/settings/toml/manifest.rs
@@ -321,18 +321,16 @@ impl Manifest {
                     ))
                 }
                 TargetType::JavaScript => {
+                    let error_message = format!(
+                        "{} Workers Sites requires using a bundler, and your configuration indicates that you aren't using one. You can fix this by:\n* setting your project type to \"webpack\" to use our automatically configured webpack bundler.\n* setting your project type to \"javascript\", and configuring a build command in the `[build]` section if you wish to use your choice of bundler.",
+                        emoji::WARN
+                    );
                     if let Some(build) = &self.build {
                         if build.command.is_none() {
-                            failure::bail!(format!(
-                                "{} Workers Sites requires using a bundler, and your configuration indicates that you aren't using one. You can fix this by:\n* setting your project type to \"webpack\" to use our automatically configured webpack bundler.\n* setting your project type to \"javascript\", and configuring a build command in the `[build]` section if you wish to use your choice of bundler.",
-                                emoji::WARN
-                            ))
+                            failure::bail!(error_message)
                         }
                     } else {
-                        failure::bail!(format!(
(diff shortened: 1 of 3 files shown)


src/settings/toml/manifest.rs
```diff
@@ -327,7 +321,7 @@ impl Manifest {
         Not inherited: Must be defined for every environment individually.
         */
         let mut target = Target {
-            target_type,                                 // Top level
```
xortive, line 330: top level is correct -- we don't let you override target_type in environments

xortive (commented):
Could we:
* forbid the rust type (it won't work)
* for the javascript type fail if no `[build]`, or a `[build]` section with no/empty `command`? Something like:

```
Workers Sites requires using a bundler, and your configuration indicates that you aren't using one. You can fix this by:
* setting your project type to "webpack" to use our automatically configured webpack bundler.
* setting your project type to "javascriot", and configuring a build command in the `[build]` section if you wish to use your choice of bundler.
```

xortive (approved):

src/upload/form/mod.rs
```diff
@@ -89,7 +97,7 @@ pub fn build(
                     log::info!("Plain JavaScript project detected. Publishing...");
                     let package_dir = target.package_dir()?;
                     let package = Package::new(&package_dir)?;
-                    let script_path = package.main(&package_dir)?;
+                    let script_path = package_dir.join(package.main(&package_dir)?);
```
xortive, line 100: why the change here?

src/settings/toml/manifest.rs
```diff
@@ -312,11 +312,32 @@ impl Manifest {
...
+                TargetType::Rust => {
+                    failure::bail!(format!(
+                        "{} Workers Sites does not support Rust type projects.",
+                        emoji::WARN
+                    ))
+                }
+                TargetType::JavaScript => {
+                    if let Some(build) = &self.build {
+                        if build.command.is_none() {
+                            failure::bail!(format!(
```
xortive, line 326: nit: would be nice if this message wasn't duplicated but I'm not too bothered by it if it's inconvenient to structure the code that way

src/upload/form/mod.rs
GregBrimble: Without it, `script_path` is relative to the entrypoint. By joining `package_dir` with `script_path`, we get an absolute path, which is needed since wrangler actually executes in the project directory (parent).

So yes, it's needed, otherwise wrangler can't find that script to deploy.

xortive (approved):

ytbigvinny27 (commented):
🔜🔜🔜🔝

PR #1645, opened 2020-11-09
Print line and column numbers for exception thrown

nataliescottdavidson:
closes #1459

PR #1835, opened 2021-04-04
Revert "Print line and column numbers for exception thrown (#1645)"

### LINKED ISSUE #1826: wrangler dev should display unhandled exceptions
#### Describe the bug

wrangler tail displays unhandled exceptions.
wrangler dev does not. the docs imply that both do.

#### Reproduce the bug

wrangler dev
code throws an error
you get a single log line ending in '500 Internal Server Error'

#### Expected behavior

as above, but an additional log line containing the exception

#### Environment and versions

Fill out the following information about your environment.

* operating system: Windows 10
* output of `wrangler -V`: 1.15.0
* output of `node -v`: 14.15.4

Electroid:
This reverts commit 74a89f7c383bc22758cbe55096ce3016c5c319d7.

Closes #1826

This commit is causing `wrangler dev` to not show uncaught exceptions. Reverting `chrome-devtools-rs` was also necessary.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -294,9 +294,9 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "chrome-devtools-rs"
-version = "0.0.0-alpha.2"
+version = "0.0.0-alpha.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


xortive:
@Electroid any idea if the underlying issue is in the changes for the lastest version of `chrome-devtools-rs`, or the changes here? It would be nice to fix that so we could keep line and exception numbers.

Electroid:
I couldn't find the root cause. It's something we can revisit after next release, but though it's better to get some exceptions than none in the meantime.

nataliescottdavidson (approved):

PR #1837, opened 2021-04-06
Add config option to switch usage model to unbound

ObsidianMinor:


Electroid:
Is the plan to make `unbound` the default? How can we do that while remaining backwards compatible?

Electroid (approved):

PR #1846, opened 2021-04-12
Custom Builds Configuration UX changes

### LINKED ISSUE #1840: [custom-builds-rc] meaning of `main` and `module` in custom builds mode differs from webpack mode
With the webpack project type, the `main` key is assumed to be the entry point of the bundle for webpack, while in custom builds mode, we use it as the entry point for the compiled worker (as in, the main module in modules `upload_format`, or the file we actually upload for service-worker `upload_format`). This was done because it matches the javascript project type's behavior in non-custom builds mode, but it will cause confusion from users porting their project from the webpack project type.

Before leaving the RC period, we should resolve this inconsistency, as any changes to this behavior would be backwards-incompatible.

After https://github.com/cloudflare/wrangler/pull/1808 is merged, use the main key in `wrangler.toml` instead of `main` and `module` in package.json. This would force users to think about the value of `main`, and keep configuration for the project in a single place.

### LINKED ISSUE #1841: [custom-builds-rc] globbing for module types / redo `upload_format` configuration
Right now, the modules `upload-format` in custom builds mode has the following behavior:

- files in `upload_dir` are considered by the following rules
- `upload_dir` must exist, and must not be the project root.
- Files in `upload_exclude` are ignored by the following rules
- Only files in `upload_include` are considered by the following rules
- If `upload_include` and `upload_exclude` are both specified, `upload_include` is used as it is more specific.
- Files with known extensions are uploaded as a specific module type
- Files with that are left over after the above rules, are an error.

We should replace this behavior with file globs for each module type, and remove `upload_include` and `upload_exclude`. `upload_dir` will be kept, to avoid having to specify the path to the upload dir for each module. The default value for `upload_dir` will still be `./dist`.

Additionally, We should replace `upload_format` with an [internally tagged enum union](https://serde.rs/enum-representations.html#internally-tagged) called `upload`.

modules configuration will now look as follows

```toml
[build]
command = "npm r build"
[build.upload]
format = "modules"
dir = "dist"
# determines the prefix path for all of the module resolution rules below.
main = "./index.mjs"
# Must be a relative path inside of dist, including `./`
# The module type is determined by the rules below. Currently, we only allow ES modules to be the main module

# The default values for each module type are listed below.
esm = ["*.mjs"]
cjs = ["*.js", "*.cjs"]
text = ["*.txt"]
data = ["*.bin"]
compiled_wasm = ["*.wasm"] # NOT a default, see note below.
```

> Our wasm module implementation is non-standard, we should require you to specify that you are using `wasm_compiled`, so if we introduce standard wasm modules in the future, we are not stuck with a non-standard default. Most folks using wasm, will be using a template to get started, and this will be set inside of the template.

service-worker mode will look as follows

```toml
[build]
command = "npm r build"
upload.format = "service-worker"
```

### LINKED ISSUE #1843: [custom-builds-rc] module paths should include `./` at the beginning
We should include `./` at the beginning of module names generated by wrangler, which will make modules meant for consumption by the Workers Runtime, also resolve their imports when executed elsewhere, if they are compatible otherwise.

alternatively, we could make this configurable, and keep our current default.

xortive:
closes #1841 
closes #1840
closes #1843

Passes unit tests, still testing locally.


src/upload/form/project_assets.rs
```diff
@@ -81,64 +87,166 @@ impl ServiceWorkerAssets {
...
-            Self::Wasm => "application/wasm",
+            Self::CompiledWasm => "application/wasm",
             Self::Text => "text/plain",
             Self::Data => "application/octet-stream",
         }
     }
 }
 
+#[derive(Clone, Debug, Default, Deserialize, PartialEq, Serialize)]
+pub struct ModuleGlobs {
```
Electroid, line 119: Should there be a catch-all `exclude` glob list?

xortive: unknown files are ignored now by default, and you can exclude files that would match an individual rule with a negative glob rule I believe (I'll confirm this in a test.

src/settings/toml/builder.rs
```diff
@@ -15,21 +14,30 @@ pub struct Builder {
...
     #[serde(default = "project_root")]
     pub cwd: PathBuf,
-    #[serde(default = "upload_dir")]
-    pub upload_dir: PathBuf,
-    pub upload_format: ScriptFormat,
-    pub upload_include: Option<Vec<String>>,
-    pub upload_exclude: Option<Vec<String>>,
     #[serde(default = "watch_dir")]
     pub watch_dir: PathBuf,
+    pub upload: UploadFormat,
```
Electroid, line 19: If I understand right now, this will look like:

```toml
[builder]
# ...
[builder.upload]
# ...
```
Would it make sense to bubble it to the top-level? The files to upload aren't really part of the "build"
```toml
[builder]
# ...
[upload]
# ...
```

Electroid (approved):

src/settings/toml/builder.rs
xortive: I think we should do this someday (it makes much more sense), but right now it doesn't make sense, since this would imply `[upload]` being able to sit on it's own, which implies moving `watch_dir` to top level, and there are a lot of things that need to be thought about to get right.

PR #1851, opened 2021-04-14
add `--format` option, including default `json` and new `pretty`

caass:
It may make sense to have this just be a --pretty flag, but if we decide we want to add more formats in the future (like forwarding raw devtools protocol stuff) then it's better to have the --format flag. Plus, we can introduce a warning like "the default is set to json but make sure to be explicit in case we change it"

Example of using `--format pretty` here:

https://user-images.githubusercontent.com/25358963/114754564-b53e2900-9d26-11eb-8ff8-763ffe95cd63.mp4

### DIFF
--- a/src/tail/log_server.rs
+++ b/src/tail/log_server.rs
@@ -1,7 +1,10 @@
+use crate::terminal::{emoji, styles};
 use hyper::server::conn::AddrIncoming;
 use hyper::server::Builder;
 use hyper::service::{make_service_fn, service_fn};
 use hyper::{Body, Method, Request, Response, Server, StatusCode};
+use serde::{Deserialize, Serialize};
+use std::convert::TryInto;
 use tokio::sync::oneshot::Receiver;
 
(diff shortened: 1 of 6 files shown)


src/tail/log_server.rs
```diff
@@ -24,26 +27,55 @@ impl LogServer {
         }
     }
 
-    pub async fn run(self) -> Result<(), failure::Error> {
-        let service = make_service_fn(|_| async { Ok::<_, hyper::Error>(service_fn(print_logs)) });
+    pub async fn run(self, format: String) -> Result<(), failure::Error> {
```
xortive, line 30: should take an `&str` if you aren't returning an object that `format` gets moved into

caass: Hmm there's some issues with lifetimes around hyper stuff using borrows, I'm going to move ownership of the string into the `LogServer` struct.

Electroid (approved):

PR #1876, opened 2021-04-26
fix: use x86_64 arch for pre-built downloads on M1 devices

nilslice:
This PR forces the use of a pre-built x86_64 binary on a aarch64/arm64 Apple system. For M1 devices specifically, it will fix `wrangler generate`, and `wrangler build` for `type = "rust"` wrangler projects.

There is another semi-related PR that will fix some additional rust/wasm issues in our template repo here: https://github.com/cloudflare/rustwasm-worker-template/pull/21, which would be nice to have merged along with this.

### DIFF
--- a/npm/binary.js
+++ b/npm/binary.js
@@ -12,11 +12,9 @@ const getPlatform = () => {
   if (type === "Linux" && arch === "x64") {
     return "x86_64-unknown-linux-musl";
   }
-  if (type === "Darwin" && arch === "x64") {
+  if (type === "Darwin" && (arch === "x64" || arch == "arm64")) {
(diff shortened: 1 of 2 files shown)


nilslice:
It's worth noting that this fix does not change any behavior on any of our previously supported platforms. And for posterity, it relies on the target M1 / aarch64/arm64 Apple device to have Rosetta 2 installed. This should already be the case, but if not, it can be installed by running:
```
/usr/sbin/softwareupdate --install-rosetta
```

Electroid (approved):

nataliescottdavidson (approved):

YoutacRSVARs (approved):

PR #1880, opened 2021-04-27
[WIP] Update tokio ecosystem

caass:
We need to upgrade to tokio 1.0 to get unblocked on using the latest version of [`cloudflare-rs`](https://github.com/cloudflare/cloudflare-rs/commit/80d249c0cc8241e1043a59c1fb6e23979ce62cae#diff-bc2a915105ef2a549d4fadc51c6282c15e76a32026e4eeb86e78e10a29951a9e).

This PR updates tokio and related libs to 1.0 so we can continue using `cloudflare-rs`. This also unblocks tail v2.

related: #1881

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ binary-install = "0.0.3-alpha.1"
 chrome-devtools-rs = { version = "=0.0.0-alpha.1", features = ["color"] }
 chrono = "0.4.19"
 clap = "2.33.3"
-cloudflare = "0.6.6"
+cloudflare = { git = "https://github.com/cloudflare/cloudflare-rs", branch = "cass/fix-tail-response" }
 config = "0.10.1"
 console = "0.13.0"
 dirs = "3.0.1"
@@ -31,8 +31,8 @@ fs2 = "0.4.3"
 futures-util = "0.3"
 globset = "0.4.6"
 http = "0.2.1"
-hyper = "0.13.9"
-hyper-rustls = "0.21.0"
+hyper = { version = "0.14.7", features = ["http2", "server", "runtime"] }
+hyper-rustls = "0.22.1"
 ignore = "0.4.17"
 indicatif = "0.15.0"
 lazy_static = "1.4.0"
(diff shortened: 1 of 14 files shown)


Cargo.toml
```diff
@@ -18,7 +18,7 @@ binary-install = "0.0.3-alpha.1"
 chrome-devtools-rs = { version = "=0.0.0-alpha.1", features = ["color"] }
 chrono = "0.4.19"
 clap = "2.33.3"
-cloudflare = "0.6.6"
+cloudflare = { git = "https://github.com/cloudflare/cloudflare-rs", branch = "cass/fix-tail-response" }
```
nilslice, line 21: 😄  safe to assume this will be merged upstream and changed here prior to us releasing?

src/kv/bulk.rs
```diff
@@ -33,6 +33,7 @@ fn bulk_api_client(user: &GlobalUser) -> Result<HttpApiClient, failure::Error> {
         config,
         Environment::Production,
     )
+    .map_err(|e| failure::format_err!("{}", e))
```
nilslice, line 36: worth considering some prefix to this error here? `[bulk api]` or something that gives a hint in case the trace isn't included (I'm unsure what `failure` will provide without specifically asking for the backtrace)? 

not saying its a must, but since you're already in here...

Cargo.toml
nataliescottdavidson: @caass , you should be able to remove this and just go with the latest release `0.8.0`. That bugfix doesn't have to do with these changes so they don't need to go through together.

src/kv/bulk.rs
caass: Hmm honestly the better solution is to move to `anyhow`, since that's what `cloudflare-rs` uses. I'll try swapping that out and see what the breakage is like.

nataliescottdavidson (commented):
`publish` and `dev` are working, but`tail` hangs endlessly on `This may take a few seconds...`. I can investigate a bit as I was just working on tail. Thanks for this @caass , getting us out of dependency hell!

src/kv/bulk.rs
caass: @nilslice see https://github.com/cloudflare/wrangler/pull/1881

caass:
> `publish` and `dev` are working, but`tail` hangs endlessly on `This may take a few seconds...`. I can investigate a bit as I was just working on tail. Thanks for this @caass , getting us out of dependency hell!

For anyone else reading (i talked with @nataliescottdavidson offline), I'm pretty sure the issue has to do with message passing between different tasks, that'll be the next commit, i think.

Cargo.toml
caass: @nataliescottdavidson i'm actually just about to release 0.8.1 with these changes 🙈

caass:
Closing in favor of #1886

PR #1881, opened 2021-04-27
failure --> anyhow

caass:
Follow up to #1880 

Switch from deprecated `failure` to much better and cooler `anyhow`. seriously [read this](https://github.com/dtolnay/case-studies/blob/master/autoref-specialization/README.md)

This makes error handling with `cloudflare-rs` easier, allowing us to preserve stack trace
information from any errors we encounter with the API.

### DIFF
--- a/src/installer.rs
+++ b/src/installer.rs
@@ -22,12 +22,12 @@ use std::io;
 use std::path::Path;
 use std::process;
 
-use failure::{self, bail, ResultExt};
+use anyhow::{anyhow, bail, Result};
 
 pub fn install() -> ! {
     if let Err(e) = do_install() {
         eprintln!("{}", e);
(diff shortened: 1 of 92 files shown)


caass:
wait how do i re-open this

src/installer.rs
```diff
@@ -45,7 +45,7 @@ pub fn install() -> ! {
     process::exit(0);
```
nilslice, line 45: what do you think about just `bail!`ing here and returning a Result from anyhow instead of the never return type? would need to update the callers, but this is an error I think we'd like to be able to diagnose from the automated user reporting I'm working on.

nilslice (approved):
nice work, @caass! great improvement

PR #1888, opened 2021-04-29
feat: capture panics and generate error report

nilslice:
Partial PR to add support for wrangler error reporting. Remaining parts are to add the CLI command:
`wrangler report <file-name>` 

- <file-name> is an optional name of the log file, e.g. `1619728882567.log` so a user can specify which log to submit. 
- otherwise, the most recent error log will be used and submitted to the Wrangler team for debugging.


src/commands/report.rs
```diff
@@ -0,0 +1,56 @@
...
+                if code == 409 {
+                    return anyhow::anyhow!(
+                        "we have already received this report, thank you for submitting!"
+                    );
+                }
+
+                anyhow::format_err!(
+                    "submission failed! please verify your credentials and try again. (status: {})",
+                    code
+                )
```
caass, line 50: this can just be `anyhow::anyhow!`

src/commands/report.rs
```diff
@@ -0,0 +1,56 @@
...
+            }
+            ApiFailure::Invalid(e) => {
+                anyhow::format_err!("submission failed! please try again. (request error: {}", e)
```
caass, line 53: this as well, `format_err` is just for compatibility with `failure`.

src/commands/report.rs
```diff
@@ -0,0 +1,56 @@
...
+}
+
+pub fn run(log: Option<&str>) -> Result<()> {
+    let user = settings::global_user::GlobalUser::new()?;
+    let report = reporter::read_log(log)?;
+    let client = http::cf_v4_client(&user)?;
+    client
+        .request(&ErrorReport(report))
+        .map(|_| ())
+        .map_err(|e| match e {
```
caass, line 37: This could be better written as 
```rust
if let Some(e) = client
  .request(&ErrorReport(report))
  .err() {
    match e {
      ...
    }
} else {
    Ok(())
}
```
if you wanna discard the success result. Or some variation of that -- `Result::err()` is designed for that case.

src/commands/report.rs
caass, line 50: To capture this error better, you can use [`Context`](https://docs.rs/anyhow/1.0.40/anyhow/trait.Context.html) to provide a custom error message and then just use `anyhow!(e)` to have anyhow keep track of the original error's source, stack trace, etc.

src/main.rs
```diff
@@ -1148,6 +1161,10 @@ fn run() -> Result<()> {
         commands::tail::start(&target, &user, format, tunnel_port, metrics_port, verbose)?;
     } else if matches.subcommand_matches("login").is_some() {
         commands::login::run()?;
+    } else if let Some(matches) = matches.subcommand_matches("report") {
+        commands::report::run(matches.value_of("log")).map(|_| {
+            eprintln!("Report submission sucessful. Thank you!");
+        })?;
```
caass, line 1167: You should use [`and_then`](https://doc.rust-lang.org/std/result/enum.Result.html#method.and_then) here, I think.

src/reporter/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+        project_info: load_project_info(),
+        args: args().collect::<Vec<_>>(),
+        panic: panic_payload(panic_info),
+        location: None,
+        backtrace: useful_frames(),
+    };
+
+    if let Some(info) = panic_info {
+        if let Some(loc) = info.location() {
+            report.location = Some(format!("{}:{}:{}", loc.file(), loc.line(), loc.column()));
```
caass, line 69: can you change the format from `<file>:<line>:<column>` to something easier to parse, like `<file>, Line <line> Column <column>`? If it doesn't have to be a specific format, it'd be nice to have something more intuitive.

src/reporter/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+        let entry = entry?;
+        let md = fs::metadata(entry.path())?;
+        if md.is_file() {
+            files.push(entry.path())
+        }
+    }
+    files.sort();
+
+    if let Some(f) = files.last() {
+        return serde_json::from_reader(BufReader::new(File::open(f)?)).map_err(|e| e.into());
```
caass, line 136: Can you separate this into a few lines to make it easier to read?

src/reporter/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+
+        if let Some(usage_model) = manifest.usage_model {
+            project_info.insert("usage_model".into(), usage_model.as_ref().into());
+        }
+    }
+
+    project_info
+}
+
+// removes frames before wrangler takes over at the panic, reduces noise
```
caass, line 225: 🙏🏻 🙏🏻 🙏🏻

src/reporter/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+            Some(s) => Some(s.to_string()),
+            None => None,
+        };
+    }
+
+    None
+}
+
+fn err_exit<S: AsRef<str>>(msg: S, code: i32) {
+    eprintln!("wrangler panic handle error: {}", msg.as_ref());
```
caass, line 275: Can you make this error message friendlier, like "Wrangler encountered an error while attempting to log the above error." Or something similar?

src/reporter/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+        return match info.payload().downcast_ref::<&str>() {
+            Some(s) => Some(s.to_string()),
+            None => None,
+        };
+    }
+
+    None
+}
+
+fn err_exit<S: AsRef<str>>(msg: S, code: i32) {
```
caass, line 274: You should make this return `!` type, which will avoid peppering in the extra `unreachable!`s in the code.

caass (changes requested):
Looks great! Left a few minor stylistic comments, but the content is good :) thanks!

src/commands/report.rs
caass: You should look at the anyhow docs for the difference between passing a `std::error::Error` implementor vs a `std::fmt::Display` implementor to `anyhow!`, passing an `Error` preserves its original stack trace, etc. I know we're not capturing this right now, but eventually we might.

src/reporter/mod.rs
caass: Line 136, I mean

src/commands/report.rs
nilslice: good catch! didn't realize that the `anyhow!` macro took format args

src/commands/report.rs
nilslice: yea, i tend to prefer mapping when I have to Err() wrap the various internal error values, but i suppose it is a bit more elegant when the Ok is discarded. updated!

src/commands/report.rs
nilslice: good call. going to leave this as-is for now, since the errors it can use in the context _might_ only come from the `ApiErrors` struct, 
```rust
#[derive(Deserialize, Debug, Default)]
pub struct ApiErrors {
    #[serde(flatten)]
    pub other: HashMap<String, JValue>,
    pub errors: Vec<ApiError>,
}
```

and there's just a whole lot there that i'm not trying to jump into right now 🤣 .. there's an `Error` impl on it, but idk

I'll leave a TODO with some detail to look into it later.

src/main.rs
nilslice: in this case, i don't return any result from the combinator - i just want to specify the Ok behavior and then bubble up the Err. `and_then` could prob work too, by returning `Ok(())` after the print? idk I this is alright as-is.

src/reporter/mod.rs
nilslice: yea for sure.  just went with the file:line:column as the standard for how system tools specify location. but we're not integrating this output into an IDE or something, so it's probably better to be explicit for humans.

src/reporter/mod.rs
nilslice: nice, TIL 🥇

src/main.rs
caass: Yeah it's fine, I was just in nitpick mode :)

src/commands/report.rs
caass: Thanks! And you're right, these are both totally fine, just a minor nit :)

src/reporter/mod.rs
caass: lol it's probably just me never having seen that format before then 😅 

thank!

src/reporter/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+        }
+    } else {
+        err_exit(format!("Wrangler encountered an unrecoverable error and failed to write the report: \n{:#?}", report), 1);
+    }
+
+    // print message to user with note about the crash and how to report it using the command
+    // `wrangler report --log=<filename.log>`
+    eprintln!(
+        r#"
+Oops! Wrangler hit a snag... please help Cloudflare debug the issue by submitting the generated error report ({})
```
Electroid, line 113:
```suggestion
Oops! wrangler encountered an error... please help Cloudflare debug this issue by submitting an error report ({})
```

src/reporter/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+
+To submit this error report to the Wrangler team now, run:
```
Electroid, line 115:
```suggestion
To submit this error report to Cloudflare, run:
```

src/main.rs
```diff
@@ -656,6 +659,16 @@ fn run() -> Result<()> {
         .subcommand(
             SubCommand::with_name("login")
                 .about(&*format!("{} Authenticate Wrangler with your Cloudflare username and password", emoji::UNLOCKED)))
+        .subcommand(
+            SubCommand::with_name("report")
+                .about(&*format!("{} Report an error caught by wrangler to the Cloudflare team", emoji::BUG))
```
Electroid, line 664:
```suggestion
                .about(&*format!("{} Report an error caught by wrangler to the Cloudflare", emoji::BUG))
```

src/reporter/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+            std::fs::File::create(wrangler_error_dir.join(format!("{}.log", report.timestamp_ms)))
+                .unwrap_or_else(|_| {
+                    err_exit("failed to create report file", 1);
+                });
+
+        if file.write_all(data.as_bytes()).is_err() {
+            err_exit("failed to write report", 1);
+        }
+    } else {
+        err_exit(format!("Wrangler encountered an unrecoverable error and failed to write the report: \n{:#?}", report), 1);
```
Electroid, line 106:
```suggestion
        err_exit(format!("wrangler encountered an unrecoverable error and failed to write the report: \n{:#?}", report), 1);
```

src/reporter/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+            ));
+        }
+    }
+
+    match std::time::SystemTime::now().duration_since(std::time::UNIX_EPOCH) {
+        Ok(ts) => {
+            report.timestamp_ms = ts.as_millis();
+        }
+        Err(e) => {
+            err_exit(format!("system time failure: {}", e), 1);
```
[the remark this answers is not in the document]
nilslice: it's _really_ unlikely that this will happen. if it does happen there's something else really wrong with the users system. 

using a value like `-1` would be a little awkward to detect, because it will always be at the opposite end of the sorted list from where we're looking (we want the biggest number == most recent created). the other issue is that as a single value, the file contents will be overwritten if another panic occurs and we'd lose the previous data. 

let's hold off on this for now and see if anyone has issues.

src/reporter/mod.rs
nilslice: no worries :) it's a sneaky little detail, but even rustc uses it in error messages, note **src/reporter/mod.rs:106:14**:

```
error: expected one of `.`, `;`, `?`, `}`, or an operator, found `::`
   --> src/reporter/mod.rs:106:14
    |
106 |         "o_0"::err_exit(format!("wrangler encountered an unrecoverable error and failed to write the report: \n{:#?}", report), 1);
    |              ^^ expected one of `.`, `;`, `?`, `}`, or an operator

error: aborting due to previous error
```

I like the more human-friendly version as you suggested for our error reporting anyway.

PR #1900, opened 2021-05-02
Updated cloudflared download link

arunesh90:
Hello 🙂

The commit included with this PR updates the download link for cloudflared, as the previous link is no longer current and is now a redirect to https://developers.cloudflare.com/cloudflare-one/connections/connect-apps instead of directly to the installation/download page at https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/install-and-setup/installation


caass (approved):
Thanks!

Issue #1859, opened 2021-04-17
Ambiguous error when uploading ES modules without the entitlement


frandiox:
#### Describe the bug

When running `wrangler dev` or `wrangler publish`, it throws 403 Forbidden error.

#### Reproduce the bug

A list of steps to reproduce the bug.

1. Clone https://github.com/cloudflare/modules-rollup-esm (official example)
2. Install dependencies, modify package.json `name`, add `account_id` to `wrangler.toml`.
3. Run `wrangler dev`
4. See error:

```bash
$> wrangler dev
...
src/index.mjs → dist/index.mjs...
created dist/index.mjs in 223ms
Error: HTTP status client error (403 Forbidden) for url (https://api.cloudflare.com/client/v4/accounts/<account-id>/workers/scripts/custom-build-test/edge-preview)
```

```bash
$> wrangler publish
...
src/index.mjs → dist/index.mjs...
created dist/index.mjs in 214ms
✨  Build completed successfully!
Error: Something went wrong! Status: 403 Forbidden, Details {
  "result": null,
  "success": false,
  "errors": [
    {
      "code": 10015,
      "message": "workers.api.error.not_entitled"
    }
  ],
  "messages": []
}
```

#### Expected behavior

Should build and deploy, but it fails in the deploy.

#### Environment and versions

Fill out the following information about your environment.

* operating system: MacOS Catalina 10.15.7
* output of `wrangler -V`: `wrangler 1.16.0`
* output of `node -v`: `v14.16.1`
* content of `wrangler.toml`:

It's literally like the example (hiding account id here):

```
name = "custom-build-test"
# type = "javascript" is required to use the `[build]` section
type = "javascript"
account_id = "<account_id>"
workers_dev = true
route = ""
zone_id = ""

[build]
command = "npm run build"
upload.format = "modules"
upload.main = "./index.mjs"

```

Electroid:
Thanks for reporting this. To get around this, you should signup your account for the Durable Objects open beta. Go to the Workers section of the dashboard, and you'll see a tab for "Durable Objects"

Soon, we will remove this restriction, but before then we should clarify the error message. So I'll keep this open

frandiox:
@Electroid related question, it says `Durable Objects is only available while you are on the Workers Bundled plan.` and, even though I'm on paid plan, I'm basically making open source and I assume most people don't have a paid account.
Will the custom builds (without DO) be available with free plans after the restriction you mention is removed?

[...]

PR #1921, opened 2021-05-13
feat: clarify error messages around durable objects beta

nilslice:
Disambiguates error message described in #1859, adds more clarity around other places where Durable Object usage in beta may conflict with `wrangler` functionality.

### DIFF
--- a/src/deploy/schedule.rs
+++ b/src/deploy/schedule.rs
@@ -1,5 +1,5 @@
-use crate::http;
 use crate::settings::global_user::GlobalUser;
+use crate::{http, util::TEMP_NOTICE_ES_MODULES_DO_BETA};
 
 use anyhow::Result;
 
@@ -37,12 +37,14 @@ impl ScheduleTarget {
             .body(build_schedules_request(&self.crons))
             .send()?;
 
-        if !res.status().is_success() {
-            anyhow::bail!(
-                "Something went wrong! Status: {}, Details {}",
-                res.status(),
-                res.text()?
-            )
+        let status = res.status();
+        if !status.is_success() {
+            let text = res.text()?;
+            if text.contains("workers.api.error.not_entitled") {
+                anyhow::bail!(TEMP_NOTICE_ES_MODULES_DO_BETA)
+            }
+
+            anyhow::bail!("Something went wrong! Status: {}, Details {}", status, text)
         }
 
         Ok(self.crons.clone())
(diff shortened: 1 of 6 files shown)


src/deploy/schedule.rs
```diff
@@ -37,12 +37,14 @@ impl ScheduleTarget {
...
-        if !res.status().is_success() {
-            anyhow::bail!(
-                "Something went wrong! Status: {}, Details {}",
-                res.status(),
-                res.text()?
-            )
+        let status = res.status();
+        if !status.is_success() {
+            let text = res.text()?;
+            if text.contains("workers.api.error.not_entitled") {
```
koeninger, line 43: Is durable objects the only possible reason for not_entitled?

nilslice: Probably not, but it's a temp quick fix until you get DO's to GA.

koeninger: Then is there a reason not to show the original error plus a "If you are trying to use Modules, you need to opt in..."?

As it stands someone's probably going to be more confused, not less, if it has nothing to do with modules or DO

nilslice: I'd think there would be a better API error, actually indicating the accurate entitlement missing. But I'm assuming you're not ready to commit to that until GA..

The original error was "Something went wrong!" + the ambiguous `workers.api.error.not_entitled`.

src/util.rs
```diff
@@ -8,3 +8,5 @@ impl<'a, 'b> ApplyToApp<'a, 'b> for App<'a, 'b> {
         f(self)
     }
 }
+
+pub const TEMP_NOTICE_ES_MODULES_DO_BETA: &str = "While Durable Objects are in Beta, ES Modules are limited to accounts which have opted-in to the Beta. You may do so by following the instructions here: https://developers.cloudflare.com/workers/learning/using-durable-objects";
```
Electroid, line 12:
```suggestion
pub const TEMP_NOTICE_ES_MODULES_DO_BETA: &str = "Your account does not have permission to do this! While Durable Objects are in Beta, the modules format is limited to accounts which have opted-in to the Beta. You may do so by following the instructions here: https://developers.cloudflare.com/workers/learning/using-durable-objects";
```

Electroid (changes requested):

Electroid (approved):

PR #1932, opened 2021-05-25
Reduce cognitive complexity in main

ObsidianMinor:
This replaces our massive clap App with a much simpler Cli struct that
has all the same information in an easier to understand and modify
format (types). Because everything is now typed (and because we've moved
clap::App creation out of main), we can further reduce the cognitive
complexity of main by removing lots of parameter checking and parsing,
removing that pesky clippy lint attribute.

Also it removes emoji in all the help messages. But they don't always work
anyway (Windows) and prevent us from doing this *necessary* refactor,
so meh.


ObsidianMinor:
I'm currently going over and verifying the API hasn't changed at all.

src/main.rs
```diff
@@ -63,1116 +66,895 @@ fn main() -> Result<()> {
...
-                Some(ref pt) => match pt {
-                    TargetType::Rust => "https://github.com/cloudflare/rustwasm-worker-template",
-                    _ => default_template,
+    /// Transfer all durable objects associated with a class in another script to a class in
+    /// this script
+    #[structopt(name = "transfer-class", long, number_of_values = 3)]
+    transfer_class: Vec<String>,
+}
+
+impl AdhocMigration {
```
nilslice, line 306: Is it necessary for this to be in `main.rs` ?

src/main.rs
```diff
@@ -63,1116 +66,895 @@ fn main() -> Result<()> {
...
+                eprintln!("Report submission sucessful. Thank you!");
+            })?;
+        }
     }
+
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
```
nilslice, line 893: Similarly, these should probably all go in a separate module along with the AdhocMigration impl and other non-CLI things. What do you think?

src/main.rs
```diff
@@ -63,1116 +66,895 @@ fn main() -> Result<()> {
...
+    /// Toggle verbose output (when applicable)
+    #[structopt(name = "verbose", long = "verbose", global = true)]
+    pub verbose: bool,
+
+    /// Path to configuration file.
+    #[structopt(
+        name = "config",
+        short = "c",
+        long = "config",
+        default_value = "./wrangler.toml",
```
nilslice, line 87: Could you change "./wrangler.toml" to "wrangler.toml" in the `default_value` attr key? I can't push to your fork.

nilslice:
LGTM - each subcommand works as expected and I didn't catch any regressions, nice work! 

Just left some minor changes, and q's on what we probably ought to move outside main.

nilslice (changes requested):
LGTM - each subcommand works as expected and I didn't catch any regressions, nice work!

Just left some minor changes, and q's on what we probably ought to move outside main.

src/main.rs
ObsidianMinor: Yes (it's processing command line input so I'd argue it belongs in main)

nilslice: ok, I see. good to merge, I may do a little cleanup to get more of this out of main.

nilslice:
> I'm currently going over and verifying the API hasn't changed at all.

@ObsidianMinor, can you confirm you saw no changes?

nilslice (approved):
Ok, confirmed offline with @ObsidianMinor that the API is consistent, and have observed the same with my testing. ✅

PR #1934, opened 2021-05-25
Handle String panic payloads when generating reports

ObsidianMinor:
Standard panics will only produce &str and String, but we were only handling &str, so this adds handling for String.


nilslice (approved):
Thanks!

PR #1941, opened 2021-05-28
Add `wrangler zone` command

jb3:
## What?

Adds a `wrangler zone` command to find the ID of a zone straight from the wrangler CLI.

Screenshot 2021-05-28 at 22 36 25

## Why?

When generating a new project with wrangler you are directed to the dashboard for the final step to find the Zone ID for the domain you want to add the Worker to. This PR introduces a command which lets you remain in wrangler for the entire creation ➡️ publish to custom route flow.

## Caveats

- I could not find a way to do permission checking. When a token does not have permission the Cloudflare API just returns no entries, to counter this I've added a message when a zone is not found to check spelling but also check wrangler permissions.
- This may require a change in how tokens are created with `wrangler login`, they should include the `Zone.Zone` permission to be able to read zone information.

## ToDo

- [x] Document the new command in the README.md
- [ ] Document the new command in the [commands](https://developers.cloudflare.com/workers/cli-wrangler/commands) documentation page.
- [ ] Update references in wrangler itself to reference the new zone command instead of directing people to the dashboard (optional, this might not be a good idea if tokens are not retroactively updated, as it will then be a more strenous task than just going to fetch the zone ID).

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,21 @@ $ wrangler publish
   - `ip`: ip to listen on. defaults to localhost
   - `port`: port to listen on. defaults to 8787
 
+### 🌎 `zone`
+
+  Fetch the Cloudflare Account & Zone ID of the supplied domain name.
+  
+  You can use the returned ID as the `zone_id` field of your `wrangler.toml` configuration file to publish a Worker on a custom domain.
+
+  ```bash
+  $ wrangler zone example.com
+  Zone information for example.com
+
+  Zone ID: e405e53ee0fb1190c104316d5d306a19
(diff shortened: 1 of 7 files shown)


Electroid:
Thanks for working on this @jb3, this looks pretty slick! We had a chat about this among the team and while we like it,  I think there's a more interesting question at play here: why should we require the `zone_id` in the first place? There's a simple answer, for now, which is our API requires it; but we could change that :)

In that future world, you would just define the routes and we would figure out which zones (yep, multi-zone support) you want to publish to. If we were to add that new API, does the need for this go away?

jb3:
Very interesting idea! Multi-zone support would be fantastic. 😄

PR #1946, opened 2021-06-01
chore: add more detail to error when encountering a bad package.json

nilslice:
Helps clarify error encountered e.g. `could not parse package.json` with possible solution to include `main` or `module` keys in the file.

### DIFF
--- a/src/upload/package.rs
+++ b/src/upload/package.rs
@@ -55,10 +55,11 @@ impl Package {
 
         let package_json: String = fs::read_to_string(manifest_path.clone())?.parse()?;
 
-        serde_json::from_str::<Package>(&package_json).map_err(|_| {
+        serde_json::from_str::<Package>(&package_json).map_err(|e| {
             anyhow::anyhow!(
-                "could not parse {}, may have invalid or missing `main` or `module` keys",
-                manifest_path.display()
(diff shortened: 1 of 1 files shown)


nilslice:
Converting this to draft, after realizing the error reporter will swallow the panic and not share any helpful detail with the end user 🤦 . Updating!

src/upload/package.rs
```diff
@@ -54,9 +54,13 @@ impl Package {
...
 
         let package_json: String = fs::read_to_string(manifest_path.clone())?.parse()?;
-        let package: Package = serde_json::from_str(&package_json)
-            .unwrap_or_else(|_| panic!("could not parse {}", manifest_path.display()));
 
-        Ok(package)
+        serde_json::from_str::<Package>(&package_json).map_err(|e| {
+            anyhow::anyhow!(
+                "could not parse {}, may have invalid or missing `main` or `module` keys: {}",
+                manifest_path.display(),
```
caass, line 61: Could you add an explanation, just a quick thing like "both keys are necessary for module format"? you can copy the messages from https://github.com/cloudflare/wrangler/blob/master/src/upload/package.rs lines 18 and 32

caass (approved):
Just one minor nit to clarify the error message (why it's necessary to have both module and main keys)

Issue #331, opened 2019-07-18
WIP RFC: Automatically configure account_id and zone_id


xortive:
**TL;DR - if config has perms for only one account id / zone id, fill it in automatically on generate/init**

- The configured route in wrangler.toml _must_ include a zone name for a zoned publish
- Zoneless publish don't need a `zone_id` or `route`, just `account_id` and `script_name`
- Global config has the global api key and email

We have all the ingredients to infer the `account_id` and `zone_id` instead of requiring the user to configure it manually using https://api.cloudflare.com/#zone-properties and https://api.cloudflare.com/#accounts-list-accounts.

Open questions:
- How does this look from a developer experience perspective? when should we fetch these parameters and how should we provide choice between them
- Is this a good idea?

granjef3:
Made this WIP while we define those open questions

EverlastingBugstopper:
This is a nice to have feature that we could implement for any single account profiles. Cloudflare allows a single profile to be associated with multiple accounts (or collections of zones), so inferring account id only works for profiles with a single account. The same goes for zone id - there can be multiple zone ids per account so you could only infer them if an account had a single associated zone. If I had to guess, the majority of folks using Wrangler would benefit from this.

ashleymichal:
i think this  should  be pretty easy to do with the work that's already been done for `whoami`. if the user already has their config set when we go to generate a  manifest, we can request the relevant account id; if there is exactly one we will fill it in, else we can message next steps.

Electroid:
@xortive Would `wrangler login` make this outdated?

rita3ko:
@Electroid i actually think we can make this better with `wrangler login`. once you're logged in, when you `wrangler init` or `wrangler generate` a project we could let you pick an account and then zone interactively and prefill it for you in `wrangler.toml`

PR #1966, opened 2021-06-23
Don't require an account_id in wrangler.toml

jyn514:
Helps with https://github.com/cloudflare/wrangler/issues/331. I recommend reviewing this commit-by-commit.

Note that unlike the suggestion in the original issue, this actually never writes to the wrangler.toml at all, which will hopefully make it easier to share Worker projects while still being able to publish them. Instead it will load the account ID from the internet if it's not present, which required a fair bit of refactoring to make it lazily loaded.

1. If `account_id` is present in `wrangler.toml`, use that.
2. If there is only one ID for the current authentication token, use that.
3. If there are multiple IDs, give an error and prompt the user to add the id to wrangler.toml.

Here's an example error:

```
👀  You have multiple accounts.
🕵️  You can copy your account_id below
+--------------------------+----------------------------------+
| Account Name             | Account ID                       |
+--------------------------+----------------------------------+
| Cloudflare Community Jam | 6174b0f52802f24a9f52e7015282898c |
+--------------------------+----------------------------------+
| Edge Workers             | 615f1f0479e7014f0bebcd10d379f10e |
+--------------------------+----------------------------------+
Error: field `account_id` is required
```

I'm going to save adding support for auto-editing wrangler.toml for a follow-up, this PR is already a little big.


src/settings/toml/manifest.rs
```diff
@@ -512,6 +513,96 @@ impl Manifest {
...
+}
+
+impl<'de> Deserialize<'de> for LazyAccountId {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        Option::deserialize(deserializer).map(Into::into)
+    }
+}
```
jyn514, line 535: This seems like something I should be able to do with the serde derive macros, but I'm not sure what the syntax is.

src/settings/toml/manifest.rs
```diff
@@ -288,21 +289,21 @@ impl Manifest {
...
+        self.account_id.load().map(String::from)
+        // if let Some(id) = result.0 {
+        //     Ok(id)
+        // } else {
+        //     let mut msg = "Your configuration file is missing an account_id field".to_string();
+        //     if let Some(environment_name) = environment_name {
+        //         msg.push_str(&format!(" in [env.{}]", environment_name));
+        //     }
+        //     anyhow::bail!("{}", &msg)
+        // }
```
jyn514, line 306: I wasn't sure whether or not to try and keep this context in the new prompt. It seems unfortunate to lose the suggestion to add the ID to the environment file; maybe I could check `account.maybe_load.is_none()` and still give the message anyway? We don't know anymore if the user *intended* for the ID to be in the environment file or not.

jyn514:
> Find a way to get rid of the Error: text at the end. I think I can do that by adding a custom error type with its own Display impl, and returning that.

No, this hits the same issue as before - `display_account_id_maybe` prints directly to stdout. I guess I can change that too? It feels pretty intrusive, I'd have to copy the emojis from `StdOut`.

Another alternative is to call `process::exit`, but that feels like a heavy hammer.

src/settings/toml/manifest.rs
```diff
@@ -512,6 +516,96 @@ impl Manifest {
...
+        self.0.get()
+    }
+
+    /// If `account_id` can be inferred automatically, do so;
+    /// otherwise, return `None`.
+    ///
+    /// Note that *unlike* `load`, this will never prompt the user or warn.
+    pub(crate) fn maybe_load(&self) -> Option<String> {
+        use crate::commands::whoami::fetch_accounts;
+        use crate::settings::global_user::GlobalUser;
```
nilslice, line 564: Do you mind pulling these use statements (along with the others in various places) up to the top of the file, so it's consistent with the rest of the code? it also is nice to avoid duplication.

src/settings/toml/manifest.rs
nilslice: I think it could be if not for the `OnceCell` wrapper. I _think_ serde supports newtype / tuple structs via derive.

src/settings/toml/manifest.rs
nataliescottdavidson: We can get rid of it. No need to prompt them if we already have it.

nataliescottdavidson (approved):
Thanks @jyn514 !

jyn514:
@nataliescottdavidson this doesn't affect `zone_id`, I wasn't sure what the long-term plan for that is. https://github.com/cloudflare/wrangler/pull/1941#issuecomment-853517522 mentions it should go away but I haven't looked into it.

nataliescottdavidson:
> @nataliescottdavidson this doesn't affect `zone_id`, I wasn't sure what the long-term plan for that is. [#1941 (comment)](https://github.com/cloudflare/wrangler/pull/1941#issuecomment-853517522) mentions it should go away but I haven't looked into it.

Yeah, that's on our TODO list, doesn't need to be in this change though.

jyn514:
Some more things I need to fix:
- `wrangler generate` shouldn't tell you to add an account ID unless you have multiple:
```
> wrangler generate
 Creating project called `worker`...
 Done! New project created /home/jnelson/src/test-project/worker/worker
🕵️  You can find your zone_id in the right sidebar of a zone's overview tab at https://dash.cloudflare.com
🕵️  You can copy your account_id below
+----------------------------+----------------------------------+
| Account Name               | Account ID                       |
+----------------------------+----------------------------------+
| Jyn514@gmail.com's Account | e1706d218241c1230155b4f91b3218af |
+----------------------------+----------------------------------+
🕵️  You will need to update the following fields in the created wrangler.toml file before continuing:
- account_id
```
- `wrangler preview` should use the account ID instead of defaulting to an empty string (this is a pre-existing bug):
```
> wrangler preview
Error: Something went wrong with the request to Cloudflare...
Could not route to /accounts/workers/scripts/worker/preview, perhaps your object identifier is invalid? [API code: 7003]
No route for that URI [API code: 7000]
```
```

**Comment by jyn514** [comment]
> wrangler preview should require an account ID instead of defaulting to an empty string (this is a pre-existing bug):

Hmm, I couldn't replicate this error. Maybe it was a transient error on the server side?

**Comment by jyn514** [comment]
I managed to replicate it - the issue is that this mishandles `account_id = ""` and assumes it's a valid ID. That's very unfortunate, because everyone with a fresh `wrangler generate` project has `account_id = ""`. I can change `generate` to stop doing that, but it won't fix it for the projects that have already been created.

**Comment by jyn514** [comment]
Hmm, with that in mind I wonder if it still makes sense to use `Option<String>`. It might make sense to keep using `""` as the "null" value that means not loaded and change all the `None` values to `""`.

File: src/settings/toml/manifest.rs (hunk 1 above)
Comment (jyn514) [reply src/settings/toml/manifest.rs]: This a moot point now that the deserialize impl handles the empty string, I don't see any way that could be a derive.

**Comment by nataliescottdavidson** [comment]
> I managed to replicate it - the issue is that this mishandles `account_id = ""` and assumes it's a valid ID. That's very unfortunate, because everyone with a fresh `wrangler generate` project has `account_id = ""`. I can change `generate` to stop doing that, but it won't fix it for the projects that have already been created.

> Hmm, with that in mind I wonder if it still makes sense to use Option<String>. It might make sense to keep using "" as the "null" value that means not loaded and change all the None values to "".

I don't think we want to go back to "" for null, your implementation is much nicer. Could we add a check in [here](https://github.com/cloudflare/wrangler/pull/1966/files#diff-c82f6d0441c7f8ad03e619c809138bbb7e0ca631652e6ad8c3c2edf43c5fbe74R549) that errors / prompts account id if the account id in `wrangler.toml` is ""?

**Comment by jyn514** [comment]
@nataliescottdavidson I changed it to treat the empty string just as if it were unset, which should hopefully prompt the user less: https://github.com/cloudflare/wrangler/pull/1966/files#diff-c82f6d0441c7f8ad03e619c809138bbb7e0ca631652e6ad8c3c2edf43c5fbe74R527-R529. I can change it to give a hard error if you like but I'm not sure how useful it would.

**Review by caass — APPROVED** [review]
> Find a way to get rid of the Error: text at the end.
Can you tell where this is coming from?
> Should I add support for auto-editing wrangler.toml? (I would slightly prefer to have it go in a follow-up, this PR is already a little big.)
I agree it should be a follow-up -- a good idea, though.
> Can the serde impls be simplified? (see below)
They look good to me
> Should the error be different when environment files are loaded? (see below)
I'm not sure what you mean by this :/
> Another alternative is to call `process::exit`, but that feels like a heavy hammer.
I kind of agree, but I also think it's not *that* bad if there's a comment above explaining why `process::exit`. If we run into problems later, it's a one-liner to revert.

I think this looks good!

**Comment by jyn514** [comment]
@caass and I discussed this on a call last night.
 - `Error:` without anything following is now `Error: field `account_id` is required`, which looks reasonably intentional. This also makes the `process::exit` thing moot.
 - @nataliescottdavidson already looked at the messages around environment files, I just forgot to remove it from the PR description.

**Comment by nilslice** [comment]
@jyn514, thanks for the work put into this. Can you check that in places where you're printing errors, that:


1. they go to stderr 
2. we exit with a non-zero exit code (so in cases like CI pipelines, the build breaks intentionally after a wrangler related issue)

**Comment by jyn514** [comment]
@nilslice when you say check, do you mean with an automated test? or just locally?

**Comment by nilslice** [comment]
@jyn514 it would be nice to have some automation here, but probably out of scope for your PR. Just a manual pass to ensure this is the case. 

I haven't reviewed any code yet, but a manual spot check would be great.

**Comment by jyn514** [comment]
Hmm, so I'm printing *errors* to stderr, but the *account info* to stdout, because that's how `wrangler whoami` behaves. Is that ok?
```
> wrangler publish 2>/dev/null
👀  You have multiple accounts.
🕵️  You can copy your account_id below
+--------------------------+----------------------------------+
| Account Name             | Account ID                       |
+--------------------------+----------------------------------+
| Cloudflare Community Jam | 6174b0f52802f24a9f52e7015282898c |
+--------------------------+----------------------------------+
| Edge Workers             | 615f1f0479e7014f0bebcd10d379f10e |
+--------------------------+----------------------------------+
> wrangler publish >/dev/null
Error: field `account_id` is required
> echo $?
1
```

PR #1968, opened 2021-06-24
Remove misleading `rustup install` and update rust-toolchain

jyn514:
Previously, CI ran `rustup install stable` just before running lints, but didn't
actually use the new toolchain. This removes the unneeded download and updates
rust-toolchain to the latest stable version.

This also fixes a bunch of clippy warnings that were introduced since 1.49.

### DIFF
--- a/src/build/mod.rs
+++ b/src/build/mod.rs
@@ -4,7 +4,7 @@ use crate::terminal::styles;
 use crate::wranglerjs;
 use crate::{commands, install};
 
-use std::path::PathBuf;
+use std::path::Path;
 use std::process::Command;
 
 use anyhow::{anyhow, Result};
@@ -68,7 +68,7 @@ pub fn build_target(target: &Target) -> Result<String> {
     }
 }
 
-pub fn command(args: &[&str], binary_path: &PathBuf) -> Command {
+pub fn command(args: &[&str], binary_path: &Path) -> Command {
     let mut c = if cfg!(target_os = "windows") {
         let mut c = Command::new("cmd");
         c.arg("/C");
--- a/src/upload/form/wasm_module.rs
+++ b/src/upload/form/wasm_module.rs
@@ -20,8 +20,8 @@ impl WasmModule {
             .ok_or_else(|| anyhow!("filename should not be empty: {}", path.display()))?;
 
         Ok(Self {
-            filename,
             path,
+            filename,
             binding,
         })
     }
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -9,13 +9,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Install Rust
-        run: |
-          rustup update stable
-          rustup default stable
-          rustup component add rustfmt
-          rustup component add clippy
-
       - name: Check Formatting
         run: cargo fmt --all -- --check
 
--- a/rust-toolchain
+++ b/rust-toolchain
@@ -1,1 +1,3 @@
-1.49.0
+[toolchain]
+channel = "1.53.0"
+components = ["rustfmt", "clippy"]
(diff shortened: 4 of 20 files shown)


jyn514:
An alternate fix for this is not to install stable at all, and use the version in rust-toolchain. But that seems a little confusing.

src/upload/form/wasm_module.rs
```diff
@@ -20,8 +20,8 @@ impl WasmModule {
             .ok_or_else(|| anyhow!("filename should not be empty: {}", path.display()))?;
 
         Ok(Self {
-            filename,
             path,
+            filename,
```
nataliescottdavidson, line 24: What's up with this change?

src/build/mod.rs
```diff
@@ -68,7 +68,7 @@ pub fn build_target(target: &Target) -> Result<String> {
     }
 }
 
-pub fn command(args: &[&str], binary_path: &PathBuf) -> Command {
+pub fn command(args: &[&str], binary_path: &Path) -> Command {
```
nataliescottdavidson, line 71: What's the purpose of changing `PathBuf` to `Path`?

ObsidianMinor:
We originally pinned to 1.49 because stable would update overnight and add new clippy warns that'd make PRs fail.

jyn514:
> We originally pinned to 1.49 because stable would update overnight and add new clippy warns that'd make PRs fail.

Got it. Does it make sense to remove the `rustup install` command then? I think that's what confused me the most, that the version installed isn't what actually runs tests.

src/build/mod.rs
jyn514: It's no less flexible, and it makes it a little simpler because you're no longer going through Deref every time.

```
warning: writing `&PathBuf` instead of `&Path` involves a new object where a slice will do
   --> src/wranglerjs/mod.rs:227:25
    |
227 | fn run_npm_install(dir: &PathBuf) -> Result<()> {
    |                         ^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg
```

src/upload/form/wasm_module.rs
jyn514: Oh, I ran `cargo clippy --fix` and I think it reordered the fields for some reason. I'll revert it.

src/build/mod.rs
nataliescottdavidson: 👍

nataliescottdavidson:
> > We originally pinned to 1.49 because stable would update overnight and add new clippy warns that'd make PRs fail.
> 
> Got it. Does it make sense to remove the `rustup install` command then? I think that's what confused me the most, that the version installed isn't what actually runs tests.

Is that the case? Which version does run the tests?

jyn514:
1.49 is the version that runs in tests, because rust-toolchain overrides the default.

nataliescottdavidson (approved):

jyn514:
> I assume 'stable' should be mostly stable, but you guys are more familiar than I am on how often the rust team releases those changes.

Hmm, so you're using "stable" in a difference sense than the release team does - they use it to mean "mostly free of bugs", while you're using it to mean "not many new lints". New lints aren't considered breaking changes, so there could possibly be a lot added in a single release. Releases happen every 6 weeks.

I'm fine with using rust-toolchain instead of stable, but if we do that I think we should try to update it regularly every six weeks or so when there's a new release.

ObsidianMinor:
I'm confused by what we're intending with this change? We originally switched to a rust-toolchain so we could control updates, since clippy warnings are hard errors in PRs a release overnight could mean everyone's PRs are broken due to no fault of their own. I'd prefer if we kept using the rust-toolchain and updated it to whatever is stable whenever possible to prevent any major breakage of all PRs.

jyn514:
Ok, I'm going to update `rust-toolchain` to the latest stable and remove the `rustup toolchain install stable` since it's misleading.

PR #1970, opened 2021-06-25
fix: skip kv pairs that have already been uploaded

nilslice:
This adds an optional param to the function which creates the kv pairs for upload. The logic was pulled from the `filter_files` function, which itself has been removed (along with its test). 

The primary motivation here is to skip any kv pair that has already been uploaded (checked by listing keys from the namespace), so we don't grow the `vec` so needlessly large.

### DIFF
--- a/src/sites/mod.rs
+++ b/src/sites/mod.rs
@@ -6,6 +6,7 @@ mod sync;
 pub use manifest::AssetManifest;
 pub use sync::sync;
 
+use std::collections::HashSet;
 use std::ffi::OsString;
 use std::fs;
 use std::hash::Hasher;
@@ -64,6 +65,7 @@ pub fn add_namespace(user: &GlobalUser, target: &mut Target, preview: bool) -> R
 pub fn directory_keys_values(
     target: &Target,
     directory: &Path,
+    exclude: Option<&HashSet<String>>,
 ) -> Result<(Vec<KeyValuePair>, AssetManifest, Vec<String>)> {
     match fs::metadata(directory) {
         Ok(ref file_type) if file_type.is_dir() => {
@@ -95,6 +97,16 @@ pub fn directory_keys_values(
 
                     validate_key_size(&key)?;
 
+                    // asset manifest should always contain all files
+                    asset_manifest.insert(url_safe_path, key.clone());
+
+                    // skip uploading existing keys, if configured to do so
+                    if let Some(remote_keys) = exclude {
+                        if remote_keys.contains(&key) {
+                            continue;
+                        }
+                    }
+
                     upload_vec.push(KeyValuePair {
                         key: key.clone(),
                         value: b64_value,
@@ -102,8 +114,6 @@ pub fn directory_keys_values(
                         expiration_ttl: None,
                         base64: Some(true),
                     });
-
-                    asset_manifest.insert(url_safe_path, key);
                 }
             }
             Ok((upload_vec, asset_manifest, file_list))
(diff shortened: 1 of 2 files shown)


nilslice:
would be great if someone could verify that the `asset_manifest` should contain the full list of file metadata even when we can skip the pre-existing kv pair.

src/sites/mod.rs
```diff
@@ -95,15 +97,23 @@ pub fn directory_keys_values(
 
                     validate_key_size(&key)?;
 
+                    // asset manifest should always contain all files
+                    asset_manifest.insert(url_safe_path, key.clone());
```
jyn514, line 101: What happens if there are duplicate keys? Will we upload both and the second gets overridden?

src/sites/mod.rs
```diff
@@ -64,6 +65,7 @@ pub fn add_namespace(user: &GlobalUser, target: &mut Target, preview: bool) -> R
 pub fn directory_keys_values(
     target: &Target,
     directory: &Path,
+    exclude: Option<&HashSet<String>>,
```
jyn514, line 68: I think it would be nice to remove the `Option` - if callers don't want to ignore any keys, they can pass an empty set.
```suggestion
    exclude: &HashSet<String>,
```

jyn514 (commented):
I'm not sure I understand why the test was removed - would it be possible to test `directory_keys_values` instead?

src/sites/sync.rs
```diff
@@ -59,82 +57,5 @@ pub fn sync(
...
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-    use crate::sites::generate_path_and_key;
-    use std::collections::HashSet;
-    use std::path::Path;
-
-    #[test]
-    fn it_can_filter_preexisting_files() {
```
aaronlisman, line 83: I guess this would be too painful to make work with `directory_keys_values` for now? Would be good to add back in later with a mock Walk or something.

src/sites/mod.rs
nilslice: create a new HashSet and pass a ref to it, instead of None?

src/sites/sync.rs
nilslice: Yea, exactly -- the old test really didn't prove much... and with the logic embedded in the other function now it's a bit hairy to test. We'd want to change the implementation to make it testable but since this is just a quick patch I think it's ok. 

cc/ @jyn514  (w/r/t your question above)

src/sites/mod.rs
jyn514: Yup, exactly: `&HashSet::new()`

src/sites/mod.rs
nilslice: can there be duplicates? these are based on files on disk no?

src/sites/mod.rs
nilslice: I don't see the benefit, can you elaborate?

jyn514: `Option<HashSet>` has two ways to represent "don't exclude any files": `Some(HashSet::new())` and `None`. `HashSet` only has one.

jyn514: This also removes the `if let Some(...)` line below, which decreases the indentation a little.

src/sites/mod.rs
jyn514: I'm not familiar with this service :sweat_smile: if there can't be duplicates it's not worth worrying about.

src/sites/mod.rs
nilslice: hmm. yea, i suppose. `None` is pretty clear from an intent standpoint. Going to leave as-is, and if its worthwhile we can revisit

src/sites/mod.rs
nilslice: i'm not that familiar with it either, but the code suggests that it's iterating over files and creating a bundle to upload. the filepaths which ultimately determine the key & value would be unique

nataliescottdavidson (approved):

Issue #1938, opened 2021-05-28
__STATIC_CONTENT_MANIFEST missing in env object when using ES modules


ttraenkler:
#### Describe the bug

When using the new ES module syntax and rollup instead of webpack as a bundler, I cannot access the workers-site binding `__STATIC_CONTENT_MANIFEST` neither through the global env variable anymore nor through the new env object.

I have tried using both wrangler 1.16.1 or 1.16.0-durable-objects-rc.0 to upload a workers-site with the same result

#### Reproduce the bug

A list of steps to reproduce the bug.

1. upload a workers-site with the latest wrangler 
2. bundle your worker with rollup using the new ESM syntax
3. try to access the `__STATIC_CONTENT_MANIFEST` as a global variable or in the `env` object -> is undefined

#### Expected behavior

```ts
env.__STATIC_CONTENT_MANIFEST
``` 
should contain the manifest string

#### Environment and versions

Fill out the following information about your environment.

* operating system: MacOS latest
* output of `wrangler -V`: wrangler 1.16.0-durable-objects-rc.0
* output of `node -v`: v14.15.4
* content of `wrangler.toml`:

```toml
account_id = "..."
kv_namespaces = [...]
name = "..."
route = ""
type = "javascript"
vars = {ENVIRONMENT = "production", DEBUG_MODE = false}
workers_dev = true
zone_id = ""

[site]
bucket = "../build"
entry-point = "workers-site"

[durable_objects]
bindings = [{name = "COUNTER", class_name = "Counter"}]

[build]
command = "yarn && yarn build"

[build.upload]
# The "modules" upload format is required for all projects that export a Durable Objects class
dir = "dist"
format = "modules"
main = "./index.mjs"

```

I think the relevant file should be `src/upload/form/mod.rs` in the wrangler repo.
Let me know if you need more information or cannot reproduce the bug without further details. Also please point me to a branch if there is already ongoing work so I could potentially contribute a PR.

irvinebroque:
Sample project that reproduces this, I ran into it as well and wanted to confirm it wasn't caused by other configuration in another project:

https://github.com/irvinebroque/repro-kv-asset-handler-modules-issue

DrewRidley:
This issue also directly affects the KV handler because the KV handler relies on this environment variable to serve requests. Without wrangler being corrected, my PR submitted over there will not function correctly.

xortive:
[...]

PR #1973, opened 2021-06-25
Added static content manifest to module workers.

DrewRidley:
The title says it all, but basically this PR reintroduces the static content manifest for workers deployed in the module format. This is important for a few reasons, but namely, its a prerequisite to my ongoing pull request to ``kv-asset-handler``. 
Related is issue #1938 . 
 
My pull request to ``kv-asset-handler`` can be found here:
https://github.com/cloudflare/kv-asset-handler/pull/200.

### DIFF
--- a/src/upload/form/mod.rs
+++ b/src/upload/form/mod.rs
@@ -128,17 +128,7 @@ pub fn build(
                         plain_texts,
                     )?;
 
-                    if let Some(asset_manifest) = &asset_manifest {
-                        log::info!("adding STATIC_CONTENT_MANIFEST to the module.");
-                        let manifest_blob = get_asset_manifest_blob(&asset_manifest)?;
-                        return modules_worker::build_form_with_manifest(
-                            &assets,
-                            session_config,
-                            manifest_blob,
-                        );
-                    }
-
-                    return modules_worker::build_form(&assets, session_config);
+                    modules_worker::build_form(&assets, session_config, text_blobs)
                 }
             },
             None => {
--- a/src/upload/form/mod.rs
+++ b/src/upload/form/mod.rs
@@ -23,12 +23,11 @@ use text_blob::TextBlob;
 use wasm_module::WasmModule;
 
 // TODO: https://github.com/cloudflare/wrangler/issues/1083
-use super::{krate, Package};
-
 struct Module {
     pub path: PathBuf,
     pub module_type: ModuleType,
 }
+
 
 pub fn build(
     target: &Target,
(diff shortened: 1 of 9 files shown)


DrewRidley:
To proceed, all I need is someone with more knowledge than me to explain how the modules actual content is exposed. I added the module to the list of modules in the form uploaded, but I have not found a way to set the modules contents.

DrewRidley:
I tested the PR and it works on my test site that was deployed. 
```
import TEXT from 'STATIC_CONTENT_MANIFEST.txt';
```
This snippet works as intended when deployed to a worker and the static manifest is there in its entirety. I have tested it extensively and it works as intended, and should not affect non-modular deployments in any way. In the event wrangler is unable to extract the manifest, it simply uploads the form, omitting the file.

src/upload/form/mod.rs
```diff
@@ -127,7 +128,17 @@ pub fn build(
                         plain_texts,
                     )?;
 
-                    modules_worker::build_form(&assets, session_config)
+                    if let Some(asset_manifest) = &asset_manifest {
```
xortive, line 131: I'd do `let manifest_blob = asset_manifest.map(|m| get_asset_manifest_blob(&m))?` to get an `Option<String>` that you could pass into `build_form`

Might need to rearrange things to handle the `Result<String>` from `get_asset_manifest_blob`, or just use `.expect()` inside of  `get_asset_manifest_blob` so it can return a `String` (@nilslice, do y'all have a different convention here nowadays?) as there's no reason serializing the asset manifest into JSON should fail due to a user caused action.

src/upload/form/mod.rs
```diff
@@ -127,7 +128,17 @@ pub fn build(
...
+                        log::info!("adding STATIC_CONTENT_MANIFEST to the module.");
```
xortive, line 132:
```suggestion
                        log::info!("adding STATIC_CONTENT_MANIFEST to the upload form.");
```

xortive (changes requested):
Thank you for working on this :smile: I think adding directly to the form is OK, just had some feedback to make this a bit more idiomatic rust but otherwise a :+1: from me.

mikecann:
Keen to see this go in so we can build DurableObjects projects and Sites making this possible: https://discord.com/channels/595317990191398933/779390076219686943/869564247825350687

gzuidhof:
I'm not sure how helpful it is - not having this merged is also making one of my projects extra difficult.

threepointone:
this was fixed in https://github.com/cloudflare/wrangler/pull/2126, closing this one

PR #2000, opened 2021-07-15
Cleanup code in socket.rs

jyn514:
This can just use `?` instead of an explicit match.

### DIFF
--- a/src/commands/dev/socket.rs
+++ b/src/commands/dev/socket.rs
@@ -148,7 +148,8 @@ async fn print_ws_messages(
                     serde_json::from_str(&*event.to_string());
                 print_json(json_parse, event.to_string());
             }
-            _ => {}
+            Ok(protocol::Runtime::Method(_)) => {}
+            Err(err) => log::debug!("{}", err),
         }
     }
     Ok(())
--- a/src/commands/dev/socket.rs
+++ b/src/commands/dev/socket.rs
@@ -115,44 +115,40 @@ async fn print_ws_messages(
     mut read: SplitStream<WebSocketStream<MaybeTlsStream<TcpStream>>>,
 ) -> Result<()> {
     while let Some(message) = read.next().await {
-        match message {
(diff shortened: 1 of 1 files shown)


src/commands/dev/socket.rs
```diff
@@ -115,44 +115,40 @@ async fn print_ws_messages(
...
-                        print_json(json_parse, event.to_string());
-                    }
-                    _ => {}
-                }
+        let message = message?;
+        let message_text = message.into_text().unwrap();
+        log::info!("{}", &message_text);
+
+        let parsed_message: Result<protocol::Runtime> = serde_json::from_str(&message_text)
+            .map_err(|e| anyhow!("Failed to parse event:\n{}", e));
```
jyn514, line 123: I just noticed we never actually handle this error - it's silently discarded in the `match`. Should it use `?` instead?
```suggestion
        let parsed_message: Result<protocol::Runtime> = serde_json::from_str(&message_text)
            .map_err(|e| anyhow!("Failed to parse event:\n{}", e))?;
```

jspspike: I'm not sure if that would work, if the message failed to parse than it's likely a heartbeat message and therefore we just want to noop and move on. If we return the Err out of the function the `try_join!` on line 61 might kill everything but if you try and it and it doesn't stop dev after receiving heartbeat messages then I think it's fine.

jyn514: Ok, we should probably at least log it then. Right now the `anyhow!` doesn't do anything, it just allocates and then immediately deallocates.

jspspike: Yeah that might be a good idea, we have that `log::info` on line 121 but maybe a debug for the result if it failed to parse

jyn514: Great, added that :)

jspspike (approved):
Looks good to me, tested and ran into no issues

PR #2004, opened 2021-07-20
Fix the link to kv documentation in README.md

threepointone:
^ Like it says, this just fixes the link for kv docs.


xtuc (approved):
Good cath, thanks

PR #2005, opened 2021-07-20
wrangler tail using WebSockets

nataliescottdavidson:


src/commands/tail.rs
```diff
@@ -1,132 +1,80 @@
...
+            let mut headers = reqwest::header::HeaderMap::new();
+            let email_hdr = reqwest::header::HeaderName::from_static("x-auth-email");
+            let api_key_hdr = reqwest::header::HeaderName::from_static("x-auth-key");
+            headers.insert(email_hdr, email.parse().unwrap());
+            headers.insert(api_key_hdr, api_key.parse().unwrap());
+            client.headers(headers)
+        }
+        GlobalUser::TokenAuth { api_token } => client.bearer_auth(api_token),
+    };
+    let res = req_builder.send().await;
```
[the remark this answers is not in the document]
nataliescottdavidson: Yep, fixed to use the cloudflare-rs client

src/cli/mod.rs
```diff
@@ -216,20 +216,7 @@ pub enum Command {
     Whoami,
 
     /// Aggregate logs from production worker
-    #[structopt(name = "tail")]
-    Tail {
-        /// Specify an output format
-        #[structopt(long, short = "f", default_value = "json", possible_values = &["json", "pretty"])]
```
Electroid, line 222: Keep pretty?

src/commands/tail.rs
```diff
@@ -1,132 +1,76 @@
...
-            Ok(socket) => Ok(socket.local_addr()?.port()),
-            Err(_) => anyhow::bail!("the specified port {} is unavailable", port),
+// Main loop, listen for websocket messages or interrupt
+async fn listen_tail(tail_tag: String) -> Result<(), ()> {
+    // ws listener setup
+    let listen_tail_endpoint = format!("wss://tail.developers.workers.dev/{}/ws", tail_tag);
+    let (mut socket, _) =
+        tokio_tungstenite::connect_async(url::Url::parse(&listen_tail_endpoint).unwrap())
+            .await
+            .expect("Can't connect");
```
Electroid, line 29: I would make more specific

src/commands/tail.rs
```diff
@@ -1,132 +1,76 @@
...
-fn find_open_port(requested: Option<u16>, default: u16) -> Result<u16> {
-    if let Some(port) = requested {
-        let addr = format!("127.0.0.1:{}", port);
-        match TcpListener::bind(addr) {
-            Ok(socket) => Ok(socket.local_addr()?.port()),
-            Err(_) => anyhow::bail!("the specified port {} is unavailable", port),
+// Main loop, listen for websocket messages or interrupt
+async fn listen_tail(tail_tag: String) -> Result<(), ()> {
+    // ws listener setup
+    let listen_tail_endpoint = format!("wss://tail.developers.workers.dev/{}/ws", tail_tag);
```
aaronlisman, line 25: Just occurred to me that we may want to return the full URL in the API response. This lets us change it more easily in the future, though I can't think of a reason we would actually need to. Could aid with testing at least.

aaronlisman (commented):
Oops I didn't mean to start a review.

src/cli/mod.rs
nataliescottdavidson: image

Cargo.toml
```diff
@@ -29,6 +29,7 @@ env_logger = "0.8.4"
 eventual = "0.1.7"
 flate2 = "1.0.18"
 fs2 = "0.4.3"
+futures = ""
```
ObsidianMinor, line 32: This is valid? We're going to have to pin it at a specific version anyway for crates.io publishing...

nataliescottdavidson:
@Electroid filters look good. 
image

Electroid (approved):

PR #2015, opened 2021-07-27
remove unnecessary cwd and fix toml defaults

xtuc:


src/settings/toml/manifest.rs
```diff
@@ -356,7 +354,7 @@ impl Manifest {
             usage_model: self.usage_model, // Top level
             wasm_modules: self.wasm_modules.clone(),
             compatibility_date: self.compatibility_date.clone(),
-            compatibility_flags: self.compatibility_flags.clone(),
+            compatibility_flags: self.compatibility_flags.clone().unwrap_or_else(|| vec![]),
```
jyn514, line 357:
```suggestion
            compatibility_flags: self.compatibility_flags.clone().unwrap_or_default()
```

src/settings/toml/manifest.rs
```diff
@@ -61,11 +60,10 @@ pub struct Manifest {
     pub wasm_modules: Option<HashMap<String, PathBuf>>,
     pub triggers: Option<Triggers>,
     pub durable_objects: Option<DurableObjects>,
-    #[serde(default, with = "string_empty_as_none")]
     pub usage_model: Option<UsageModel>,
     pub compatibility_date: Option<String>,
     #[serde(default)]
-    pub compatibility_flags: Vec<String>,
+    pub compatibility_flags: Option<Vec<String>>,
```
jyn514, line 66: Why did this change? Was there some bug with it before?

xtuc: if you read and write the configuration again, it would add `compatibility_flags = []`, this change makes the round trip identical

jyn514: > if you read and write the configuration again, it would add compatibility_flags = [], this change makes the round trip identical

When would that happen, though? Are you using wrangler as a library or something?

xtuc: Yes we use wrangler as a lib to generate a project and edit the wrangler.toml

geelen:
I wasn't aware wrangler could update the TOML at all? From my discussion about migrating the `dir` property: https://github.com/cloudflare/wrangler/issues/1951#issuecomment-862452882 I

> Note modifying wrangler.toml in particular is tricky without messing with formatting -- we have toml-edit in wrangler now, but it only supports toml 0.4 syntax while the main parser toml-rs supports the full toml 1.0 syntax (namely, dotted keys). There's a good chance some folks are using toml 1.0 syntax, and we'd fail to edit their config as a result.

Does that mean we're already effectively only assuming support for Toml 0.4? And we could reopen that discussion about auto-updating config?

src/settings/toml/builder.rs
```diff
@@ -14,8 +15,6 @@ const UPLOAD_DIR: &str = "dist";
 #[serde(deny_unknown_fields)]
 pub struct Builder {
     pub command: Option<String>,
-    #[serde(default = "project_root")]
-    pub cwd: PathBuf,
```
xortive, line 18: Why do we need to remove `cwd`? Even if not super useful, this config option made it into a stable release and we'll break folks who started using it if we remove it.

xortive:
> I wasn't aware wrangler could update the TOML at all? From my discussion about migrating the `dir` property: [#1951 (comment)](https://github.com/cloudflare/wrangler/issues/1951#issuecomment-862452882) I
> 
> > Note modifying wrangler.toml in particular is tricky without messing with formatting -- we have toml-edit in wrangler now, but it only supports toml 0.4 syntax while the main parser toml-rs supports the full toml 1.0 syntax (namely, dotted keys). There's a good chance some folks are using toml 1.0 syntax, and we'd fail to edit their config as a result.
> 
> Does that mean we're already effectively only assuming support for Toml 0.4? And we could reopen that discussion about auto-updating config?

Wrangler only edits tomls provided in templates, which at least for CF templates have always been toml 0.4.

It'd be really nice to auto-update user configs, and I think in practice folks using toml 0.5/1.0 syntax will be rare, we don't use it in our examples.

src/settings/toml/manifest.rs
ObsidianMinor: Why not add a `skip_serializing_if = "Vec::is_empty()"` annotation instead?

JacobMGEvans:
Thanks for creating this PR! Wrangler v1 is now deprecated and support is only being provided for critical updates or security concerns. As such, we are closing this PR.

New versions of Wrangler are maintained in the [workers-sdk repo](https://github.com/cloudflare/workers-sdk). If you are running into a similar issue with wrangler v2, please report it or create a new PR in the [workers-sdk repo](https://github.com/cloudflare/workers-sdk). For more info about wrangler v1 deprecation, please check out our [blog post](https://blog.cloudflare.com/deprecating-wrangler-v1).

Issue #946, opened 2019-12-11
[dev] pass --inspect flag to get websocket endpoint for chrome://inspect


EverlastingBugstopper:
node provides an interesting way to attach a debugger to its REPL

```
$ node --inspect
Debugger listening on ws://127.0.0.1:9229/long-hash-hash-hash-hash
For help, see: https://nodejs.org/en/docs/inspector
```

You can then navigate to chrome://inspect and it will show a node instance. If you run console.log in the node process, it will show up in the console on the browser. 

I think it would be a good idea to add `--inspect` to `wrangler dev`. This would mean that when your dev server starts up, it acts the same way in the terminal, except it also takes all websocket messages it receives and passes it to a local websocket listening on your machine. This would work the same way that the node inspector does, and you'd be able to navigate to `chrome://inspect` and see wrangler dev there instead of node.

A pitfall with this is that the runtime's implementation of the devtools protocol is incomplete (you can't set breakpoints). But I still think it's worth having - especially for interacting with large nested objects and viewing network requests.

PierBover:
@EverlastingBugstopper when `dev` crashes we'd need to manually reopen `chrome://inspect` right?

Since I imagine the node process would be killed.

EverlastingBugstopper:
I'm not sure actually - would need to see how it all works. My guess is yes - but we'd attempt to achieve parity with whatever `node` does.

Issue #1965, opened 2021-06-23
`wrangler tail` failing silently — consider adding `-p 0` as default?


geelen:
Had a bug last night where starting and stopping `wrangler tail` a few times had resulted in the default port, `8080` getting blocked. I believe it's a `cloudflared` problem behind the scenes, though I don't know what would be involved in having `wrangler` detect the error state and report it correctly to the user.

The workaround, at least on my machine, is to pass `-p 0` (thanks to [this stackoverflow answer](https://unix.stackexchange.com/a/55918)). This chooses a random, unused port, and in my testing `wrangler tail -p 0` works 100% of the time, every time.

This could be the default argument, unless it's a platform-specific thing?

PR #2016, opened 2021-07-28
Implement `wrangler dev --inspect`

jyn514:
Helps with #946.

## How do I use this?

1. `wrangler dev --inspect`, which should have output similar to this:
```
💁  watching "./"
[2021-07-28 13:38:20] GET worker.jnelson.workers.dev/ HTTP/1.1 200 OK
🕵️  Open chrome://inspect, click 'Configure', and add localhost:9230
```
2. Follow the instructions about chrome://inspect. You should see a screen like this: image
3. Click 'inspect', which should bring up a devtools popup:
image

## What features are implemented?

- Sources tab
- Profiler tab

## What features are not implemented?

- Memory tab. This loads fine, but doesn't report accurate information. This is likely an issue that needs to be fixed upstream in the remote Worker.
- Breakpoints. These need to be implemented upstream in the remote Worker.
- Console tab. I'm not sure why this isn't working.
- 'Trace' button on chrome://inspect. Hitting this will cause Chrome to send a GET request without a path, to which wrangler returns a 400 Bad Request. I suspect there's something wrong with the JSON endpoint that wouldn't be too hard to fix.
- Closing and re-opening the devtools window only sporadically works; you sometimes need to restart wrangler to fix it. This happens because wrangler currently uses a single persistent websocket connection to the remote. I tried changing it to open a new websocket each time Chrome connects, but that fails with 502 errors on the later connections, and this at least sometimes works.
- Making changes to your code will live-update the preview, but will not live-update the websocket. This can be confusing, since the code you're debugging is no longer running on cloudflareworkers.com. I have no idea how to make this work.

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -177,6 +177,10 @@ pub enum Command {
         /// but can be set to http
         #[structopt(name = "upstream-protocol")]
         upstream_protocol: Option<Protocol>,
+
+        /// Inspect the worker using Chrome DevTools
+        #[structopt(long)]
+        inspect: bool,
     },
 
     /// Publish your worker to the orange cloud
--- a/src/commands/dev/socket.rs
+++ b/src/commands/dev/socket.rs
@@ -49,21 +49,8 @@ pub async fn listen(
     // we loop here so we can issue a reconnect when something
     // goes wrong with the websocket connection
     loop {
-        // TODO: replace this with a prewarm call
-        if inspect.is_some() {
-            // Startup the edge isolate.
-            // TODO: does this need to be https in some cases?
-            // let scheme = if server_config.host.is_https() { "https" } else { "http" };
-
-            // Sleep for a bit first to give the webserver time to start up.
-            tokio::time::sleep(Duration::from_millis(200)).await;
-            // TODO: should this be retried? It definitely doesn't need special logic to refresh the token,
-            // since anyone can access the worker's HTTP server.
-            reqwest::get(format!("http://{}", server_config.listening_address)).await?;
-        }
         let sender = refresh_session_sender.clone();
         let mut ws_stream = connect_retry(&socket_url, sender).await;
-        // let mut ws_stream = connect_retry(|| connect_async(&socket_url), sender).await.0;
 
         // console.log messages are in the Runtime domain
         // we must signal that we want to receive messages from the Runtime domain
--- a/src/commands/dev/socket.rs
+++ b/src/commands/dev/socket.rs
@@ -1,4 +1,8 @@
+use std::convert::Infallible;
+use std::fmt::Display;
+use std::net::SocketAddr;
 use std::sync::mpsc::Sender;
+use std::sync::Arc;
 use std::time::Duration;
 
 use chrome_devtools as protocol;
@@ -6,7 +10,14 @@ use chrome_devtools as protocol;
 use futures_util::future::TryFutureExt;
 use futures_util::sink::SinkExt;
 use futures_util::stream::{SplitStream, StreamExt};
+use http::header::SEC_WEBSOCKET_KEY;
+use http::{HeaderValue, Request, Response, StatusCode};
+use hyper::upgrade::Upgraded;
+use hyper::{Body, Server};
+use log::{debug, error, info, trace, warn};
+use tokio::try_join;
 use tokio_stream::wrappers::UnboundedReceiverStream;
+use tokio_tungstenite::tungstenite::error::ProtocolError;
 
 use crate::terminal::colored_json_string;
 use crate::terminal::message::{Message, StdErr, StdOut};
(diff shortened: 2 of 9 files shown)


jyn514:
> Console tab. I'm not sure why this isn't working.

I take it back - this works if you hit 'Open dedicated DevTools for Node', but not if you hit 'inspect'.

jyn514:
Hmm - so about the 'Console' thing - when trying this with Deno, which has a similar feature, I'm seeing it send back
```
{"method":"Debugger.paused","params":{"callFrames":[{"callFrameId":"-4899094861937594869.1.0","functionName":"","functionLocation":{"scriptId":"79","lineNumber":0,"columnNumber":0},"location":{"scriptId":"79","lineNumber":0,"columnNumber":0},"url":"file:///home/jnelson/src/test-wrangler/index.js","scopeChain":[{"type":"module","object":{"type":"object","className":"Object","description":"Object","objectId":"-4899094861937594869.1.1"},"startLocation":{"scriptId":"79","lineNumber":0,"columnNumber":0},"endLocation":{"scriptId":"79","lineNumber":12,"columnNumber":0}},{"type":"global","object":{"type":"object","className":"Window","description":"Window","objectId":"-4899094861937594869.1.2"}}],"this":{"type":"undefined"}}],"reason":"debugCommand","hitBreakpoints":[]}}
```
I wonder if Chrome is waiting to see that before sending `Runtime.evaluate`? If so, that gives me hope it would be fixed automatically when I implement breakpoints.

jyn514:
> 'Trace' button on chrome://inspect. Hitting this will cause Chrome to send a GET request without a path, to which wrangler returns a 400 Bad Request. I suspect there's something wrong with the JSON endpoint that wouldn't be too hard to fix.

This is also broken for node and deno, so probably not a bug on our end.

src/cli/mod.rs
```diff
@@ -175,6 +175,10 @@ pub enum Command {
         /// but can be set to http
         #[structopt(name = "upstream-protocol")]
         upstream_protocol: Option<Protocol>,
+
+        /// Inspect the worker in Chrome devtools
```
Electroid, line 179:
```suggestion
        /// Inspect the worker using Chrome DevTools
```

src/commands/dev/socket.rs
```diff
@@ -19,66 +30,309 @@ use tokio_tungstenite::{connect_async, tungstenite, MaybeTlsStream, WebSocketStr
...
-        let (keep_alive_tx, keep_alive_rx) = mpsc::unbounded_channel();
+        // parse all incoming messages and print them to stdout
+        if let Some(worker_name) = &inspect {
+            StdErr::help("Open chrome://inspect, click 'Configure', and add localhost:9230");
+            // StdErr::help(&format!("Open devtools://devtools/bundled/inspector.html?experiments=true&v8only=true&ws=localhost:{port}{path} in Chrome", port=DEVTOOLS_PORT, path=socket_url.path()));
 
-        // every 10 seconds, send a keep alive message on the channel
-        let heartbeat = keep_alive(keep_alive_tx);
+            // Construct our SocketAddr to listen on...
+            let addr = SocketAddr::from(([127, 0, 0, 1], DEVTOOLS_PORT));
```
Electroid, line 77: Instead of hard-coding a port, can we just pass port `0` and allow the OS to allocate one for us? `wrangler tail` has had issues with that.

src/commands/dev/socket.rs
```diff
@@ -19,66 +30,309 @@ use tokio_tungstenite::{connect_async, tungstenite, MaybeTlsStream, WebSocketStr
...
+        .forward(remote_write)
+        .map_err(|e| WebsocketError {
+            inner: e,
+            from_chrome: true,
+        });
+    try_join!(edgeworker_to_chrome, chrome_to_edgeworker)?;
+    Ok(())
+}
+
+async fn start_websocket(
```
Electroid, line 207: Do we do anything special during the WebSocket handshake that requires us to write out the handshake, mostly manually? You can checkout how it's done in the `tails-v2` branch, but there is a higher-level `tokio_tungstenite::connect_async`

Electroid (commented):
This is super cool! Left some comments, but from a product-perspective is awesome :)

src/commands/dev/socket.rs
jyn514: Hmm, that would be a little more work for the end user, because they'd have to change the port Chrome looks at each time they restart Wrangler.

What trouble has `tail` had?

src/commands/dev/socket.rs
jyn514: I didn't know that existed! I'll take a look, thanks :)

(This would be accept_async, but your point still stands.)

src/commands/dev/socket.rs
Electroid: Ah okay, if Chrome uses a stable port this is fine. There was an issue with port conflict with the old `tail` that uses `cloudflared`. Issue was here: https://github.com/cloudflare/wrangler/issues/1965

jyn514: Yup, node and deno both use 9229, so this uses 9230 to avoid conflicting. I think Chrome already has 9222 enabled by default, so if we used that it would be one less step, but I'm not sure what already uses that port.

src/commands/dev/socket.rs
jyn514: Hmm, I'm not sure how this would work: `accept_async` requires that I consume the stream, and doesn't let me return an HTTP response - it wants access to the raw TCP socket, which I don't have at this point because this is in a `hyper` request handler. I couldn't get access to the raw socket without avoiding `hyper` altogether, which I'd rather not do since I still have to serve other HTTP requests in the meantime.

jyn514:
Prewarming for authenticated previews isn't implemented upstream yet, but once it is this diff should make it work for `--inspect`: https://gist.github.com/jyn514/1120f4fe889f50ff241e6b1ae8335b12

jkup:
This is awesome!!!

I had two questions:

1. The breakpoints are behaving strangely for me. I know you said in a comment they are not working but for me they do sort of work they just don't pause. For example if add some code to a Worker like
```js
  for (let i =0; i < 100; i++) {
    console.log(i);
  }
  return new Response(`Hello worker!`, {
    headers: { 'content-type': 'text/plain' },
  })
```

and put a breakpoint on the for loop and inside it.. it just flickers really fast 100 times. Is there a way to just not let people add breakpoints or is that more trouble than it's worth?

2. I see the Memory tab loading great but I'm seeing strange results when I try to analyze an app. Do you have a memory consumption example working I could check out? On my end it seems like the Heap allocation readings I'm getting don't change no matter how much memory I consume.

Thanks!

jyn514:
> Is there a way to just not let people add breakpoints or is that more trouble than it's worth?

Hmm, I think it might be possible by overriding https://chromedevtools.github.io/devtools-protocol/1-3/Debugger/#method-getPossibleBreakpoints to return an empty list. But I'm not sure whether that would be easier or harder than just getting breakpoints to work; I would rather spend time on that.

Note that any sort of change to the info being reported has to be made upstream in the Workers runtime; if we did it on Wrangler's end by overriding the response it means more delay between when the feature gets implemented and when people can use it (since they'd have to update their version of wrangler).

> I see the Memory tab loading great but I'm seeing strange results when I try to analyze an app. Do you have a memory consumption example working I could check out?

No, I don't have a working example of this. I'll remove it from the "Features implemented" section of the PR description.

caass (approved):
LGTM 👍🏻

Issue #2021, opened 2021-08-08
Account ID is not loaded when setting up cron triggers


lujjjh:
#### Describe the bug

After I configured [`triggers.crons`](https://developers.cloudflare.com/workers/cli-wrangler/configuration#triggers) in the wrangler.toml, `wrangler publish` failed with the messages below:

```
✨  Build completed successfully!
Error: Something went wrong with the request to Cloudflare...
Could not route to /accounts/workers/scripts/***/schedules, perhaps your object identifier is invalid? [API code: 7003]
No route for that URI [API code: 7000]
```

Note that `:account_id` is missing in the API URL.

I went through the code and discovered that `account_id` is not loaded when constructing the `DeployTarget::Schedule`.

https://github.com/cloudflare/wrangler/blob/d2f49e2d800c72160963952314f35ac0e72644f3/src/settings/toml/manifest.rs#L277

#### Reproduce the bug

A list of steps to reproduce the bug.

1. In any wrangler project, add `triggers.crons = ["*/5 * * * *"]` to the wrangler.toml.
2. Run `wrangler publish` without any `CF_*` environment variables.

#### Expected behavior

Cron triggers are configured without errors.

#### Environment and versions

Fill out the following information about your environment.

* operating system: Ubuntu 20.04
* output of `wrangler -V`: wrangler 1.19.0
* output of `node -v`: v16.6.1
* content of `wrangler.toml`:

```toml
name = "***"
type = "javascript"
workers_dev = true

[triggers]
crons = ["*/5 * * * *"]
```

PR #2023, opened 2021-08-09
Load account ID before uploading cron triggers

jyn514:
Fixes https://github.com/cloudflare/wrangler/issues/2021.

### DIFF
--- a/src/deploy/schedule.rs
+++ b/src/deploy/schedule.rs
@@ -11,20 +11,6 @@ pub struct ScheduleTarget {
 }
 
 impl ScheduleTarget {
-    pub fn build(
-        account_id: Option<String>,
-        script_name: String,
-        crons: Vec<String>,
-    ) -> Result<Self> {
-        // TODO: add validation for expressions before pushing them to the API
-        // we can do this once the cron parser is open sourced
-        Ok(Self {
-            account_id: account_id.unwrap_or_default(),
-            script_name,
-            crons,
-        })
-    }
-
     pub fn deploy(&self, user: &GlobalUser) -> Result<Vec<String>> {
         log::info!("publishing schedules");
         let schedule_worker_addr = format!(
--- a/src/settings/toml/manifest.rs
+++ b/src/settings/toml/manifest.rs
@@ -278,8 +278,11 @@ impl Manifest {
         };
 
         if let Some((crons, account)) = crons {
-            let scheduled =
-                deploy::ScheduleTarget::build(account.cloned(), script.clone(), crons.to_vec())?;
+            let scheduled = deploy::ScheduleTarget {
+                account_id: account.cloned().unwrap_or_default(),
+                script_name: script.clone(),
+                crons: crons.to_vec(),
+            };
             deployments.push(DeployTarget::Schedule(scheduled));
         }
 
--- a/src/settings/toml/manifest.rs
+++ b/src/settings/toml/manifest.rs
@@ -262,24 +262,24 @@ impl Manifest {
 
         let crons = match env {
             Some(e) => {
-                let account_id = e
-                    .account_id
-                    .as_ref()
-                    .or_else(|| self.account_id.if_present());
+                let account_id = match e.account_id.as_ref() {
+                    Some(id) => id,
+                    None => self.account_id.load()?,
+                };
                 e.triggers
                     .as_ref()
                     .or_else(|| self.triggers.as_ref())
                     .map(|t| (t.crons.as_slice(), account_id))
             }
-            None => self
-                .triggers
-                .as_ref()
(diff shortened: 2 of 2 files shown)


nataliescottdavidson (approved):

jyn514:
The error comes because wrangler is loading the global account ID when it isn't set in config.toml, and the test expects it to error instead:
```
---- settings::toml::tests::deployments::it_errors_on_get_deployments_missing_account_id stdout ----
thread 'settings::toml::tests::deployments::it_errors_on_get_deployments_missing_account_id' panicked at 'assertion failed: manifest.get_deployments(environment).is_err()', src/settings/toml/tests/deployments.rs:99:5
```
I'm not sure how to fix this. I can't just change the test to expect this to pass, or it will fail for people running `cargo test` without logging in first. I also can't just let it be since that fails it if you _have_ run `wrangler login`. My only idea is to sandbox this somehow, so it uses a different config file when running tests than when running normally? This is in an integration test so we'd have to set it externally somehow, either through an API meant only for tests or through an environment variable.
```

**Comment by jyn514** [comment]
Something is wrong here. This test was passing before: https://github.com/cloudflare/wrangler/blob/99473b72b72f6251155f089f32fa8eeccbd1875f/src/settings/toml/tests/deployments.rs#L1072
But that test doesn't set the account_id in the toml file anywhere. So it must have been possible for this to work without an account ID before.

**Comment by jyn514** [comment]
... but it doesn't :/

```
$ wrangler --version
wrangler 1.17.1-rc.1
$ wrangler publish
🕵️  You can copy your account_id below
+----------------------------+----------------------------------+
| Account Name               | Account ID                       |
+----------------------------+----------------------------------+
| Jyn514@gmail.com's Account | e1706d218241c1230155b4f91b3218af |
+----------------------------+----------------------------------+
Error: field `account_id` is required to deploy to workers.dev
```

jspspike:
I think I know the issue you're having. These tests don't run the entire publish step, they just test [`get_deployments`](https://github.com/cloudflare/wrangler/blob/99473b72b72f6251155f089f32fa8eeccbd1875f/src/settings/toml/manifest.rs#L207) which just reads the `wrangler.toml` and spits out things that need to deployed. This doesn't actually deploy them so normally wouldn't need to make an api call. Now for the cron step it will try to load the `account_id` and since it's not present in the `wrangler.toml` with your new thing it'll try to make an api call and then fail since it has no credentials

jyn514:
@jspspike thanks, that was helpful! Here's what I've tried after that:
1. Never load the api_token in tests: https://github.com/cloudflare/wrangler/commit/30281d44874264bd4c7fea8289798828c4499850. That doesn't work because `LazyAccountId::load` will still try to fetch the account_id, which gives an auth error from Cloudflare's API.
2. Never load the account_id in tests: https://github.com/cloudflare/wrangler/commit/c3ad80efc0f938b40a221dbd93fe448c5cbaf454. That doesn't work because we test that `get_deployments` fails if there's no account ID :/ I don't know how to keep that test and still have the correct behavior.

jyn514:
In particular there's a lot of global state around `GlobalUser` I don't know how to manage. Really I want the account_id to be configurable per-test, but that seems like a really big refactor I'm not sure how to do.

@nilslice @nataliescottdavidson @jspspike do one of you have suggestions?

jyn514:
A really awful hack I thought of is to add a new `with_default_account_id(val: Option<String>, f: impl FnOnce)` function that uses a Mutex or thread-local to decide what the default should be. But that seems really ugly.

jyn514:
> A really awful hack I thought of is to add a new #[cfg(test)] fn with_default_account_id(val: Option<String>, f: impl FnOnce) function that uses a Mutex or thread-local to decide what the default should be. But that seems really ugly.

This turned out to be both broken _and_ racy :cry:

jspspike:
>Never load the account_id in tests: c3ad80e. That doesn't work because we test that get_deployments fails if there's no account ID :/ I don't know how to keep that test and still have the correct behavior:

If this is the only test that is failing, then this change should make it so that you don't need an `account_id` in you `wrangler.toml` to get deployments. So it might be safe to remove the test? Ideally replace the test with something that mocks out the api call and still works but we don't have that kind of mocking for tests at the moment. I think someone else on @cloudflare/wrangler should probably decide if that's okay to do

jyn514:
Ok, I got this working without too many hacks. The way it works is by forcing tests to explicitly write out the `account_id` in wrangler.toml when they want it to be present; otherwise, `get_deployments()` will fall back to loading it, which will always error out when `cfg(test)` is set. That makes sure that:
- The test is reproducible between different users
- The test doesn't depend on what the default value of the account_id is (it can only depend on there being an error if it isn't present)

PR #2024, opened 2021-08-10
Allow worker with no routes to be published

### LINKED ISSUE #1700: wrangler push command

I developed a workers script, which is useful for handling website traffic when the server is under maintenance. Of course, I won't route this workers to any address, nor publish it to workers.dev domain. So I need a command that can push my script update to Cloudflare Workers without routing it to a any route or workers.dev domain.

#### Basic example

wrangler push - a command to push latest script changes to the Cloudflare Workers without routing it to any route.

jspspike:
Fixes #1700

### DIFF
--- a/src/settings/toml/manifest.rs
+++ b/src/settings/toml/manifest.rs
@@ -289,7 +289,7 @@ impl Manifest {
         };
 
         if durable_objects.is_none() && deployments.is_empty() {
-            anyhow::bail!("Please specify your deployment routes or `wrangler_dev = true` inside of your configuration file. For more information, see: https://developers.cloudflare.com/workers/cli-wrangler/configuration#keys")
+            StdOut::warn("No depolyment routes specified, worker will not be triggered. Please specify your deployment routes or set `wrangler_dev = true` inside of your configuration file in order to trigger your worker. For more information, see: https://developers.cloudflare.com/workers/cli-wrangler/configuration#keys");
         }
 
         Ok(deployments)


jspspike:
Confused if the behavior of `when_top_level_zoneless_env_zoned_single_route_empty` is actually intended. It essentially has `wrangler_dev = true` in the top level of the toml but then uses an environment where `wrangler_dev` is not set. Should the environment inherit the top level `wrangler_dev` and then create a zoneless (workers.dev) route or create a zoned route since the environment doesn't have `wrangler_dev` set?

nilslice:
@jspspike - looking at this now. An aside, but relevant to your PR... would you add a commit that updates this error message:
> ⚠️  No depolyment routes specified, worker will not be triggered. Please specify your deployment routes or set `wrangler_dev = true` inside of your configuration file in order to trigger your worker. For more information, see: https://developers.cloudflare.com/workers/cli-wrangler/configuration#keys

So that instead of `wrangler_dev = true` it is corrected to print `workers_dev = true`?

nilslice:
https://github.com/cloudflare/wrangler/blob/4e18717d7c632335db8198252d311c40d95d67d8/src/settings/toml/tests/deployments.rs#L808-L828

The test you mention (embedded above) generates a `wrangler.toml` manifest with the following configuration:

```toml
name = "top_level_zoneless_env_zoned_single_route_empty"
type = "webpack"
account_id = "fakeaccountid"
workers_dev = true

[env.test]
route = ""
zone_id = "samplezoneid"
```

Looking at this, I would assume the intention of these settings is to have no `Zoned` route, since there is no `route` defined despite the `zone_id` being present, but to inherit the `workers_dev` key into the environment. I'm not sure how the script naming works there (maybe append `-test` to the value in `name`?). 

I assume this should be inherited based on the way we [document the configuration](https://developers.cloudflare.com/workers/cli-wrangler/configuration#keys), seen here:
image

Maybe @Electroid could comment on this assumption - and if we're in agreement, I think the test might need to be changed to expect 1 zoneless route (the workers.dev _test_ route)? This would also mean some code is changed somewhere to inherit this top-level `workers_dev` key.

jspspike:
Going off of the docs linked I made `workers_dev` get passed down to environments. This made some tests need to be changed where both zoned and zoneless routes would be created if the environment got `workers_dev = true` from the top level and also had a zone/routes in the environment.

I went through the changed tests and tried to verify that the behavior is intended, but any changed tests should be looked at closely by the reviewer to make sure some intended behavior isn't changed since this will affect a lot of things in terms of what routes are deployed on publish

tests/fixtures/wrangler_toml.rs
```diff
@@ -124,11 +124,15 @@ impl WranglerToml {
...
     ) -> WranglerToml {
-        WranglerToml {
+        let wrangler_toml = WranglerToml {
             workers_dev: Some(workers_dev),
             account_id: Some(account_id),
             ..WranglerToml::webpack(name)
-        }
+        };
+
+        eprintln!("{:#?}", &wrangler_toml);
```
Electroid, line 133: Debug statement?

jspspike: Yes to print the toml during tests, all the other toml helper functions had them except this one for some reason

PR #2028, opened 2021-08-13
add a NotADirectoryError type for sites and add a loop for interactive confirm 

ocsfrank:
These changes were two todos in `terminal/interactive.rs` and `sites/mod.rs`


src/sites/mod.rs
```diff
@@ -61,6 +63,17 @@ pub fn add_namespace(user: &GlobalUser, target: &mut Target, preview: bool) -> R
     Ok(site_namespace)
 }
 
+#[derive(Debug, Clone)]
+struct NotADirectoryError;
```
mrbbot, line 67: Looks good! :slightly_smiling_face: One tiny thing, this should probably be `pub` since the function throwing this is public. Do you know why we wanted a separate error type here?
```suggestion
pub struct NotADirectoryError;
```

ocsfrank: I don't have the context for the reason why as it was a todo comment left by someone else. I would think that an explicit error would give users more context about what's going wrong.

mrbbot (approved):

mrbbot (approved):

mrbbot (approved):

PR #2029, opened 2021-08-13
Remove build.rs

jyn514:
While working on tests for `get_deployments()` in https://github.com/cloudflare/wrangler/pull/2023, I found that any change to
`src/settings/toml/tests/mod.rs` would cause the whole Wrangler library to be rebuilt, not just the
tests. I found it was because Cargo reran the build.rs on every change to every file, which was far slower than necessary.
I could have changed the build script to output `rerun-if-changed`, but it turns out it wasn't even
necessary in the first place; so this improves the first-time build time as well.

- Use `CARGO_MANIFEST_DIR` instead of the working directory of the build script
- Use `--cfg debug_assertions` instead of the `PROFILE` cargo sets for build scripts

  In theory these could be different, but in practice we never build release artifacts with debug
  assertions, so this shouldn't affect the behavior unless someone explicitly builds from source and
  also modifies `Cargo.toml` to enable assertions.


src/install/target.rs
```diff
@@ -9,7 +9,4 @@ pub const x86_64: bool = cfg!(target_arch = "x86_64");
 pub const aarch64: bool = cfg!(target_arch = "aarch64");
 
 // Capture if {Wrangler} is in release or debug mode
-pub const DEBUG: bool = cfg!(feature = "debug");
-
-// Capture source location, only in debug mode.
-pub const SOURCE_DIR: &str = env!("SOURCE_DIR");
+pub const DEBUG: bool = cfg!(debug_assertions);
```
jyn514, line 12: Actually, the only places this is used is in `wranglerjs::install`: https://github.com/cloudflare/wrangler/blob/19cdb329124a69cb09b7f8c4a5c50b5d85827102/src/wranglerjs/mod.rs#L292-L296
So I wonder if a better check is `current_dir() == env!("CARGO_MANIFEST_DIR")`.

jspspike (approved):
Tested locally and it looks good to me

Issue #1525, opened 2020-08-25
Regression: cannot start a local unauthenticated dev server without an account id in wrangler.toml


0dinD:
#### Environment

* operating system: Linux (Debian 10 Buster)
* output of `rustc -V`: N/A
* output of `node -v`: v14.8.0
* output of `wrangler -V`: 👷 ✨  wrangler 1.11.0
* contents of `wrangler.toml`:
```toml
name = "test"
type = "javascript"
workers_dev = true
```

#### Steps to reproduce
1. Create a valid worker
2. Run `wrangler dev`

#### What did you expect to see?
```
💁  Running preview without authentication.
💁  watching "./"
👂  Listening on http://127.0.0.1:8787
```

#### What did you see instead?
```
Error: field `account_id` is required to deploy to workers.dev
```

#### Further comments
This is a regression from wrangler 1.10, where no account id was required in wrangler.toml. I don't think it should be required if you don't want to authenticate and just want to run the worker on your local machine for quick development and testing. 

The `dev` command actually does work as expected if you supply an account id in wrangler.toml, but it doesn't even have to be valid as it's clearly not used when you don't authenticate. So the following is a workaround:
```toml
name = "test"
type = "javascript"
workers_dev = true
account_id = "not_a_valid_id"
```

nataliescottdavidson:
This is a documentation issue. `wrangler dev` runs on the edge, not your local machine. Calls to localhost:8787 are routed to a preview worker. So your prod workers will be unchanged, but it is a real worker on the edge. You need to be authenticated for that to work.

My understanding is that the old `wrangler dev` is now `wrangler preview`, you can run locally using that.

0dinD:
@nataliescottdavidson Thanks for clarifying! However, I don't agree with this statement:

> ...it is a real worker on the edge. You need to be authenticated for that to work.

Clearly, I don't need to be authenticated, because the `dev` command works just fine in wrangler `1.10`. In wrangler `1.11`, nothing has changed except for the fact that `dev` will produce an error if `account_id` is empty, which is what this issue is about. The account id is clearly not used for any sort of authentication since everything works the same in `1.11` as in `1.10` so long as you specify **anything** in account_id, such as `account_id = "not_a_valid_id"`.

[...]

PR #2030, opened 2021-08-16
Allow running `wrangler dev` without an account_id

jyn514:
Previously, running it with no account_id and no config in ~/.wrangler would give an error:

```
Error: config path does not exist /home/jnelson/.wrangler/config/default.toml. Try running `wrangler login` or `wrangler config`
```

This is not technically a regression from loading account_id lazily, because in wrangler 1.17 it would just error earlier that account_id was required:
```
🕵️  You can find your account_id in the right sidebar of your account's Workers page
Error: field `account_id` is required to deploy to workers.dev
```

However, it seems odd that you'd have to have an account to run an unauthenticated preview.
This changes wrangler to only require an account_id or config file if you're actually deploying your site.

Fixes https://github.com/cloudflare/wrangler/issues/1525

### DIFF
--- a/src/settings/toml/tests/deployments.rs
+++ b/src/settings/toml/tests/deployments.rs
@@ -44,7 +44,7 @@ fn it_can_get_multiple_deployments() {
             zone_id: ZONE_ID.to_owned(),
         }),
         DeployTarget::Zoneless(ZonelessTarget {
-            account_id: ACCOUNT_ID.to_owned(),
+            account_id: Some(ACCOUNT_ID.to_string()).into(),
             script_name: script_name.to_owned(),
         }),
     ];
@@ -66,7 +66,7 @@ fn it_can_get_a_top_level_zoneless_get_deployments() {
     let actual_deployments = manifest.get_deployments(environment).unwrap();
     let expected_deployments = vec![DeployTarget::Zoneless(ZonelessTarget {
         script_name: script_name.to_string(),
-        account_id: ACCOUNT_ID.to_string(),
+        account_id: Some(ACCOUNT_ID.to_string()).into(),
     })];
 
     assert_eq!(actual_deployments, expected_deployments);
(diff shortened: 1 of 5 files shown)


jyn514:
So, I'm hitting more test failures ... the current ones are
```
thread 'settings::toml::tests::deployments::it_gets_deployments_with_route_and_workers_dev_true' panicked at 'assertion failed: manifest.get_deployments(environment).is_err()', src/settings/toml/tests/deployments.rs:508:5

thread 'settings::toml::tests::deployments::it_gets_deployments_with_routes_and_workers_dev_true' panicked at 'assertion failed: manifest.get_deployments(environment).is_err()', src/settings/toml/tests/deployments.rs:523:5

---- settings::toml::tests::deployments::it_errors_on_get_deployments_missing_account_id stdout ----
thread 'settings::toml::tests::deployments::it_errors_on_get_deployments_missing_account_id' panicked at 'assertion failed: manifest.get_deployments(environment).is_err()', src/settings/toml/tests/deployments.rs:99:5
```

I think this is because they expect `get_deployments` to fail eagerly, but now this only fails in `deploy()`.

jyn514:
@caass mentioned wanting to test this with Durable Objects and KV namespaces, but I'm not sure how to do that ... happy to do so if someone can write up instructions.

src/settings/toml/tests/deployments.rs
```diff
@@ -85,19 +85,26 @@ fn it_errors_on_get_deployments_missing_name() {
...
-    test_toml.account_id = None;
-    let toml_string = toml::to_string(&test_toml).unwrap();
-    let manifest = Manifest::from_str(&toml_string).unwrap();
-
-    let environment = None;
-
-    assert!(manifest.get_deployments(environment).is_err());
-}
+// #[test]
+// fn it_errors_on_get_deployments_missing_account_id() {
```
jspspike, line 89: Maybe just delete this outright?

jspspike (approved):
Tested locally and seems to be working as intended, other than that one commented out tests it looks good to me

PR #2038, opened 2021-08-19
Move TOML table fields come later than other fields

### LINKED ISSUE #2037: Serializing manifest result in TOML error
#### Describe the bug

When trying to serialize `wrangler::settings::toml::Manifest` with some non-none `HashMap` value, the TOML serializer throws [ValueAfterTable error](https://github.com/alexcrichton/toml-rs/blob/d8a9a98cbbaa3c734e972ed8809ff62a6f4a781c/src/ser.rs#L131-L134). This is because TOML requires that table fields must come after other fields. 

#### Reproduce the bug

The following code panics.

```rust
let manifest = Manifest {
    vars: Some(vec![
        (String::from("FOO"), String::from("some value")),
    ].into_iter().collect()),
    ..Default::default()
};
println!("{}", toml::to_string(&manifest).unwrap());
```

#### Expected behavior

The manifest is correctly serialized.

#### Environment and versions

* output of `wrangler -V`:
`1.19.0`

antiphoton:
Fixes #2037

### DIFF
--- a/src/settings/toml/manifest.rs
+++ b/src/settings/toml/manifest.rs
@@ -718,9 +718,11 @@ mod tests {
     #[test]
     fn serialize() {
         let manifest = Manifest {
-            vars: Some(vec![
-                (String::from("FOO"), String::from("some value")),
-            ].into_iter().collect()),
+            vars: Some(
+                vec![(String::from("FOO"), String::from("some value"))]
+                    .into_iter()
+                    .collect(),
+            ),
             ..Default::default()
         };
         assert!(toml::to_string(&manifest).is_ok());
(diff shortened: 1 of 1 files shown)


src/settings/toml/manifest.rs
```diff
@@ -714,4 +714,15 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn serialize() {
+        let manifest = Manifest {
+            vars: Some(vec![
+                (String::from("FOO"), String::from("some value")),
+            ].into_iter().collect()),
```
mrbbot, line 723: Looks good! 🙂  Would you be able to format this with `rustfmt` so it passes the linter?
```suggestion
            vars: Some(
                vec![(String::from("FOO"), String::from("some value"))]
                    .into_iter()
                    .collect(),
            ),
```

mrbbot (changes requested):

src/settings/toml/manifest.rs
antiphoton: Done

mrbbot (approved):

antiphoton:
@mrbbot Could you re-run the GitHub actions workflows?

mrbbot (approved):

threepointone:
Thanks for this PR!

PR #2048, opened 2021-08-30
OAuth integration for Wrangler login

ocsfrank:
Still in staging, but it should be ready for review.

Please also ignore the cargo.toml, I will update it once the service goes to prod. The CI test are failing because of `oauth2` library.


nataliescottdavidson:
While you are waiting for changes to `oauth2` crate to get merged upstream, you can specify the source branch of the repository like so: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories

ocsfrank:
> While you are waiting for changes to `oauth2` crate to get merged upstream, you can specify the source branch of the repository like so: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories

I see, thanks! I just added a couple lines needed for authentication in the staging environment, so it was easier to just clone the repo and keep it private on my local machine. Once we get the service to prod, we won't need that authentication method.

threepointone:
This appears to be working well! Some feedback:
 - when it opens the browser, we log the browser url in the terminal. It's super verbose and feels unnecessary. @ocsfrank mentioned how it's for when the browser fails to open, which is a possibility in wsl. As a compromise, I suggest we only log the url in  wsl. 
 -  I manually edited the token and deleted the characters, then when I run wrangler dev I get this message 
image
but I would've expected the "invalid/corrupt token" messaging we recently added. This should probably be fixed? It shouldn't block landing this PR, it can happen in a followup PR. 

Doing just a written approval here for the working, feel free to land after a code review.

nataliescottdavidson (changes requested):

src/commands/whoami.rs
```diff
@@ -100,6 +112,8 @@ fn fetch_api_token_email(
         Err(e) => match e {
             ApiFailure::Error(_, api_errors) => {
                 let error = &api_errors.errors[0];
+                // TODO: this message also appears when access token is expired, but it has the user:read scope
```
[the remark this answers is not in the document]
ocsfrank: I see. Sorry, I forgot to remove this "TODO". I made a change so that whenever a GlobalUser is created, it will update the OAuth token if it is expired. 

I just tested the case when we use a deleted API token, and it returns an "invalid access token". So, it should be fine?

src/login/mod.rs
```diff
@@ -1,104 +1,220 @@
...
-        pubkey_encoded
-    ))?;
-
-    let encrypted_token_str = poll_token(pubkey_filtered)?;
-    let encrypted_token = base64::decode_block(encrypted_token_str.as_str())?;
-    let mut token_bytes: [u8; 128] = [0; 128];
-
-    rsa.private_decrypt(&encrypted_token, &mut token_bytes, Padding::PKCS1)?;
-    let token = str::from_utf8(&token_bytes)?.trim_matches(char::from(0));
+    if wsl::is_wsl() {
```
ocsfrank, line 68: Is there a better way to check if the current process is run on WSL without having to include another dependency?

src/commands/dev/edge/mod.rs
```diff
@@ -34,7 +35,11 @@ pub fn dev(
     let runtime = TokioRuntime::new()?;
     loop {
         let target = target.clone();
-        let user = user.clone();
+        let mut user = user.clone();
+
+        // Check if oauth token is expired
+        let _res = check_update_oauth_token(&mut user).expect("Failed to refresh access token");
```
nilslice, line 41: While it isn't likely that the user will be able to do anything if the `expect` is reached from an Err value, we may want to consider letting the `dev` caller handle this error or print something more useful here. `"Failed to refresh access token"` isn't as informative as whatever might come back from the API for this call. 

@ocsfrank could you dig in a bit and see what other options are available here? The end-goal is providing as much to the `wrangler` user as possible for debugging.. as this stands, we just print this panic message and maybe a backtrace before suggesting they `wrangler report` the issue. But, it's unlikely that we can help them, or that they can help themselves without more info.

src/login/mod.rs
```diff
@@ -1,104 +1,220 @@
...
+    let auth_code = params_values[1];
+    let recv_csrf_state = params_values[2];
+
+    // Check CSRF token to ensure redirect is legit
+    let recv_csrf_state = CsrfToken::new(recv_csrf_state.to_string());
+    if recv_csrf_state.secret() != csrf_state.secret() {
+        anyhow::bail!(
+            "Redirect URI CSRF state check failed. Received: {}, expected: {}",
+            recv_csrf_state.secret(),
+            csrf_state.secret()
```
nilslice, line 108: Are these values safe to print? I honestly have no idea, so just checking :)

src/login/mod.rs
```diff
@@ -1,104 +1,220 @@
...
+    if let Ok(user) = user_create {
+        if let GlobalUser::OAuthTokenAuth { .. } = user {
+            // Try to invalidate previous token
+            let result = revoke_token(&user);
+            if result.is_err() {
+                log::debug!("Failed to invalidate OAuth token before login.");
+            }
+        }
     }
 }
```
nilslice, line 220: Generally in this file `src/login/mod.rs` (besides tests), there is a lot of `unwrap` and `expect` usage. I think some of these are in places that will create a lot of noise from `wrangler report` usage, and could be cleaned up a bit. Consider some `?` usage in place of these where returning the error (even if it's printed later on) could give the user some way of debugging an issue. 

Many of them might be fine. Just adding a comment here to suggest taking another pass at error handling along this path.

src/settings/global_user.rs
```diff
@@ -20,16 +21,33 @@ use std::io::Write;
...
         let environment = Environment::with_whitelist(ENV_VAR_WHITELIST.to_vec());
 
         let config_path = get_global_config_path();
-        GlobalUser::build(environment, config_path)
+        let mut new_user = GlobalUser::build(environment, config_path);
+
+        // Check if oauth token is expired
+        if let Ok(ref mut oauth_user) = new_user {
+            let _res =
+                check_update_oauth_token(oauth_user).expect("Failed to refresh access token");
```
nilslice, line 48: Another case to consider more useful error handling / context-sharing with the end user.

src/settings/global_user.rs
```diff
@@ -86,40 +166,97 @@ impl GlobalUser {
...
+        //      1) (partial or complete) OAuth token + API token
+        //      2) (partial or complete) OAuth token + Global API key (partial or complete)
+        //      3) Invalid authentication methods (e.g. partial Oauth token, partial Global API key, and empty configuration file + no environment variables)
+        // API token has priority over global API key both in environment variables and in configuration file
+        if (api_token.is_ok()
+            && (oauth_token.is_ok() || refresh_token.is_ok() || expiration_time.is_ok()))
+            || ((oauth_token.is_ok() || refresh_token.is_ok() || expiration_time.is_ok())
+                && (email.is_ok() || api_key.is_ok()))
+        {
+            let error_info = "\nMore than one authentication method (e.g. API token and OAuth token, or OAuth token and Global API key) has been found in the configuration file. Please use only one.";
```
nilslice, line 198: We can probably add more detail to this error:
- print the path to the config file in case they want to manually edit it
- suggest running `wrangler logout` (assuming this clears _all_ auth info on disk?)

src/settings/global_user.rs
```diff
@@ -86,40 +166,97 @@ impl GlobalUser {
...
+            Ok(Self::GlobalKeyAuth {
+                email: email.expect("Failed to read email"),
+                api_key: api_key.expect("Failed to read api_key"),
+            })
+        } else if oauth_token.is_ok() && refresh_token.is_ok() && expiration_time.is_ok() {
+            Ok(Self::OAuthTokenAuth {
+                oauth_token: oauth_token.expect("Failed to read OAuth token"),
+                refresh_token: refresh_token.expect("Failed to read OAuth refresh token"),
+                expiration_time: expiration_time
+                    .expect("Failed to read access token expiration time"),
```
nilslice, line 214: I think all of these `expect` calls can just be `unwrap`, since you've already attested the are `Ok` in the if checks

src/settings/global_user.rs
```diff
@@ -86,40 +166,97 @@ impl GlobalUser {
...
+            })
+        } else {
+            // Empty configuration file and no environment variables, or missing variable for global API key
+            let error_info = "\nNo (valid) authentication method has been found.";
```
nilslice, line 218: Could add more detail here too.. could running `wrangler login` help? let's offer some suggestions for a user to try.

nilslice (changes requested):

src/settings/global_user.rs
ocsfrank: Yes! That would help, and it's being printed in the next line with the call to `            Self::show_config_err_info(Some(error_info.to_string()), config)`.

src/login/mod.rs
ocsfrank: Nice catch! I think it's possible for someone to use the previous `printed URL + the expected csrf_state` to get an authorization grant. Better be safe and remove it

src/login/http.rs
```diff
@@ -87,23 +89,24 @@ pub async fn http_server_get_params() -> Result<String> {
...
+            Err(_) => anyhow::bail!(display_error_info("Local HTTP Server binding failed.")),
         }
     });
 
     // Receive authorization code and csrf state from HTTP server
-    let params = runtime.block_on(async {
+    let params = tokio::task::spawn(async move {
         match rx.recv().await {
             Some(values) => values,
             None => "error".to_string(),
```
nilslice, line 106: since we have a history of difficult-to-debug channel errors, maybe we could be more descriptive and log this error (unless it is already being logged somewhere in the calling context). 

do you know if it is possible that the `rx` can receive `None` at any point _before_ it gets `Some(values)` and should therefore be polled?

src/settings/global_user.rs
```diff
@@ -202,21 +202,23 @@ impl GlobalUser {
...
             let more_info = format!("{}\nIf you'd like to edit the configuration file, it can be found at {}. Consider also running {} to clean up the configuration file.", error_info, config_path.to_str().unwrap(), wrangler_logout_msg);
 
-            Self::show_config_err_info(Some(more_info.to_string()), config)
-        } else if api_token.is_ok() {
+            Self::show_config_err_info(Some(more_info), config)
+        } else if let Ok(api_token_value) = api_token {
             Ok(Self::ApiTokenAuth {
-                api_token: api_token.unwrap(),
+                api_token: api_token_value,
             })
```
nilslice, line 207: Nit: these could be re-written to be a bit more concise: 

```rust
} else if let Ok(api_token) = api_token {
            Ok(Self::ApiTokenAuth { api_token})
// ... 
```

the same is true for all the branches below. if the variable binding (in this case inside the if let, shadowing the outer one) is the same as the field name, you can omit the field name and just pass in the variable name in its place. Not major, but might make the code a bit nicer to read.

src/login/http.rs
ocsfrank: Logging is a good idea. The only case that I am aware of is when the sender side is closed for any reason, and `rx` still tries to read something. In this case, it should receive `None`.

src/settings/global_user.rs
ocsfrank: I see! Will include these changes in the next commit.

nilslice (approved):
Nice work!

threepointone (approved):

nataliescottdavidson (approved):
I have verified that the changes work as expected on my machine.

PR #2055, opened 2021-09-01
fix: override rust type to javascript

nilslice:
the `rust` type seems to only have been added as a way to automatically trigger wasm-pack, but it's much better to put that in a custom `[build]` step instead. eventually, we really should phase out the `rust` type from `wrangler` altogether, but this at minimum makes the new rust template work.

### DIFF
--- a/src/settings/toml/manifest.rs
+++ b/src/settings/toml/manifest.rs
@@ -143,7 +143,11 @@ impl Manifest {
             config_template_doc["workers_dev"] = toml_edit::value(default_workers_dev);
         }
         if let Some(target_type) = &target_type {
-            config_template_doc["type"] = toml_edit::value(target_type.to_string());
+            if target_type.to_string() == "rust" {
+                config_template_doc["type"] = toml_edit::value(TargetType::JavaScript.to_string());
+            } else {
+                config_template_doc["type"] = toml_edit::value(target_type.to_string());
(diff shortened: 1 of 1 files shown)


nilslice:
@threepointone, nope! this only takes affect when calling `wrangler generate`. all existing `type = rust` projects remain unaffected.

threepointone:
But new rust projects won't use wasm-pack? As in, they'll work differently from what they work like before? (Sorry for the noob question, just wondering if this is a breaking change). Rust projects will have to do their own compilation after this, and will behave like a 'javascript' (aka no build) project? 

(Long term I agree we should do away with 'type' completely)

nilslice:
The build for new rust projects uses a "custom build" command, and is not compatible with old-style rust projects. But this change doesn't affect any old projects who still use `type = "rust"`. Those will still be recognized by wrangler build and continue to compile. 

With this change, only new projects which are generated with `--type rust` will have different behavior.

nilslice:
> Rust projects will have to do their own compilation after this, and will behave like a 'javascript' (aka no build) project?

Yes, but in the new template, we provide a new build command that handles the compilation specifically how `workers-rs` projects need to be built.

threepointone (approved):

mrbbot (approved):

PR #2057, opened 2021-09-02
1.19.2

nilslice:
Also includes a small fix to the `changelog-generator` which failed when I ran it due to a `pr.body` prop being null. This just does a presence check on that prop before it's accessed and then ultimately added to an array for output.


threepointone (approved):
🚢

Issue #2054, opened 2021-09-01
Improve error message for invalid auth token


jkup:
#### Describe the feature

I ran into a problem where I had an invalid auth_token in `~/.wrangler/config/default.toml`. I didn't know that file existed and whenever I would try to run a command like `wrangler dev` I would get:

`Error: ⚠️  Code 9109: Invalid access token`

Which wasn't super helpful. @threepointone helped me debug and showed me where the file was. Deleting it fixed my problem! Maybe we could update that message? Suggesting users delete that token? Or maybe even provide a `wrangler logout` command which will clean things up?

#### Describe the alternatives

Deleting the file and trying again worked great. I just didn't know the file existed.

ocsfrank:
Looks like you read into my mind haha! I'm currently adding a logout feature for the OAuth support.

ocsfrank:
Error message should be improved with #2060.

threepointone:
Maybe this can be closed now?

ocsfrank:
Yes, this should be solved.

PR #2060, opened 2021-09-02
Improve `wrangler dev` error message for invalid tokens.

ocsfrank:
This should display more info when dev fails(issue #2054) because of invalid tokens or invalid request headers.

### DIFF
--- a/src/commands/whoami.rs
+++ b/src/commands/whoami.rs
@@ -125,7 +125,7 @@ pub(crate) fn fetch_accounts(user: &GlobalUser) -> Result<Vec<Account>> {
                         StdOut::info("Your API token might be expired, or might not have the necessary permissions. Please re-authenticate wrangler by running `wrangler login` or `wrangler config`.");
                     } else if error.code == 6003 {
                         // 6003 error code = Invalid request headers. A common case is when the value of an authorization method has been changed outside of wrangler commands
-                        StdOut::info("Your authentication method might be corrupted (e.g. API token value has been altered). Please consider re-authenticating wrangler through `wrangler login` or `wrangler config`.")
+                        StdOut::info("Your authentication method might be corrupted (e.g. API token value has been altered). Please re-authenticate wrangler by running `wrangler login` or `wrangler config`.")
                     }
 
                 }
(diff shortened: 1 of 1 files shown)


src/commands/whoami.rs
```diff
@@ -116,7 +116,24 @@ pub(crate) fn fetch_accounts(user: &GlobalUser) -> Result<Vec<Account>> {
...
     match response {
         Ok(res) => Ok(res.result),
-        Err(e) => anyhow::bail!(http::format_error(e, None)),
+        Err(e) => {
+            match e {
+                ApiFailure::Error(_, ref api_errors) => {
+                    let error = &api_errors.errors[0];
+                    if error.code == 9109 {
+                        // 9109 error code = Invalid access token
+                        StdOut::info("Token might be expired or might not have the necessary permissions. Please consider running `wranger logout` or re-authenticate wrangler through `wrangler login` or `wrangler config`.");
```
threepointone, line 125: This will require https://github.com/cloudflare/wrangler/pull/2048 to land so wrangler logout can work, yes?

threepointone, line 125:
```suggestion
                        StdOut::info("Your API token might be expired, or might not have the necessary permissions. Please re-authenticate wrangler by running `wrangler login` or `wrangler config`.");
```
Maybe we shouldn't even suggest logging out, if wrangler login will wipe out the previous token?

src/commands/whoami.rs
```diff
@@ -116,7 +116,24 @@ pub(crate) fn fetch_accounts(user: &GlobalUser) -> Result<Vec<Account>> {
...
+                    } else if error.code == 6003 {
+                        // 6003 error code = Invalid request headers. A common case is when the value of an authorization method has been changed outside of wrangler commands
+                        StdOut::info("Your authentication method might be corrupted (e.g. API token value has been altered). Please consider re-authenticating wrangler through `wrangler login` or `wrangler config`.")
```
threepointone, line 128: Similar suggestion here. Maybe we shouldn't recommend logging out, when they'll have to do a login again anyway (and if login wipes out the previous token)

threepointone (approved):
Approving with suggestions.

src/commands/whoami.rs
ocsfrank: Oops. Yes, wrangler login will completely override the previous token.

PR #2061, opened 2021-09-03
fix: remove panic when send fails, log error instead

nilslice:
Removes the panic on send failure from the last remaining place for this particular `--watch` error.

### DIFF
--- a/src/wranglerjs/mod.rs
+++ b/src/wranglerjs/mod.rs
@@ -109,7 +109,9 @@ pub fn run_build_and_watch(target: &Target, tx: Option<Sender<()>>) -> Result<()
                     if write_wranglerjs_output(&bundle, &wranglerjs_output, custom_webpack).is_ok()
                     {
                         if let Some(tx) = tx.clone() {
-                            tx.send(()).expect("--watch change message failed to send");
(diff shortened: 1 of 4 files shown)


src/wranglerjs/mod.rs
```diff
@@ -109,7 +109,9 @@ pub fn run_build_and_watch(target: &Target, tx: Option<Sender<()>>) -> Result<()
                     if write_wranglerjs_output(&bundle, &wranglerjs_output, custom_webpack).is_ok()
                     {
                         if let Some(tx) = tx.clone() {
-                            tx.send(()).expect("--watch change message failed to send");
+                            if tx.send(()).is_err() {
+                                log::error!("wranglerjs watch operation failed to notify");
```
jyn514, line 113: Can you also print the error instead of discarding it?
```suggestion
                                log::error!("wranglerjs watch operation failed to notify: {}", err);
```
And change it to `if let` on the line above.

nilslice: Good catch - thanks. Also update some other locations with similar behavior.

jyn514 (approved):

PR #2064, opened 2021-09-08
Allow the API endpoint to be configured via an environment variable

ocsfrank:
Adds the option to configure the API endpoint through an environment variable.
For example, `export ENDPOINT_BASE_URL=some_url`.

### DIFF
--- a/src/http/cf.rs
+++ b/src/http/cf.rs
@@ -1,3 +1,4 @@
+use std::env;
 use std::time::Duration;
 
 use cloudflare::framework::async_api;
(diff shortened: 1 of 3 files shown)


src/http/cf.rs
```diff
@@ -12,17 +13,37 @@ use crate::http::{feature::headers, DEFAULT_HTTP_TIMEOUT_SECONDS};
 use crate::settings::global_user::GlobalUser;
 use crate::terminal::emoji;
 use crate::terminal::message::{Message, StdOut};
+
+const ENDPOINT_URL: &str = "ENDPOINT_URL";
```
nilslice, line 17: Nit: but could you rename this to something like `ENDPOINT_BASE_URL` ? Otherwise it sort of sounds like you're able to control the entire URL, not just the host / base of the URL.

ocsfrank: Makes sense. Updated in 54b4860.

nilslice (approved):
LGTM

nilslice (changes requested):
Sorry for the revert here.. I think we should also change the env var name to further specify it, in case there is overlap with another env var set on the system. 

`ENDPOINT_BASE_URL` could be for anything, but `CF_API_BASE_URL` is probably as specific as we need and is pretty clear about what it is expected to be and when it should be set. 

What do you think?

ocsfrank:
> Sorry for the revert here.. I think we should also change the env var name to further specify it, in case there is overlap with another env var set on the system.
> 
> `ENDPOINT_BASE_URL` could be for anything, but `CF_API_BASE_URL` is probably as specific as we need and is pretty clear about what it is expected to be and when it should be set.
> 
> What do you think?

I also agree that we should be as specific as possible. We never know what other environment variables could be present. 👍

nilslice (approved):

PR #2071, opened 2021-09-15
Frank/fix dev refresh preview

Electroid:


src/commands/dev/edge/watch.rs
```diff
@@ -38,15 +38,26 @@ pub fn watch_for_changes(
...
+                    // if the API error code is 10049, then it's an expired preview token
+                    if err.0.contains("10049") {
+                        refresh_session_channel.send(Some(()))?;
+                        break;
+                    }
+
+                    // otherwise it is a non recoverable error
+                    StdOut::warn(&format!(
+                        "{}\nPlease terminate `wrangler dev` using Ctrl+C.",
+                        &err.0
```
Electroid, line 57: Hmm, can we not restart it for the user? That's essentially what it use to do.

Cargo.lock
```diff
@@ -498,15 +498,6 @@ version = "0.8.2"
...
 checksum = "ea221b5284a47e40033bf9b66f35f984ec0ea2931eb03505246cd27a963f981b"
 
-[[package]]
-name = "cpufeatures"
-version = "0.1.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "66c99696f6c9dd7f35d486b9d04d7e6e202aa3e8c40d553f2fdf5e7e0c6a71ef"
-dependencies = [
- "libc",
-]
```
jyn514, line 508: This seems like a good change, but I'm not sure what it has to do with this PR?

src/commands/dev/edge/watch.rs
jyn514, line 57: Shouldn't this be a hard error?
```suggestion
                    // otherwise it is a non recoverable error
                    anyhow::bail!(err.0);
```

mrbbot: The reason this wasn't a hard error initially was because if the user writes some code with a syntax error for instance (and we get a 400 bad request response), we still want wrangler to watch the file for changes and then try upload the code again when the user fixes their code.

ocsfrank: Yes, I believe we can do that. Fixing

ocsfrank: Not sure what all the possible API errors are, but we can add a case for wrangler to still continue watching if there's a syntax error.

ocsfrank: I tried `anyhow::bail!(err.0);`, but it will not shut down `wrangler dev` as it will just return from the spawned thread.

src/commands/dev/edge/watch.rs
```diff
@@ -51,11 +51,14 @@ pub fn watch_for_changes(
                     if err.0.contains("10049") {
                         refresh_session_channel.send(Some(()))?;
                         break;
+                    } else if err.0.contains("10021") {
+                        // Continue to watch for changes if it's a syntax error
+                        StdOut::warn(&format!("{}\nPlease update the your code.", &err.0));
```
jyn514: Is the additional context necessary? It would be helpful to give an example here of what the upstream error looks like.

Either way, this isn't grammatical.
```suggestion
                        StdOut::warn(&format!("{}\nPlease update your code.", &err.0));
```

jyn514 (changes requested):

src/commands/dev/edge/watch.rs
ocsfrank: After formatting, the error looks like this:
```
⚠️  Something went wrong with the request to Cloudflare...
Uncaught SyntaxError: Unexpected token 'return'
  at line 9
 [API code: 10021]
```

threepointone:
Please add a description to this PR.

threepointone (approved):

src/commands/dev/edge/watch.rs
```diff
@@ -38,15 +38,29 @@ pub fn watch_for_changes(
...
                 if let Some(err) = err.downcast_ref::<setup::BadRequestError>() {
-                    // no need to refresh session here, bad request indicates problem with user code
-                    StdOut::warn(&err.0);
+                    // if the API error code is 10049, then it's an expired preview token
+                    if err.0.contains("10049") {
+                        refresh_session_channel.send(Some(()))?;
+                        break;
+                    } else if err.0.contains("10021") {
+                        // Continue to watch for changes if it's a syntax error
+                        StdOut::warn(&format!("{}\nPlease update your code.", &err.0));
```
threepointone, line 56: Checking, does err contain the actual syntax error detected, with a line number, and will it get logged?

ocsfrank: Screen Shot 2021-09-15 at 2 37 30 PM

Yes, it shows the syntax error and line number. This is printed to `stdout`, and the `reqwest` library logs the HTTP response received from the API.

jyn514 (approved):

PR #2080, opened 2021-09-20
Only warn about compatibility dates on publish

### LINKED ISSUE #2078: Only warn about compatibility date for `wrangler publish`
We only need to poke about the `compatibility_date` when the developer is running `wrangler publish`, not the other commands.

Electroid:
Changes the compatibility date warning to only occur when using `wrangler publish`. We do not want this warning to be sent when using other commands like `wrangler tail`, etc.

Closes #2078

### DIFF
--- a/src/cli/publish.rs
+++ b/src/cli/publish.rs
@@ -28,6 +28,11 @@ pub fn publish(
 
     log::info!("Getting project settings");
     let manifest = Manifest::new(&cli_params.config)?;
+
+    // Only warn about a missing compatibility date on publish,
+    // to avoid spamming other commands like `subdomain` and `tail` where it is not relevant.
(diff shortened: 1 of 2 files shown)


threepointone (commented):
- We should probably have this for dev too, or you risk getting divergent behaviour.
- This should probably be an error on publish, not a warning (or folks who don't manually publish won't even see it)

Electroid:
Hm, I don't think we should error on publish. While this calls for action, it's not urgent or important enough to break builds. If we were deprecated something that might be one thing, but this will continue to work. I think a warning is fine.

ocsfrank (approved):
Looks good to me!

threepointone:
Could you add the warning for `dev` as well? Let's merge this in and release it right after.

PR #2082, opened 2021-09-21
update codeowners with current team

nataliescottdavidson:


threepointone (approved):

jyn514:
Why do this per person instead of adding people to Cloudflare/wrangler? This was we'll have to update both the GitHub team and the CODE OWNERS file when someone joins or leaves.

Issue #210, opened 2019-06-04
How to unpublish a Workers script


DrDaveD:
Now that there can be any number of Workers scripts, how can we delete one we no longer want (in my own domain)?  I don't find it in the documentation or in wrangler --help.

DrDaveD:
Ah, I found it in the web api of dash.cloudflare.com, under "...".  It required me to create a workers.dev domain first, which I didn't think I was going to need, but it works.  Anyway it would probably be good to document it and/or add a CLI interface.

xtuc:
There no subcommand in Wrangler to do it. I would suggest going to the UI and deleting what script/route you don't need anymore.

ashleygwilliams:
we can implement this in the CLI relatively simply. i think this is an excellent feature request- let's plan on it for 1.1.0.

we can "unpublish" in 2 ways:
- we can delete the script (workers.dev or custom domain)
- we can make the script not public on a domain (workers.dev)

i propose:

`workers unpublish`
- workers.dev, removes being public on subdomain
- custom domain, deletes route script is associated with 

`workers delete`
- for both workers.dev and custom domain, deletes the script from the account

thoughts?

xtuc:
I would personally be against unpublishing. Deleting a domain, route or a script is dangerous, if you do it on mistake it will take down your prod. 

Now that workers are free, I don't see a reason to urgently unpublish a worker. 

I'm happy with people doing that in UI tho.

EverlastingBugstopper:
Perhaps you could have an "Are you sure" prompt that requires you to type in the name of the script again. I think this would be useful.

leo-ar:
This would be a convenient feature. I was about to open a feature request for it.
Right now I'm accomplishing the same thing by scripting requests to the API. In addition to removing the scripts the KVs have to be removed too.
Although writing these scripts it's a one time cost, it'd be convenient not having to do it 😸

alonsohki:
This is a key feature to have for deployment pipelines where feature branches are automatically deployed for preview. What's the current status?

yafanasiev:
+1. We are thinking about moving some of our apps to worker sites, and not having a convenient way to remove the feature branches deployments is a little discouraging.

halzy:
This would be a fantastic! I have an app that is completely on CF Workers and having to log into the UI to delete the CI environments that are spun up is an annoyance. If there is a confirmation step, please also have a `-y` flag to automatically approve it.

nhoughto:
[...]

PR #2086, opened 2021-09-24
Add `wrangler delete` for scripts

ocsfrank:
This PR adds a command `wrangler delete` that allows to delete a Workers script.

`wrangler delete` will delete a script based on the script name and account ID found in the `wrangler.toml` file. 

Without any flags, the command will ask for confirmation steps when deleting the script, and any bound Durable Objects.
It's also possible to pass a flag `--force` to skip all the confirmation steps. **WARNING:** `--force` **will also delete the Durable Objects bound to a script.** 

Should solve #210.


src/cli/delete.rs
```diff
@@ -0,0 +1,20 @@
...
+    force: bool,
+    account_id: Option<String>,
+    script_id: Option<String>,
+) -> Result<(), anyhow::Error> {
+    let user = GlobalUser::new()?;
+    if no_interactive {
+        if let (Some(account_id), Some(script_id)) = (account_id, script_id) {
+            commands::delete::delete_script(&user, force, &account_id, &script_id)
+        } else {
+            anyhow::bail!("Both account-id and script-id must be provided.")
```
Electroid, line 15: I think `script-name` would be more consistent with how we reference this elsewhere.

src/cli/mod.rs
```diff
@@ -201,6 +203,39 @@ pub enum Command {
         migration: AdhocMigration,
     },
 
+    /// Delete a Workers script from an account.
+    #[structopt(name = "delete")]
+    Delete {
+        /// Disables interactive mode
+        #[structopt(long = "no-interactive", short = "n")]
```
Electroid, line 210: `non-interactive`

src/cli/mod.rs
```diff
@@ -201,6 +203,39 @@ pub enum Command {
...
+        )]
+        force: bool,
+        /// Account ID to delete the script from. Required if interactive mode is disabled.
+        #[structopt(
+            long = "account-id",
+            short = "aid",
+            requires = "no-interactive",
+            requires = "script-id"
+        )]
+        account_id: Option<String>,
```
Electroid, line 228: Even in non-interactive mode, we we use the default `account_id`? (from the credentials) If there are more than 1 options, then we can require it.

src/commands/delete/mod.rs
```diff
@@ -0,0 +1,201 @@
...
+
+// Interactive mode
+pub fn run(user: &GlobalUser) -> Result<(), anyhow::Error> {
+    // Fetch and display user accounts
+    let accounts = fetch_accounts(user)?;
+    let (valid_accounts, table) = format_accounts(accounts)?;
+    println!("{}", &table);
+    let account_name = interactive::get_user_input(
+        "Please enter the name of the account you wish to delete a Workers script from",
+    );
```
Electroid, line 42: Can this be skipped if they have 1 account?

Electroid (changes requested):

src/cli/mod.rs
jyn514: Hmm, wrangler uses the account_id from `wrangler.toml` everywhere else, right? Why have it as a command line parameter here instead?

src/commands/delete/mod.rs
```diff
@@ -0,0 +1,201 @@
...
+    if !valid_accounts.contains_key(&account_name) {
+        anyhow::bail!("Account name doesn't match.")
+    }
+
+    //  Fetch and display scripts related to the account
+    let account_id = valid_accounts.get(&account_name).unwrap();
```
jyn514, line 48: This seems really different from how IDs are handled everywhere else. This should use `manifest.account_id.load()` instead.

jyn514 (changes requested):

src/cli/mod.rs
ocsfrank: The idea I had in mind was to let users run `wrangler delete` from any directory even if they don't have a Workers script on their machine.

nilslice: > The idea I had in mind was to let users run wrangler delete from any directory even if they don't have a wrangler.toml.

I can see this being a nice feature, but since we don't support deploying a project w/o a `wrangler.toml`, then it's probably safer & more consistent to maintain the same restriction when deleting. 

It is not uncommon for other deployment systems to _reference_ code artifacts to deploy (or tear down) without having the artifact / source locally though. So I do see this as generally a good idea, but maybe not for now for wrangler as it stands.

ocsfrank: > > The idea I had in mind was to let users run wrangler delete from any directory even if they don't have a wrangler.toml.
> 
> I can see this being a nice feature, but since we don't support deploying a project w/o a `wrangler.toml`, then it's probably safer & more consistent to maintain the same restriction when deleting.
> 
> It is not uncommon for other deployment systems to _reference_ code artifacts to deploy (or tear down) without having the artifact / source locally though. So I do see this as generally a good idea, but maybe not for now for wrangler as it stands.

I see, that makes a lot of sense. Thanks everyone for the feedback!

ocsfrank:
Changed to read all project info from `wrangler.toml` and added a PR to `cloudflare-rs` (https://github.com/cloudflare/cloudflare-rs/pull/166) for adding the API endpoints.

JacobMGEvans:
Thanks for creating this PR! Wrangler v1 is now deprecated and support is only being provided for critical updates or security concerns. As such, we are closing this PR.

New versions of Wrangler are maintained in the [workers-sdk repo](https://github.com/cloudflare/workers-sdk). If you are running into a similar issue with wrangler v2, please report it or create a new PR in the [workers-sdk repo](https://github.com/cloudflare/workers-sdk). For more info about wrangler v1 deprecation, please check out our [blog post](https://blog.cloudflare.com/deprecating-wrangler-v1).

Issue #2095, opened 2021-10-07
KV bulk delete override user confirm input


andersquist:
#### Describe the feature

Adding a flag so that `wrangler kv:bulk delete` don't ask for user confirmation when deleting key value pairs. This would be really nice to have in an automation use case.

#### Describe the alternatives

Using the Cloudflare API and do the bulk delete that way. It works but it would be nice if one could get things done solely using the Wrangler CLI.

threepointone:
Happy to accept a PR that adds a `--force` flag for this operation.

threepointone:
We released this in 1.19.4, thanks for the PR!

PR #2100, opened 2021-10-09
feat: force flag for kv bulk delete command to avoid user confirmation prompt

andersquist:
Adresses  #2095.  Make it possible to override any user confirmation prompt when performing KV bulk deletes.


threepointone:
This looks good! I'll do a proper review early next week, thanks so much!

PR #2114, opened 2021-10-31
add __STATIC_CONTENT_MANIFEST in vars for module workers

g45t345rt:
The `__STATIC_CONTENT_MANIFEST` is no accessible when using the module worker site.

This pr includes the manifest in the `env` variables.
```ts
export default async (request: Request, env: EnvInterface, ctx) => {
    const manifest = JSON.parse(env.__STATIC_CONTENT_MANIFEST)
}
```

Related issues
https://github.com/cloudflare/wrangler/issues/1938
https://github.com/cloudflare/wrangler/pull/1973


threepointone (approved):
This looks good to me! Will land after it passes CI. Thanks for this!

g45t345rt:
Holy smokes, I honestly did not think this was going to land.

Right now, I'm using a custom wrangler build just for publishing modules workers with static site assets.

Thanks a lot!

threepointone:
does it actually work as expected btw? I'll need to test it locally before merging this, but can you confirm that you can see the manifest on the env object?

g45t345rt:
Yes it works as expected.

image

Now that I think about it, it might not be a good idea :S
What if the manifest is large? If the manifest is more than 1kb it will not store the value.

I might make another pull request that works with textblobs instead

```ts
import manifest from '__STATIC_CONTENT_MANIFEST`
```

threepointone:
I don't know if that would be a good idea/how it would work with es module workers. I'll think about this some more.

threepointone (changes requested):
Yeah thinking about this some more, this can't be an env var. Sorry :( Thinking about it some more.

g45t345rt:
No worries I'll work on an alternative.

dcousens:
@threepointone for reference, in https://github.com/cloudflare/wrangler/issues/1938 (pull request as https://github.com/cloudflare/wrangler/pull/1973), @xortive [mentioned this could be supported as](https://github.com/cloudflare/wrangler/issues/1938#issuecomment-867200289) a module that can be imported.

@g45t345rt what were your thoughts on that pull request?

g45t345rt:
@dcousens The pr is great but I don't think it will be merged.

This pr import all text_blobs as a module (including __STATIC_CONTENT_MANIFEST)
https://github.com/cloudflare/wrangler/pull/2126

PR #2118, opened 2021-11-03
Fix missing fmt argument in error message

dtolnay:
- The first commit fills in a missing argument to `{}` in config.rs. Without this, the error message would literally be 'Your token has status "{}"' with curly braces in it instead of the token status.

- The second commit closes an inconsistency between how two places were doing bail macro args: the one in delete.rs used single `{` assuming the string would *not* get put into std::fmt, while the one in put.rs used double `{{` assuming that std::fmt *would* get applied to it. This change makes room for the bail macro to catch issues like this statically in a future version.

- The third commit is done by rustfmt; I am guessing the long line changed by the second commit previously made rustfmt fail out without formatting anything in the whole `match` expression.

### DIFF
--- a/src/commands/config.rs
+++ b/src/commands/config.rs
@@ -59,7 +59,7 @@ pub fn validate_credentials(user: &GlobalUser) -> Result<()> {
                 if success.result.status == "active" {
                     Ok(())
                 } else {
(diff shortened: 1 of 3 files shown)


jyn514 (approved):
Thanks for the fixes! Excited to hear that you're thinking of preventing this statically :)

Issue #2117, opened 2021-11-01
Tracking issue for fixing `cargo audit` issues


caass:
the output of `cargo audit` gives the following:

<details><summary>Click to expand</summary>

```sh
❯ cargo audit
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 370 security advisories (from /Users/cass/.cargo/advisory-db)
    Updating crates.io index
    Scanning Cargo.lock for vulnerabilities (350 crate dependencies)
Crate:         chrono
Version:       0.4.19
Title:         Potential segfault in `localtime_r` invocations
Date:          2020-11-10
ID:            RUSTSEC-2020-0159
URL:           https://rustsec.org/advisories/RUSTSEC-2020-0159
Solution:      No safe upgrade is available!
Dependency tree: 
chrono 0.4.19
├── wrangler 1.19.4
├── toml_edit 0.2.1
│   └── wrangler 1.19.4
├── sloggers 1.0.1
│   └── cloudflare 0.8.7
│       └── wrangler 1.19.4
├── slog-term 2.8.0
│   ├── sloggers 1.0.1
│   └── cloudflare 0.8.7
├── plist 0.5.5
│   └── os-version 0.1.1
│       └── wrangler 1.19.4
├── oauth2 4.1.0
│   └── wrangler 1.19.4
└── cloudflare 0.8.7

Crate:         time
Version:       0.1.44
Title:         Potential segfault in the time crate
Date:          2020-11-18
ID:            RUSTSEC-2020-0071
URL:           https://rustsec.org/advisories/RUSTSEC-2020-0071
Solution:      Upgrade to >=0.2.23
Dependency tree: 
time 0.1.44
├── zip 0.5.13
│   └── binary-install 0.0.3-alpha.1
│       └── wrangler 1.19.4
├── syncbox 0.2.4
│   └── eventual 0.1.7
│       └── wrangler 1.19.4
├── eventual 0.1.7
└── chrono 0.4.19
    ├── wrangler 1.19.4
    ├── toml_edit 0.2.1
    │   └── wrangler 1.19.4
    ├── sloggers 1.0.1
    │   └── cloudflare 0.8.7
    │       └── wrangler 1.19.4
    ├── slog-term 2.8.0
    │   ├── sloggers 1.0.1
    │   └── cloudflare 0.8.7
    ├── plist 0.5.5
    │   └── os-version 0.1.1
    │       └── wrangler 1.19.4
[...]

PR #2122, opened 2021-11-03
Bump os-version to 0.2.0

a1ien:
In 0.2 `os-version` use new stable version of `plist`
In plist 1.2.2(which is not published yet) remove chrono dependency and use time 0.3 which fix part of #2117


nataliescottdavidson (approved):

threepointone:
Thanks for this!

PR #2123, opened 2021-11-03
Remove serde-hjson transitive dependency

a1ien:
serde-hjson crate is not maintained anymore
With removing this dependency we also remove duplicate of old serde
Also in new version of `config` crate hjson also removed


a1ien:
Or maybe we want only toml config?

nataliescottdavidson (approved):

threepointone:
Thanks for this!

PR #2124, opened 2021-11-04
Update hash comment to changed XxHash64 usage

dcousens:
As introduced in https://github.com/cloudflare/wrangler/pull/1221

### DIFF
--- a/src/sites/mod.rs
+++ b/src/sites/mod.rs
@@ -260,7 +260,7 @@ fn generate_url_safe_path(path: &Path) -> Result<String> {
 // Adds the XXhash hash of the path's file contents to the url-safe path of a file to
 // generate a versioned key for the file and its contents. Returns the url-safe path prefix
 // for the key, as well as the key with hash appended.
-// e.g (sitemap.xml, sitemap.ec717eb2131fdd4f.xml)
+// e.g (sitemap.xml, sitemap.ec717eb213.xml)
 pub fn generate_path_and_key(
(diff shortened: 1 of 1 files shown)


src/sites/mod.rs
```diff
@@ -257,10 +257,10 @@ fn generate_url_safe_path(path: &Path) -> Result<String> {
     Ok(path.to_string())
 }
 
-// Adds the SHA-256 hash of the path's file contents to the url-safe path of a file to
+// Adds the XXhash hash of the path's file contents to the url-safe path of a file to
 // generate a versioned key for the file and its contents. Returns the url-safe path prefix
 // for the key, as well as the key with hash appended.
-// e.g (sitemap.xml, sitemap.ec717eb2131fdd4fff803b851d2aa5b1dc3e0af36bc3c8c40f2095c747e80d1e.xml)
+// e.g (sitemap.xml, sitemap.ec717eb2131fdd4f.xml)
```
dcousens, line 263:
```suggestion
// e.g (sitemap.xml, sitemap.ec717eb213.xml)
```
Actually this is truncated to 10 characters on L281

threepointone:
This would break existing sites and churn installations. I don't think we should do this .

threepointone (changes requested):
As mentioned.

threepointone:
Oh wait, this doesn't change the implementation

threepointone (approved):
Thanks for this!

PR #2126, opened 2021-11-05
add __STATIC_CONTENT_MANIFEST as a text module for modules workers

g45t345rt:
The `__STATIC_CONTENT_MANIFEST` is not accessible when using modules workers `[sites]`

This pr makes it possible to import the content of manifest as text.

It is also in favor of storing the manifest in vars
https://github.com/cloudflare/wrangler/pull/2114

Here is how to get the content
```ts
import manifest from '__STATIC_CONTENT_MANIFEST'

export default {
  fetch(request, env, ctx) {
    console.log(JSON.parse(manifest))
  }
}
```

If you are developing with Miniflare you need

```ts
new Miniflare({
  modulesRules: [{ type: 'Text', include: ['__STATIC_CONTENT_MANIFEST'] }],
  ...
})
```
or

```console
miniflare --modules-rule "Text=__STATIC_CONTENT_MANIFEST"
```

You also have to create a file named `__STATIC_CONTENT_MANIFEST` in `/dist`

Plus if you are using TS you have to declare the module like this

```ts
declare module '__STATIC_CONTENT_MANIFEST' {
  const manifest: string
  export default manifest
}
```

It's kind of pain to setup if you are developing locally...

Related issues
https://github.com/cloudflare/wrangler/issues/1938
https://github.com/cloudflare/wrangler/pull/1973


threepointone:
cc @mrbbot can we automate this module creation for miniflare?

threepointone (approved):
Landing this. We'll likely have to do a release of kv-asset-handler as well so the globals don't break the runtime.

PR #2127, opened 2021-11-08
Fix some migration errors

xortive:
* Don't include migrations in Target for preview/dev uploads

### DIFF
--- a/src/settings/toml/manifest.rs
+++ b/src/settings/toml/manifest.rs
@@ -27,7 +27,7 @@ use crate::settings::toml::route::RouteConfig;
 use crate::settings::toml::site::Site;
 use crate::settings::toml::target_type::TargetType;
 use crate::settings::toml::triggers::Triggers;
-use crate::settings::toml::{Target, UploadFormat};
+use crate::settings::toml::Target;
 use crate::terminal::{
     emoji,
     message::{Message, StdOut},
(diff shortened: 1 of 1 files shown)


src/settings/toml/manifest.rs
```diff
@@ -349,6 +349,12 @@ impl Manifest {
                         if build.command.is_none() {
                             anyhow::bail!(error_message)
                         }
+                        if matches!(build.upload, UploadFormat::Modules { .. }) {
```
threepointone, line 352: This might not be true as of later tonight. We're landing https://github.com/cloudflare/wrangler/pull/2126, and making a release of https://github.com/cloudflare/kv-asset-handler and will update the docs showing hwo it works with es modules.

src/settings/toml/manifest.rs
```diff
@@ -27,7 +27,7 @@ use crate::settings::toml::route::RouteConfig;
 use crate::settings::toml::site::Site;
 use crate::settings::toml::target_type::TargetType;
 use crate::settings::toml::triggers::Triggers;
-use crate::settings::toml::Target;
+use crate::settings::toml::{Target, UploadFormat};
```
threepointone, line 30:
```suggestion
use crate::settings::toml::Target;
```

threepointone (approved):
Tentatively approving, the suggestion should fix the lint error too.

PR #2129, opened 2021-11-09
Remove unneeded heartbeat logic with sketchy time addition

nataliescottdavidson:


threepointone (approved):
Tentatively approving pending the lint fix, but I also have a question - is the heartbeat not necessary anymore? What is the consequence of removing this code?

nataliescottdavidson:
> Tentatively approving pending the lint fix, but I also have a question - is the heartbeat not necessary anymore? What is the consequence of removing this code?

The expiry duration is static for v2, so it is no longer needed.

Issue #1704, opened 2020-12-29
wrangler generate template support main branch


xtuc:
Instead of default to the master branch we should check for the main branch first or allow the user to provide the branch's name

p-j:
Hi @xtuc 
I've made a pull request for that a couple of month ago, still waiting for it to be acknowledge though.
https://github.com/cloudflare/wrangler/pull/1653
Happy to take feedbacks in and help move this forward, that's actually an issue for a template I've made and use for some of my projects.

jldec:
Hi,
I'm hitting this while following the [docs for durable objects](https://developers.cloudflare.com/workers/learning/using-durable-objects) which is using https://github.com/cloudflare/durable-objects-template

```
$ wrangler generate rate-limit-actions https://github.com/cloudflare/durable-objects-template
🔧   Creating project called `rate-limit-actions`...
Error: ⛔   Git Error: failed to find branch `master`                                                                                                     
Error: tried running command:
/Users/jleschner/Library/Caches/.wrangler/cargo-generate-0.5.0/cargo-generate generate --git https://github.com/cloudflare/durable-objects-template --name rate-limit-actions --force
exited with exit code: 1
```

jldec:
see https://discord.com/channels/595317990191398933/799437470004412476/828395478244524082

jyn514:
Github actually lets you determine the primary branch automatically, there's no need to hard-code `main` or `master`. The latest version of `cargo-generate` does this automatically:
```
$ cargo generate --git https://github.com/cloudflare/websocket-template --name websocket
⚠️   Unable to load config file: /home/jnelson/.local/lib/cargo/cargo-generate.toml
🔧   Generating template ...
[ 1/14]   Done: .github/workflows/deploy.yml
[ 2/14]   Done: .github/workflows
[ 3/14]   Done: .github
[ 4/14]   Done: .gitignore
[ 5/14]   Done: .prettierrc
[ 6/14]   Done: CODE_OF_CONDUCT.md
[ 7/14]   Done: LICENSE_APACHE
[ 8/14]   Done: LICENSE_MIT
[ 9/14]   Done: README.md
[10/14]   Done: index.js
[11/14]   Done: package-lock.json
[12/14]   Done: package.json
[13/14]   Done: template.js
[14/14]   Done: wrangler.toml
🔧   Moving generated files into: `/home/jnelson/src/wrangler/websocket`...
[...]

PR #2132, opened 2021-11-10
Update version of cargo-generate

jyn514:
This fixes generating projects based on templates with a default branch other than "master".

Fixes https://github.com/cloudflare/wrangler/issues/1704.

It might make sense to eventually support passing a branch other than the default, but this at least fixes the hard error you get if the repo doesn't have a master branch.

### DIFF
--- a/src/install/dependencies.rs
+++ b/src/install/dependencies.rs
@@ -1,2 +1,2 @@
 pub const WASM_PACK_VERSION: &str = "0.10.0";
-pub const GENERATE_VERSION: &str = "0.5.0";
+pub const GENERATE_VERSION: &str = "0.11.0";


threepointone (approved):

threepointone:
^ this appears to be failing tests

jyn514:
Hmm, that's strange, the tests pass on linux ... I looked a little in that code and it looks like it's downloading some stuff from `workers.cloudflare.com` - @threepointone do you have a Mac? Can you download https://workers.cloudflare.com/get-binary/ashleygwilliams/cargo-generate/v0.11.0/x86_64-apple-darwin.tar.gz and tell me whether it runs ok for you?

PR #2133, opened 2021-11-12
Give a 500 internal error instead of aborting the connection on error

jyn514:
When proxying connections, wrangler will currently completely abort the TCP connection if `client.request()` returns an error (e.g. if it can't connect to the remote server). This changes it to display the error to the client instead.

### DIFF
--- a/src/commands/dev/edge/server/http.rs
+++ b/src/commands/dev/edge/server/http.rs
@@ -7,6 +7,7 @@ use std::sync::{Arc, Mutex};
 
 use anyhow::Result;
 use chrono::prelude::*;
+use http::StatusCode;
 use hyper::service::{make_service_fn, service_fn};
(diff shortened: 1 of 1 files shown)


src/commands/dev/edge/server/http.rs
```diff
@@ -47,7 +48,7 @@ pub async fn http(
                 let req_method = parts.method.to_string();
                 let now: DateTime<Local> = Local::now();
                 let path = get_path_as_str(&parts.uri);
-                async move {
+                let resp = async move {
```
jyn514, line 51: err actually I don't think this does what I want because `?` returns from the whole closure, not just the async block. Let me fix that.

jyn514:
Just realized it seems kind of silly to do this for edge + http and nothing else ... going to close this until I think of a way to do it more consistently.

Issue #2079, opened 2021-09-18
WebSocket does not work with `wrangler dev`


mnixry:
#### Describe the bug

Cloudflare Worker supports establish experimental WebSocket connection with users, which could work with production environment perfectly. But development server started by `wrangler dev` does not work.

With default config, we cannot built any WebSocket connection because worker cannot receive `Upgrade` header

Once set up configs below, Worker can receive `Upgrade` header and could establish a WebSocket connection, but it disconnected immediately

```toml
local_protocol = "http"
upstream_protocol = "http"
```

#### Reproduce the bug

Create a worker provides WebSocket server with [official demo](https://github.com/cloudflare/websocket-template/blob/main/index.js)

#### Environment and versions

Fill out the following information about your environment.

* operating system: Manjaro 21.1.4
* output of `wrangler -V`: wrangler 1.19.3-rc.1
* output of `node -v`: v16.9.1
* content of `wrangler.toml`:
```toml
name = "onebot-worker"
type = "javascript"
zone_id = ""
account_id = ""
route = ""
workers_dev = true

[dev]
local_protocol = "http"
upstream_protocol = "http"

[build]
command = "npm run build"

[build.upload]
format = "service-worker"
```

threepointone:
Good catch, we should definitely fix this.

threepointone:
- As a workaround, please consider using https://miniflare.dev/ for local development involving WebSockets (This is a project blessed by cloudflare, and we'll be collaborating with the package and its author more deeply). 
- The proper fix involves updating our local proxy to upgrade the connection and proxying websocket messages correctly, which we'll do as soon. 

I'm going to leave this issue open till we have a proper fix.

jyn514:
Fixed in https://github.com/cloudflare/wrangler/pull/2135 and https://github.com/cloudflare/wrangler/pull/2153.

ben-xD:
Unfortunately, I am not able to connect to my localhost worker - which I started with `cloudflare dev`.

I get 2 errors in Firecamp (websocket client):
- `error: connect ECONNREFUSED 127.0.0.1:8787`
- `1006: A connection was closed abnormally, when a status code is expected.`

It works when connecting to a `worker.dev` worker. 
![firecamp@2x](https://user-images.githubusercontent.com/24711048/160297701-4baeca73-8a7b-4490-941f-1b0cf1bba835.png)

[...]

Issue #1910, opened 2021-05-04
Cannot connect to websocket with wrangler dev


universse:
#### Describe the bug

I keep getting this message when trying to connect to websocket using `wrangler dev`. Non-websocket routes work fine though.

```
[2021-05-04 22:54:59] GET dev-bug.ulrjch.workers.dev/ws HTTP/1.1 400 Bad Request
[2021-05-04 22:55:08] GET dev-bug.ulrjch.workers.dev/ HTTP/1.1 200 OK
```
![image](https://user-images.githubusercontent.com/15166715/117025215-4d7b6c80-ad2d-11eb-8530-ea325f5c5217.png)

#### Reproduce the bug

1. Clone https://github.com/universse/wrangler-bug
2. Enter `account_id` in `wrangler.toml`
3. Run `wrangler dev`
4. Open `index.html`
5. Above error messages would appear in terminal.

Optional: It works fine when published.
1. Run `wrangler publish`
2. Replace websocket link in `index.html`
3. Open `index.html`
4. Websocket connection is successful.

#### Expected behavior

Should be able to connect to websocket using `wrangler dev`.

#### Environment and versions

Fill out the following information about your environment.

* operating system: Windows 10 Home
* output of `wrangler -V`:  wrangler 1.16.0-durable-objects-rc.0
* output of `node -v`: v15.10.0
* content of `wrangler.toml`: per repo

universse:
I have tried `type: "webpack"` for `wrangler.toml` as well. It doesn't work still.

Electroid:
We currently do not support WebSocket debugging in `wrangler dev`, but obviously we should. Will get back to you as we figure out the complexity for this addition.

dan-online:
Hey it's September now, how's this doing? I've found the same issue with wrangler dev and can confirm it only works when published.

```bash
$ wrangler -V
wrangler 1.19.3
```

threepointone:
We've scheduled to fix this in q4 (ie - before the end of this year) Until then, the recommendation is to use miniflare https://github.com/cloudflare/wrangler/issues/2079#issuecomment-924759723 Sorry for the delay on this, we'll fix it soon.

seeekr:
Just a heads up to maybe update the note at https://developers.cloudflare.com/workers/learning/using-websockets#writing-a-websocket-server in case this should already be fixed in production!

threepointone:
[...]

PR #2135, opened 2021-11-16
Proxy websocket connections when using authenticated (realish) preview

jyn514:
Previously, Wrangler would return a "101 Switching Protocols" request
and then immediately close the TCP connection. This changes it to instead
continue proxying the connection to the remote worker.

This is simpler than `wrangler dev --inspect` (and uses a different codepath) because it only
proxies the TCP connection directly rather than trying to inspect each websocket message.

# How do I use this?

1. Create a worker that uses WebSockets. I used https://github.com/cloudflare/websocket-template with the `protocol` in template.js changed from `wss` to `ws`.
2. Run `wrangler dev http http`.

Note that `wrangler dev` (i.e. with the default of `https` for the upstream connection) is currently broken. I am looking into this internally; it's an issue with the Cloudflare edge network and not with wrangler.

Unauthenticated preview is not currently supported.

Helps with https://github.com/cloudflare/wrangler/issues/2079 and https://github.com/cloudflare/wrangler/issues/1910.

### DIFF
--- a/src/commands/dev/edge/server/http.rs
+++ b/src/commands/dev/edge/server/http.rs
@@ -8,6 +8,7 @@ use std::sync::{Arc, Mutex};
 use anyhow::Result;
 use chrono::prelude::*;
 use hyper::service::{make_service_fn, service_fn};
+use hyper::upgrade::OnUpgrade;
 use hyper::{Body, Client as HyperClient, Server};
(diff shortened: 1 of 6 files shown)


Electroid (approved):
🎉 
Note: this will only work with HTTP, not HTTPS.

PR #2139, opened 2021-11-23
Add `wrangler1` as an alias

TehShrike:
So that when @cloudflare/wrangler is installed along wrangler2, they can each be referenced in npm run scripts.

See https://github.com/cloudflare/wrangler2/pull/40


threepointone (changes requested):
I don't think we'll do this here. We've landed the wrangler2 PR tho.

threepointone:
re-opening this, looks like it would be useful after all. Thanks for the PR!

threepointone (approved):

PR #2146, opened 2021-12-03
r2 bindings and subcommands

taylorlee:
this pr implements two pieces of r2 functionality for the wrangler client:

**commands**: add r2 sub-commands to create/delete/list buckets
I opted for sub-commands like `wrangler r2 bucket create`, rather than how kv does `wrangler kv:namespace create`.

**bindings**: add support for binding a bucket to a worker
this implementation is essentially copy-pasted from the kv binding code, and functions very similarly. 
preview isn't implemented yet, even though preview_bucket_name can be set currently.


Rastarev:
Excellent

jspspike (approved):

Issue #2161, opened 2021-12-18
1.19.6 Regression Segmentation fault


nickbabcock:
#### Describe the bug

Upgraded from 1.19.5 to 1.19.6 and now am receiving seg faults when I execute `wrangler publish`

> ✨  Build completed successfully!
> 🌀  Using namespace for Workers Site "__rakaly-dev-workers_sites_assets"
> ✨  Success
> 🌀  Uploading site files
> ✨  Successfully published your script to
> dev.rakaly.com/* => stayed the same
>
> Segmentation fault (core dumped)

I see in the 1.19.6 notes (https://github.com/cloudflare/wrangler/pull/2159) that it should fix a segfault (https://github.com/cloudflare/wrangler/issues/1464)

Current workaround is to downgrade to 1.19.5 via:

```bash
cargo install --version 1.19.5 wrangler
```

#### Reproduce the bug

Current reproduce steps are unknown

#### Expected behavior

No segfault

#### Environment and versions

Fill out the following information about your environment.

- Ubuntu 20.04
- node v16.13.1

threepointone:
@jyn514 ^

jyn514:
@threepointone I don't know a way to fix this other than just a straight revert. I'm slightly surprised this introduced *new* segfaults - I would expect `wrangler dev` to have crashed even before my change.

For what it's worth, this only happens when wrangler is already exiting, so it doesn't have any impact other than a scary error message.

threepointone:
Oh, if it's not actually breaking workflows, we should be fine with it for a while. Thanks for the notes @jyn514, I appreciate you! We can revisit this in january.

vlovich:
If people are relying on the exit code as part of automation I could see this as breaking a workflow....

jyn514:
@vlovich this only affects `wrangler dev`, which isn't scriptable. Nothing else calls process::exit other than the argument parsing.

nickbabcock:
> @vlovich this only affects `wrangler dev`, which isn't scriptable. Nothing else calls process::exit other than the argument parsing.

Apologies for the confusion, the segmentation fault occurs during `wrangler publish`

jyn514:
@nickbabcock does this reproduce consistently? Or only some of the times that you run `wrangler publish`?

If it reproduces consistently, could you post a backtrace of the core dump? You may have to run `ulimit -c unlimited` first to get wrangler to dump a core.

nickbabcock:
[...]

PR #2166, opened 2021-12-22
Check for updates in the foreground, not the background

jyn514:
This avoids creating a new OpenSSL context while the main thread exits the process.

The issue that #2161 ran into is that `try_recv` is non-blocking, so it would exit the process before the spawned thread finished creating the `reqwest::Client`.

Fixes https://github.com/cloudflare/wrangler/issues/2161.


threepointone (approved):
🦶

PR #2173, opened 2022-01-03
Experimental worker-to-worker bindings support

danielrs:
Adds support for experimental worker-to-worker bindings, letting you call a worker from another worker. The Wrangler config field looks like follows:

```toml
experimental_services = [{ name = "foo", service = "foo-service", environment = "production" }]
```
Where:
- `name` is the name of the binding in your worker code (alias `binding`)
- `service` is the service to reference
- `environment` is the environment of the script to reference (e.g. `production`, `staging`, etc)

Since this feature is experimental, you'll get a warning when you use it:

image

### What doesn't work?

- The API still doesn't support worker-to-worker bindings for previews, so `wrangler dev` will fail for now


jyn514:
@danielrs what's the upgrade path once this is no longer experimental? Will users need to change `experimental_services` to `services`? If so, there should probably be at least one version of wrangler that supports both.

My preference is probably to make the name `services` to start and keep the warning, though, that removes the need for a transition.

danielrs:
@jyn514 That's the plan, the versions that support both would have a deprecation warning on the old field. Wrangler V2 is already using `experimental_services` sadly, so that would make `wrangler.toml` files incompatible with each other. @threepointone @Electroid What are your thoughts?

threepointone:
I don't think this feature should be in wrangler v1 at all, tbh. I was under the impression that we landed support in wrangler v2 for internal testing only, not for customers to use.

ObsidianMinor:
It would kinda suck though if we GA service bindings before wrangler2 had an official 1.0 release that's compatible with wrangler1.

danielrs:
We all decided to postpone this feature in Wrangler V1 until it becomes GA. We'll keep working on this in Wrangler V2, and backport it once it's ready.

PR #2214, opened 2022-03-08
Revert #2096 - `[sites]` Don't delete unused assets

threepointone:
This reverts the changes made in #2096 that would preserve already uploaded assets with `[site]`, even if they weren't being used. This lads to longer upload times for sites with rapidly changing assets, and a burgeoning kv store. We're reverting this for now, and will add a proper fix later that expires unused assets on upload.

cc @jgentes


JacobMGEvans (approved):

jgentes:
We'll work on cutting a new release to get this out.

threepointone:
Before that, could we also implement expiring assets like we're doing in https://github.com/cloudflare/wrangler2/pull/587?

jgentes:
@caass is going to take a look, I agree it would be great to wait for that fix.

PR #2221, opened 2022-03-16
fix: expire unused assets on site upload

caass:
backport of https://github.com/cloudflare/wrangler2/pull/587

currently, we just delete unused assets when uploading a new version
of a workers-sites project. this can cause problems when a user
accessing an old site gets a 404 when going to a page when the new
assets haven't been pushed out to the edge yet, but the old assets have
already been deleted.

this commit addresses this by marking old assets for expiration
in five minutes, rather than deleting them outright. this should give
the new worker enough time to propogate to the edge before the old
assets are taken down.

### DIFF
--- a/src/commands/publish.rs
+++ b/src/commands/publish.rs
@@ -2,6 +2,7 @@ use std::env;
 use std::path::Path;
 
 use anyhow::Result;
+use cloudflare::endpoints::workerskv::write_bulk::KeyValuePair;
 use indicatif::{ProgressBar, ProgressStyle};
 use reqwest::blocking::Client;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


jspspike (approved):
Tested and seems to work as intended.

@jgentes This change might make publish take a lot longer. Maybe for the future we should make a bulk endpoint to get a set of values with a set of keys so each key doesn't have a request.

threepointone:
Looks like this is breaking people's deploys https://github.com/cloudflare/wrangler/issues/2224 We should fix this asap. I'll work on a solution for w2 myself.

threepointone:
I posted https://github.com/cloudflare/wrangler/issues/2224 for w2

Issue #2261, opened 2022-05-27
Add a message to Wrangler 1 indicating that the user should try Wrangler 2 instead.


petebacondarwin:

JacobMGEvans:
Is the messaging when running/initializing when using Wrangler 1 or improving the messaging in the GitHub and NPM package?

petebacondarwin:
The idea, in my mind, is that every time you run Wrangler 1 you get a "deprecation" warning that suggests you should try updating to the latest wrangler, giving instructions on how to do so.

PR #2273, opened 2022-06-22
Add deprecation warning

caass:
Closes https://github.com/cloudflare/wrangler/issues/2261
Screen Shot 2022-06-22 at 7 25 59 AM

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -9,8 +9,6 @@ use wrangler::cli::{exec, Cli, Command};
 use wrangler::commands;
 use wrangler::installer;
 use wrangler::reporter;
-use wrangler::terminal::message::{Message, StdErr};
-use wrangler::terminal::{emoji, styles};
 use wrangler::version::check_for_updates;
 
 use anyhow::Result;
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -20,6 +22,22 @@ fn main() -> Result<()> {
     }
     env_logger::init();
 
+    StdErr::deprecation_warning(&format!(
+        r#"{} {} {}
+
+You are using an outdated version of Wrangler (v{}).
+Please upgrade to v2 ({}).
```
petebacondarwin, line 29:
```suggestion
Please upgrade to the latest version ({}).
```

sidharthachatterjee (approved):
Looks good! This will be helpful in pushing folks to wrangler2

threepointone (changes requested):
Blocking so we can remove "Deprecation warning" and make it less scary.

src/main.rs
rozenmd: The problem is that major versions aren't sacred, and we could make a wrangler v3 if we needed to introduce a breaking change.

So "the latest version" saves us from having to update it several times in the future.

src/main.rs
```diff
@@ -20,6 +22,22 @@ fn main() -> Result<()> {
...
+Detailed migration instructions can be found here:
+{}
+"#,
+        emoji::WARN,
+        styles::warning("DEPRECATION WARNING"),
```
JacobMGEvans, line 34: Rather than "Deprecation" 
```suggestion
        styles::warning("UPGRADE AVAILABLE"),
```

src/terminal/message.rs
```diff
@@ -94,12 +94,22 @@ impl Message for StdErr {
...
+            .border_style(BorderStyle::Round)
+            .border_color(BorderColor::Cyan)
+            .margin(1)
+            .build();
+        eprint!("{}", billboard.as_str(msg));
     }
 
-    fn deprecation_warning(_msg: &str) {
-        panic!("Can't display billboard warning to stderr.")
+    fn deprecation_warning(msg: &str) {
```
JacobMGEvans, line 106:
```suggestion
    fn upgrade_message(msg: &str) {
```

caass:
If memory serves, we had decided to incorporate this instead into the normal self-update check, so that's what I've done here.

petebacondarwin (changes requested):

petebacondarwin (approved):

PR #2291, opened 2022-08-13
Fixed AES OCB fails to encrypt some bytes

imhunterand:
AES OCB mode for 32-bit x86 platforms using the AES-NI assembly optimised implementation will not encrypt the entirety of the data under some circumstances. This could reveal sixteen bytes of data that was preexisting in the memory that wasn't written. In the special case of "in place" encryption, sixteen bytes of the plaintext would be revealed.

Since OpenSSL does not support OCB based cipher suites for TLS and DTLS,
they are both unaffected.

**CVE-2022-2097**
`7.5/ 10`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1689,7 +1689,7 @@ checksum = "ff011a302c396a5197692431fc1948019154afc178baf7d8e37367442a4601cf"
 
 [[package]]
 name = "openssl-src"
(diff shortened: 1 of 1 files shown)


jspspike:
I had an issue building due to the version change not being semver, made a PR and merged it. Thanks for bringing this up!

PR #2306, opened 2022-09-08
Update README.md

rozenmd:


GregBrimble (approved):

petebacondarwin:
Drive by comment...
I know that in this case it is a minimal change, but it is really helpful if the commit message has a bit more context. 
If in the future, I am looking down the history and see `"Update README.md"` for this commit, I have to go into the diff to see what actually changed.
Something like `"Add README note to point users at the new version of Wrangler"` or something would be better.

PR #2314, opened 2022-09-29
chore: deprecation

JacobMGEvans:
Internal ticket: Messaging for Wrangler1 deprecation

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -17,17 +17,13 @@ fn main() -> Result<()> {
 
     StdErr::deprecation_warning(&format!(
         r#"{} {} {}
-You are using deprecated Wrangler 1 (v{}).
-Please upgrade to v2 ({}).
-Detailed migration instructions can be found here:
-{}
+The version of Wrangler you are using is now deprecated.
+Please update to the latest version of wrangler to prevent critical errors.
+Run `npm uninstall -g @cloudflare/wrangler && npm install -g wrangler` to update to the latest version.
 "#,
         emoji::NO_ENTRY,
         styles::warning("DEPRECATED"),
         emoji::NO_ENTRY,
-        env!("CARGO_PKG_VERSION"),
-        styles::url("https://github.com/cloudflare/wrangler2"),
-        styles::url("https://developers.cloudflare.com/workers/wrangler/migration/migrating-from-wrangler-1/")
     ));
 
     if let Ok(me) = env::current_exe() {
--- a/src/main.rs
+++ b/src/main.rs
@@ -23,7 +23,7 @@ Detailed migration instructions can be found here:
 {}
 "#,
         emoji::NO_ENTRY,
-        styles::warning("DEPRECATION"),
+        styles::warning("DEPRECATED"),
         emoji::NO_ENTRY,
         env!("CARGO_PKG_VERSION"),
         styles::url("https://github.com/cloudflare/wrangler2"),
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -20,6 +22,21 @@ fn main() -> Result<()> {
     }
     env_logger::init();
 
+    StdErr::deprecation_warning(&format!(
```
rozenmd, line 25: If we're only using `StdErr::deprecation_warning` here, why are we also defining `StdErr::billboard`?

src/main.rs
```diff
@@ -20,6 +22,21 @@ fn main() -> Result<()> {
...
+        r#"{} {} {}
+You are using deprecated Wrangler 1 (v{}).
```
rozenmd, line 27:
```suggestion
You are using Wrangler 1 (v{}), which is deprecated.
```

src/main.rs
JacobMGEvans: Previous code defined billboard for stdout, I figured it was necessary to have both because of the code here: 
https://github.com/cloudflare/wrangler/blob/38887e9a976292e0cdfa979097b88d4cf1ad9810/src/terminal/message.rs#L49

rozenmd: @jspspike / @caass pls halp

src/main.rs
```diff
@@ -20,6 +22,17 @@ fn main() -> Result<()> {
     }
     env_logger::init();
 
+    StdErr::deprecation_warning(&format!(
+        r#"{} {} {}
+The version of Wrangler you are using is now deprecated.
+Please update to the latest version of wrangler to prevent critical errors.
+Run `npm uninstall -g @cloudflare/wrangler && npm install -g wrangler` to update to the latest version.
```
rozenmd, line 29: They might have installed it via cargo, no?

JacobMGEvans: I will add the Cargo bit back.

JacobMGEvans: @lrapoport-cf Might need a different message template for this part.

lrapoport-cf: @JacobMGEvans @rozenmd how about the following:

```suggestion
If originally installed via npm, run `npm uninstall -g @cloudflare/wrangler && npm install -g wrangler` to update to the latest version.
If originally installed via cargo, run [xyz instructions]
```

JacobMGEvans: I am ok with this messaging, I will figure out the related `cargo` to Wrangler2 steps

JacobMGEvans: @rozenmd I updated the message

src/main.rs
caass: Because `billboard` and `deprecation_warning` are both methods defined as part of a [trait](https://doc.rust-lang.org/book/ch10-02-traits.html), both of them need to be defined -- you can think of it as implementing an interface for a class in JS, and you only really want access to one of the class methods.

You can leave the billboard method unimplemented, however, and just use the `unimplemented!()` macro to tell the compiler "yea so i know i just said i would do this but i'm actually not gonna":

```rust
    fn billboard(_msg: &str) {
        unimplemented!("Can't display billboard to stderr.")
    }
```

caass (approved):
Looks fine to me, it's ok to have dead code in `billboard` I think -- it's not hurting anybody.

mrbbot (approved):

PR #2316, opened 2022-09-30
chore: remove reporter

JacobMGEvans:
As part of the deprecation efforts, removed reporter as we will no longer be supporting Wrangler 1


JacobMGEvans:
Screen Shot 2022-09-30 at 2 38 40 PM
Screen Shot 2022-09-30 at 2 38 58 PM

caass (commented):
Does this also remove the autogeneration of reports on failure and the text after a failure indicating you should submit a report?

JacobMGEvans:
> Does this also remove the autogeneration of reports on failure and the text after a failure indicating you should submit a report?

I will double check the message but I deleted the entire "report" file which autogenerated the report on `panic`

PR #2350, opened 2023-02-21
1.20.1

jspspike:


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "wrangler"
-version = "1.20.0"
+version = "1.20.1"
```
penalosa, line 3: Do we want to deprecate in a patch version?

penalosa (approved):

Cargo.toml
jspspike: I agree, I'll open a new PR

